PR #59, opened 2026-03-25
fix: overrides fast xml parser dependency

floriangolling:


floriangolling:
fix: override fast-xml-parser to >=4.5.4 (CVE-2026-25896)

fast-xml-parser 4.1.3–4.5.3 has a critical entity encoding bypass
(CVSS 9.3) where a dot in a DOCTYPE entity name is treated as a regex
wildcard during entity replacement, allowing attackers to shadow
built-in XML entities with arbitrary values and achieve XSS.

The vulnerable version is pulled in transitively by
@langchain/anthropic@0.3.x (which depends on fast-xml-parser ^4.4.1).
Upgrading @langchain/anthropic to 1.x (which drops fast-xml-parser)
is not feasible as it requires @langchain/core 1.x, breaking our
current 0.3.x setup. A pnpm override is the least disruptive fix.

Can be removed once @langchain/anthropic 0.3.x ships a fix upstream
or we migrate to 1.x.

gwitwer (approved):

gwitwer:
Closing: superseded. main's lockfile resolves only fast-xml-parser 4.5.7 since #254 (2a97d64e). This branch also pins packageManager back to pnpm 8 while main is on pnpm 10. Thanks for the fix.

PR #63, opened 2026-06-04
fix(sandbox): treat all terminal process statuses as done when polling

raxl8:
The sandbox SDK's process polling only treated `status === "completed"` as finished, so any command that exited non-zero (reported by the sandbox-api as `"failed"`) was never recognized as done — `exec`/`execStream`/`waitForProcess` kept polling until the exec timeout and then threw, even though the command had finished in ~1s. This change makes all terminal statuses (`completed`, `failed`, `killed`, `stopped`) resolve with the real exit code, and defaults a terminal status that omits an exit code to non-zero so a failure can't be silently read as success. It adds 10 unit tests covering the regression, each terminal status, the polling/timeout backstop, and the synchronous short-circuit paths, plus a patch changeset for `@sapiom/sandbox`. This is the durable SDK-side fix for a process-hang that a downstream workflow was previously mitigating with a client-side workaround.

### DIFF
--- a/.changeset/sandbox-poll-terminal-status.md
+++ b/.changeset/sandbox-poll-terminal-status.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/sandbox": patch
+---
+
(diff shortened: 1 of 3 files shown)


gwitwer (approved):
ty!

PR #68, opened 2026-06-05
chore(ci): disable auto-publish, document manual sandbox release

gwitwer:
## What

- **Disable CI auto-publish.** `publish.yml` switches from the `push` (on changelog) trigger to `workflow_dispatch` only.
- **Add `PUBLISHING.md`** documenting the working manual release process.

## Why

npm's `@sapiom` org requires an **interactive 2FA flow** (browser / security key) to publish. CI — and `pnpm changeset publish` / `pnpm release` run non-interactively — can't perform that flow, so the Publish step failed on every "version packages" merge (the confusing `403`/`404` we just chased down). Until a 2FA-bypass automation token is in place, automated publish can't work, so it shouldn't run and fail noisily.

`PUBLISHING.md` captures the process that actually works, and the gotchas that bit us:
- bare `npm publish` ships `workspace:*` unresolved → broken package (what happened to `0.8.1`)
- `pnpm pack` + interactive `npm publish <tarball>` resolves the workspace dep **and** gets the 2FA prompt
- how to verify, and how to deprecate a bad version

Scoped to `@sapiom/sandbox`, the only maintained package for now. The workflow is kept (manual-dispatch) and the doc has a "reviving automation" section for when the universal fix lands.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,11 +1,12 @@
 name: Publish
 
+# Auto-publish on release is disabled: npm requires an interactive 2FA flow
+# (browser / security key) that CI cannot perform, so the publish step always
+# fails. Packages are published manually for now — see PUBLISHING.md. This
+# workflow is kept as manual-dispatch-only so it can be revived once a
+# 2FA-bypass automation token is in place.
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - '**/CHANGELOG.md'
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


xoxo-david (dismissed):

gwitwer:
Added the **Option C** local OTP flow to `PUBLISHING.md`:

- Primary publish command is now `pnpm changeset publish --otp=<code>` — resolves `workspace:*` *and* authenticates without a browser.
- Documented the required npm account setup: a **TOTP authenticator app** + 2FA level **"Authorization and writes"**. That second setting is what caused the earlier `404` — `changeset publish` only prompts/sends an OTP when `tfa.mode === auth-and-writes`, otherwise it publishes without one and npm rejects it.
- Kept the security-key tarball method as a fallback.

xoxo-david (approved):

PR #87, opened 2026-06-19
feat(cli): host-switch + auth hardening, retarget to backend tenant API

gwitwer:
## Summary

- **Auth header fix**: `GatewayClient` now sends `x-api-key` (the header the backend `UnifiedAuthGuard`'s `ApiKeyStrategy` reads) instead of the old `x-sapiom-api-key` that the engine gateway accepted.
- **Retarget to backend**: default host changed to `https://api.sapiom.ai`; `run` now POSTs to `POST /executions` with `{definitionId, input}` in the body instead of `POST /definitions/:id/execute` (matching the backend tenant controller).
- **Host switch**: every networked command (`run`, `deploy`, `link`, `logs`, `signal`) gains `--target local|prod` and `--host <url>` flags. `local` resolves to `http://localhost:3000`.
- **Persisted target**: `sapiom config set-target <local|prod>` writes to `~/.sapiom/config.json`. Host resolution order: `SAPIOM_HOST` env > `--host` > `--target` > persisted config > project `sapiom.json` host > prod default.
- **CLI stays thin**: all logic changes are in `@sapiom/orchestration-core`; CLI adds flag-parsing and resolution only.

## Auth header confirmed

The backend `ApiKeyStrategy` in `UnifiedAuthGuard` extracts the key from `x-api-key` header or `Authorization: Bearer sk_*`. The old header `x-sapiom-api-key` is not recognized by the backend. `sapiom login` stores the device-flow token as `apiKey` in credentials; `resolveApiKey()` reads it and the core client sends it as `x-api-key`.

## Commands now targeting backend tenant API

| Command | Backend route | Status |
|---------|--------------|--------|
| `run` | `POST /v1/workflows/executions` | Working |
| `logs` (execution) | `GET /v1/workflows/executions/:id` | Working |
| `logs` (list) | `GET /v1/workflows/executions` | Working |
| `signal` | `POST /v1/workflows/executions/:id/signals` | Working |
| `link` (list) | `GET /v1/workflows/definitions` | Working |
| `deploy` / `link --create` | `POST /v1/workflows/definitions`, push-credentials, builds | Pending backend deploy routes (separate effort) |

## Test plan

- [x] `pnpm --filter @sapiom/orchestration-core build` passes
- [x] `pnpm --filter @sapiom/cli build` passes  
- [x] `pnpm --filter @sapiom/orchestration-core test` — 28 tests pass (updated for new host/header/endpoint)
- [x] `pnpm --filter @sapiom/cli test` — 11 new tests pass (resolveHost precedence)
- [x] `pnpm --filter @sapiom/cli lint` clean
- [x] `pnpm --filter @sapiom/orchestration-core lint` clean

Generated with Claude Code

### DIFF
--- a/packages/cli/src/__tests__/cli-config.test.ts
+++ b/packages/cli/src/__tests__/cli-config.test.ts
@@ -1,0 +1,133 @@
+/**
+ * Tests for host-switch resolution and CLI-level configuration.
+ *
+ * The key invariants:
+ *   - SAPIOM_HOST env always wins
+ *   - --host flag beats --target flag
+ *   - --target local maps to localhost:3000, prod to api.sapiom.ai
+ *   - Persisted config (target / host) is the fallback before project host
+ *   - Project host from sapiom.json is the last fallback before default
+ *   - Default is the production backend
+ */
+import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+// We test resolveHost by mocking XDG_CONFIG_HOME to point at a temp dir so
+// the test never touches the real ~/.sapiom/config.json.
+import { resolveHost } from '../lib/cli-config.js';
+
+const PROD = 'https://api.sapiom.ai';
+const LOCAL = 'http://localhost:3000';
+
+function makeTmpDir(): string {
+  const dir = path.join(os.tmpdir(), `sapiom-cli-test-${process.pid}-${Math.random().toString(36).slice(2)}`);
+  mkdirSync(dir, { recursive: true });
+  return dir;
+}
+
+function writeConfigJson(xdgDir: string, content: object): void {
+  const sapiomDir = path.join(xdgDir, 'sapiom');
+  if (!existsSync(sapiomDir)) mkdirSync(sapiomDir, { recursive: true });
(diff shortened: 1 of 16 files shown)


positiveblue:
This is only for workflows right? It never points to unified gateway? 

If it does we need to make sure for those req it does use `x-sapiom-api-key` so it uses the collapsed path.

positiveblue (approved):

gwitwer:
Workflows-only. The client's base is always `${host}/v1/workflows` (the backend tenant API) and every request goes through it — it never points at the unified gateway, and there's no path that builds a `*.services` host or `/v1/{coding,compute,git,…}` URL. So the only header it sends is `x-api-key`, which is what the backend's `UnifiedAuthGuard`/`ApiKeyStrategy` reads. Capability/gateway calls happen server-side during execution (the engine mints its own per-run token), not from this client. If we ever add a direct gateway-bound request here, those would use `x-sapiom-api-key` for the collapsed path — but there are none today.

xoxo-david:
@gwitwer about:
> The client's base is always ${host}/v1/workflows (the backend tenant API) and every request goes through it — it never points at the unified gateway

why are we doing it this way? the unified gateway *does* proxy all workflows-engine routes and handles authenticating the Sapiom API key + pinning the tenant based on it

Issue #88, opened 2026-06-20
examples: x402-demo.sapiom.ai returns 403 — HTTP examples fail


royalsanga24:
## Problem

The demo server used by the HTTP examples (`DUMMY_SERVER_URL=https://x402-demo.sapiom.ai`) returns **403** on all requests, including public endpoints like `/api/public/time`.

Running `npm start` in `examples/axios` (same for `fetch` and `node-http`) fails every step with:

`✗ Failed: Request failed with status code 403`

## Reproduce
```bash
curl -v https://x402-demo.sapiom.ai/api/public/time
```

Returns 403 with a Cloudflare block page (server: cloudflare), not JSON from the demo app.

Cloudflare Ray ID: a0ec5e7d5d63e62a

For comparison, https://api.sapiom.ai returns 200 from the same machine — so this doesn't look like an API key / .env issue.

erhnysr:
## Diagnosis: server-side Cloudflare block, not a client/URL/header bug

Reproduced and traced. **Root cause: a Cloudflare WAF/firewall is returning a hard `403` to *all* traffic for `x402-demo.sapiom.ai`, before any request reaches the demo app.** Nothing in the examples repo (client code, headers, or the URL string) can fix it — it needs the demo's operator to restore the origin / repair the Cloudflare rule.

### What I reproduced (from a different machine/IP than the reporter)
- **DNS resolves fine** → Cloudflare IPs (`172.67.71.244`, `104.26.0.175`, `104.26.1.175`); TLS connects. So it is *not* a stale/NXDOMAIN hostname.
- **Every path returns `403`**, including the no-auth public ones:
  - `GET /` → 403 · `GET /api/public/time` → 403 · `GET /api/public/status` → 403
- The body is the Cloudflare **"Sorry, you have been blocked"** page (`server: cloudflare`, `cf-ray` present, `cf-error-details`): *"This website is using a security service to protect itself from online attacks. The action you just performed triggered the security solution."* This is Cloudflare's managed-WAF/firewall block — **not** a JS "checking your browser" challenge and **not** the origin app.
- **`https://api.sapiom.ai` returns `200`** from the same machine (Express behind the same Cloudflare). So Cloudflare/the org isn't broadly broken; the block is specific to the demo hostname. Matches the reporter's comparison.
- Two independent IPs (reporter's + mine, in AMS) both blocked with different Ray IDs → a broad rule affecting the whole hostname, not a per-IP block.

### Ruling out the other candidates
[...]

PR #89, opened 2026-06-20
fix(examples): fail fast when demo server is unreachable

royalsanga24:
## Summary
- Add a startup health check for the HTTP examples (axios, fetch, node-http)
- Probe `/api/public/time` with plain `fetch` before the SDK runs
- Detect Cloudflare 403 blocks and exit with a clear error message
- Document troubleshooting in `examples/README.md`

Follow-up to #<your-issue-number>.

## Test plan
- [x] `npm start` in `examples/axios` — exits early with Cloudflare message when demo is blocked
- [ ] When demo server is restored, examples proceed past the health check

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -150,6 +150,12 @@ The Sapiom SDK handles all authorization and payment automatically - your code j
 **Connection refused / timeout**
 - Check that `DUMMY_SERVER_URL` is correct and the test server is running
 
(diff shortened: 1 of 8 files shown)


gwitwer:
Closing in favour of #689, which addresses issue #88 fully (runs the HTTP demos against a local demo server) rather than only failing fast. Thanks.

PR #91, opened 2026-06-22
fix(orchestration-core): resolve bundled templates dir in the ESM build

gwitwer:
## Summary
`orchestrations init` failed with `Unknown template 'default'. No templates are bundled.` for any ESM consumer (the CLI and the published package), unless the caller set `SAPIOM_TEMPLATES_DIR`. The bundled `templates/` directory was resolved relative to the compiled module, but in the ESM build that anchor was empty.

## Root cause
`moduleDir` was `typeof __dirname !== 'undefined' ? __dirname : ''`. In the ESM build there is no `__dirname`, so `moduleDir` was `''` and `templates/` resolved against the current working directory instead of the package root.

## Fix
Resolve the module directory for both build formats:
- CommonJS → `__dirname`
- ESM → derived from `import.meta.url` via `fileURLToPath`

`import.meta.url` is read through `eval` so the single source file still compiles under `module: commonjs` (a literal `import.meta` is a hard compile error there). Node-targeted.

## Testing
- `pnpm --filter @sapiom/orchestration-core build` — both CJS and ESM builds compile.
- From a clean directory with **no** `SAPIOM_TEMPLATES_DIR` set, `orchestrations init my-agent` now scaffolds the full project (README, package.json, tsconfig.json, index.ts, etc.) — previously this failed in the ESM/CLI path.

## Checklist
- [x] Conventional commit
- [x] No secrets
- [x] Manually verified init works in the ESM build

### DIFF
--- a/packages/orchestration-core/src/scaffold.ts
+++ b/packages/orchestration-core/src/scaffold.ts
@@ -11,16 +11,35 @@
  */
 import { cpSync, existsSync, mkdirSync, readdirSync, readFileSync, renameSync, statSync, writeFileSync } from 'node:fs';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 import { OrchestrationError } from './errors.js';
 
 /**
- * Directory of this module. `__dirname` exists in the CommonJS build; the ESM
- * build has no `__dirname`, so an ESM caller must supply the templates location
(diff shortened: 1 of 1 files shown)


evtran0209 (approved):
LGTM! 🔥

PR #101, opened 2026-06-23
feat(tools): add contentGeneration capability with optional storage stitch (SAP-892)

evtran0209:
## What

Adds a typed **`fal`** capability to `@sapiom/tools` — image generation over the Sapiom Fal gateway, with the optional **`storage`** param that wires up the server-side Fal × file-storage stitch (SAP-887). Closes **SAP-892**.

```ts
import { createClient } from "@sapiom/tools";
const sapiom = createClient({ apiKey: process.env.SAPIOM_API_KEY });

const out = await sapiom.fal.run({
  model: "fal-ai/flux/schnell",
  input: { prompt: "a red bicycle", num_images: 1 },
  storage: { visibility: "private" }, // optional — persist each output
});
out.images[0].url;      // Fal's hosted URL
out.images[0].file_id;  // present when `storage` was passed → use with fileStorage
```

## Design

Mirrors the existing `fileStorage` capability exactly:
- Standalone `run(input, transport?, baseUrl?)` on the shared `Transport`; bound onto `createClient().fal`, the ambient `fal` namespace, and the `@sapiom/tools/fal` subpath.
- `FalHttpError` (status + parsed body) on non-2xx — exported from the barrel.
- Shape-faithful entry in the stub client (`fal.run`), per-capability README + a row in the package README capabilities/composing sections, changeset (**minor**).
- Host override via `SAPIOM_FAL_URL` (defaults to the prod gateway).

**One deliberate deviation, documented in code:** `fal` is a **passthrough** (both `input` and the response are Fal-native snake_case, matching Fal's docs), whereas `fileStorage` is a Sapiom-native API mapped to camelCase. The stitch only annotates `file_id` / `storage_error` inline, so the response is forwarded verbatim rather than remapped.

## Acceptance (SAP-892)
- [x] typed Fal fn with optional `storage`; binding tests pass
- [x] published **after** the file-storage fn (SAP-885 already shipped `0.2.0`)

## Tests
9 binding tests vs an injected scripted `fetch` mock: body shape + credential, `storage` merge/omit, multi-image `file_id`/`storage_error` passthrough, slash-preserving model paths (no `%2F`, empty segments dropped), the async-model `400` → `FalHttpError`, the empty-model guard (no fetch), and `createClient().fal` wiring (default host + credential + storage merge).

Local: typecheck ✅ · jest **51/51** ✅ · eslint (0 warnings) ✅ · build cjs+esm ✅ · prettier ✅


evtran0209:
## Revised: provider-neutral surface

Per feedback that **user-facing SDK APIs shouldn't leak upstream provider names**, the capability was renamed from `fal` → **`contentGeneration`** before merge:

| Before | After |
|---|---|
| `sapiom.fal.run({ model, input, storage })` | `sapiom.contentGeneration.images.create({ prompt, storage? })` |
| `FalHttpError` | `ContentGenerationHttpError` |
| `@sapiom/tools/fal` | `@sapiom/tools/content-generation` |
| `SAPIOM_FAL_URL` | `SAPIOM_CONTENT_GENERATION_URL` |

- Structured for sibling **`.video` / `.audio`** sub-namespaces (SAP-970) — same neutral shape.
- `model` is now **optional with an internal default**, so the common call (`create({ prompt })`) names no provider at all; advanced params pass through.
- Surface-only rename of unpublished code; stub/tests/READMEs/exports/changeset all updated. typecheck + lint + build + **52 tests** green.

**Out of scope (flagged, tracked separately):** the gateway **host** (`fal.services…`) and **model ids** (`fal-ai/…`) still reference the provider — those are direct-API / infra concerns (DNS + a curated model-alias catalog), not the SDK surface.

xoxo-david (approved):

PR #107, opened 2026-06-25
feat(tools): add contentGeneration.video.create (async video + optional storage)

evtran0209:
## Summary
Adds `contentGeneration.video.create` to `@sapiom/tools` — the async sibling of `images.create`. Video generation submits the job and polls until the result is ready, then resolves, so callers `await` it just like an image. An optional `storage` param persists the output and the returned `video` carries a `fileId`.

```typescript
const out = await contentGeneration.video.create({
  prompt: "a calm ocean wave at sunset",
  storage: { visibility: "private" }, // optional — persist the output
});
out.video?.url;     // hosted URL
out.video?.fileId;  // present when `storage` was passed → use with fileStorage
```

## Changes
- **`createVideo` / `contentGeneration.video.create`** — submit → poll until ready → return the result. Configurable `pollIntervalMs` (default 5s) / `timeoutMs` (default 5min). Throws `ContentGenerationHttpError` on a failed submit; an `Error` on timeout.
- camelCase surface mapped from the snake_case wire (`content_type→contentType`, `file_id→fileId`, `storage_error→storageError`); `params` forwarded verbatim; `storage` truthy-checked — mirrors `images.create`.
- Wired into `createClient().contentGeneration.video` + the `stub` client.
- README (capability + per-dir) + changeset (minor).

## Testing
- [x] 6 new unit tests (submit+poll+map, storage→fileId, storage_error mapping, submit-failure-never-polls, timeout, client binding) — **15/15 content-generation tests pass**
- [x] typecheck + lint clean

## Notes
- Async by design: the video isn't in the submit response, so `create` polls until ready (the poll also persists the output when `storage` is set — no extra call).
- Scope: video only; audio is a trivial follow-up (same path, different default model).

## Related
- Closes: SAP-1048
- Refs: SAP-892 (images)

## Checklist
- [x] Conventional commit
- [x] Tests pass; typecheck + lint clean
- [x] No breaking changes
- [x] Changeset included
- [x] Self-reviewed

### DIFF
--- a/.changeset/content-generation-video.md
+++ b/.changeset/content-generation-video.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": minor
+---
+
+Add `contentGeneration.video.create` — generate a video from a prompt, with an optional `storage` param. Video generation is asynchronous: `create` submits the job and polls until the result is ready (configurable `pollIntervalMs` / `timeoutMs`), then resolves — so you `await` it just like `images.create`. When `storage` is passed, the output is persisted and the returned `video` carries a `fileId`. camelCase surface, mapped from the wire.
--- a/packages/tools/README.md
+++ b/packages/tools/README.md
@@ -71,7 +71,7 @@ Each capability is a namespace, importable from the barrel or its own subpath (e
 | `repositories` | Private, in-network git repos | [src/repositories](./src/repositories/README.md) |
 | `agent` | Coding agents (LLM execution) | [src/agent](./src/agent/README.md) |
 | `fileStorage` | Tenant-scoped object storage (presigned URLs) | [src/file-storage](./src/file-storage/README.md) |
-| `contentGeneration` | Media generation (images; video/audio soon), with optional `storage` | [src/content-generation](./src/content-generation/README.md) |
+| `contentGeneration` | Media generation (images + video; audio soon), with optional `storage` | [src/content-generation](./src/content-generation/README.md) |
 
 ## Composing capabilities
 
--- a/packages/tools/src/client.ts
+++ b/packages/tools/src/client.ts
@@ -42,6 +42,8 @@ import * as contentGeneration from "./content-generation/index.js";
 import type {
   ImageCreateInput,
   ImageGenerationResult,
+  VideoCreateInput,
+  VideoGenerationResult,
 } from "./content-generation/index.js";
 
 export interface Sapiom {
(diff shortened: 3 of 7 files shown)


evtran0209:
## Self-review

Reviewed the diff across correctness + cleanup, then applied judgment. Net: one fix; the rest intentionally left (with reasons).

**Fixed (`5b5a2f1`):** a non-ok poll response left its body unconsumed → now drained so the connection can be reused; polling continues. Added a test that a transient non-ok poll is tolerated and the result still returns once ready.

**Considered and intentionally left as-is:**
- Completion is detected by `video.url` presence — the result contract guarantees it on completion.
- A non-ok poll is tolerated rather than fail-fast on 4xx: a poll can be non-2xx while the job is still generating, so failing fast would risk the happy path. `timeoutMs` is the backstop for a result that never arrives.
- A malformed handle / timeout throws a plain `Error` (not `ContentGenerationHttpError`) — those aren't HTTP errors, and it's consistent between the two.
- `video` types/mappers mirror the `images` ones rather than sharing generics — matches the per-capability pattern already in this module.

16/16 content-generation tests pass; typecheck + lint clean.

evtran0209:
Good question — and the dispatchable-tools angle is the interesting part.

**Today it's client-side polling — no webhook, nothing blocks server-side.** `video.create` submits the job, gets back a handle, and the SDK polls the result endpoint until it's ready, then resolves. The server side is just submit + a stateless status/result endpoint — it doesn't hold a connection open or block waiting, and there's no callback. The poll loop lives in the SDK (`createVideo`), with configurable `pollIntervalMs` / `timeoutMs`. So inside a workflow step, `await contentGeneration.video.create(...)` does occupy the runner for the duration — it's a blocking await like any long call.

**On `pauseUntilSignal` / dispatchable tools:** correct that this is what we'd want, and `video.create` as written does **not** fit it. It returns the final result (`VideoGenerationResult`), not a `DispatchHandle` — there's no `dispatch: { correlationId, resultSignal }` — so `pauseUntilSignal(contentGeneration.video.create(...))` isn't expressible, and the step blocks rather than going idle.

To make it dispatchable it would mirror `agent.coding.launch` (the one capability that implements `DispatchHandle` today):
- a `video.launch()` that returns a `DispatchHandle` immediately (so the step pauses instead of polling);
- a stable result signal + a flat resume-payload shape (+ schema) the resumed step receives as its input;
- forwarding the workflow resume token so completion can be routed back to the right execution;
- and the key new piece: a **server-side completion trigger** that fires the resume signal when the video finishes. `agent.coding` gets that from its executor; a generation tool whose completion is currently observed by *polling* would need an equivalent server-side trigger added (a completion worker, or provider-side callbacks). This part is mostly server-side, not SDK.

And yes, it generalizes: the orchestration layer is capability-agnostic (it only reads `correlationId` + `resultSignal`), so any tool that can expose a launch-handle + a completion signal becomes dispatchable the same way — that's the right target for all indeterminate-duration tools.

The server-side specifics have a clear path modeled on the coding-run flow — I'll write that up and we can scope a small follow-up for `video.launch` + the completion trigger. (This PR is the blocking-`create` variant, which is still the right primitive for non-workflow/standalone callers; `launch` would be additive.)

gwitwer (approved):

PR #160, opened 2026-07-07
feat(orchestration-core): ExecutionProjection types + REST-parity inspect/listExecutions (SAP-1341)

gwitwer:
## Summary

Foundation for **S5 — SDK Inspection Parity**. Adds the canonical projection types to `@sapiom/orchestration-core` and rewires `inspect()` / `listExecutions()` to return the **same** tree + per-node cost + trace refs as the REST `ExecutionProjection` — no divergent SDK model.

Closes SAP-1341.

### Before
`inspect()` returned `{ execution: ExecutionDetail }` where `ExecutionDetail = { id, status, currentStep?, steps?: StepRecord[] }` — flat: no tree, no cost, no trace identity. `listExecutions()` returned `{ executions: ExecutionDetail[] }`.

### After
- **`packages/orchestration-core/src/types.ts`** (new) — single-owner projection types mirroring `interfaces.md`: `ExecutionProjection`, `StepProjection`, `CostNode { authorizedUsd, capturedUsd, settleState }`, `ExecutionRef { executionId, traceRoot, name, status }`, `DispatchRef`, `StepError { message, trace, traceUnavailableReason }`, `StepEvent`, `SettleState`. Tickets 2/3/4 extend this module without re-touching `inspect.ts`.
- **`decode.ts`** (new) — tolerant normalization of the raw REST body into a complete projection. The SDK stays a thin passthrough (no cost reshaping); decode only fills a fully-populated shape.
- **`inspect() -> ExecutionProjection`** (tree: `traceRoot/traceParent/traceId/spanId`, `parentExecutionId/rootExecutionId`, typed `children`; per-node `CostNode`; typed `DispatchRef`/`StepError`).
- **`listExecutions() -> ExecutionRef[]`** — tree-aware (`traceRoot` degrades to the row id).
- USD amounts carried as **strings**; `authorizedUsd` vs `capturedUsd` **never collapsed**; `settleState` carried through unchanged at every node granularity.
- **Graceful decode of pre-seam runs**: no `traceParent`/lineage → tree derived from the run's own id; missing cost → flat zeroed `CostNode`. Mirrors how the REST DTO degrades, so the SDK never throws on older executions.
- Minimal consumer updates (`cli` logs, `mcp` orchestrations tool) to the new return shapes so the monorepo compiles — full CLI/tools work is tickets 3/4.

## Tests
- `inspect.parity.spec.ts` — asserts `inspect()` decodes a **checked-in REST `ExecutionProjection` fixture** to the identical shape/nesting (parity), that cost legs are never collapsed, and that `DispatchRef`/`StepError` are typed.
- Pre-seam fixture — asserts graceful decode (degraded tree + flat cost, no throw), incl. empty/garbage bodies.
- `listExecutions()` tree-awareness + non-array fallback.
- Fixtures anonymized — no internal hostnames/customer data (**public repo**).

## Verification
`pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm test` all green for `orchestration-core` (68 tests), `cli` (19), `mcp` (68).

## Note on interfaces.md vs live REST DTO
Types mirror the canonical `interfaces.md` contract (the ticket's single-owner spec). The current backend/engine `ExecutionDetailDto` has partially drifted (children use a dispatch-edge `ExecutionRef`; per-step cost isn't emitted yet). SDK pins move in lockstep with the engine (~SAP-1018); the tolerant `decode` bridges the gap so today's reduced bodies still decode cleanly. Flagging for the reviewer.

## Acceptance criteria
- [x] `inspect()` returns the same tree + per-node cost + trace refs as the REST DTO (parity test green against a fixture).
- [x] `listExecutions()` returns `ExecutionRef[]` (tree-aware).
- [x] `authorizedUsd`/`capturedUsd`/`settleState` present at every node granularity; never collapsed.
- [x] Pre-seam executions decode without error (degraded tree + flat cost).
- [x] No internal hostnames/customer data in examples or fixtures.

### DIFF
--- a/.changeset/sap-1341-execution-projection-inspection.md
+++ b/.changeset/sap-1341-execution-projection-inspection.md
@@ -1,0 +1,17 @@
+---
+"@sapiom/orchestration-core": minor
+"@sapiom/cli": patch
+"@sapiom/mcp": patch
+---
+
+Bring `inspect()` / `listExecutions()` to REST `ExecutionProjection` parity (tree + per-node cost + trace refs), replacing the flat inspection shape.
+
+**Breaking (return shapes):**
+
+- `inspect(opts, client)` now resolves the decoded `ExecutionProjection` **directly** (previously `{ execution }`). It carries the dispatch tree (`traceRoot`/`traceParent`/`traceId`/`spanId`, `parentExecutionId`/`rootExecutionId`, typed `children`), per-step `spanId`/`events`/`dispatch`, and a structured `StepError` (`trace` is now a `StepErrorTrace` of source-mapped frames, not a string).
+- `listExecutions(client)` now resolves `ExecutionRef[]` **directly** (previously `{ executions }`).
+- The flat `ExecutionDetail` / `StepRecord` types are removed; import the new projection types (`ExecutionProjection`, `StepProjection`, `CostNode`, `ExecutionRef`, `DispatchRef`, `StepError`, `StepEvent`) instead.
+
+**Cost is honest, never fabricated:** `cost` is `CostNode | null` at run and step granularity. The execution-detail read is cost-agnostic today (authoritative cost lives at `/executions/:id/spend`), so an absent cost decodes to `null`, not a misleading `$0`. `authorizedUsd`/`capturedUsd`/`settleState` are never collapsed when cost is present.
+
+The engine must emit the corresponding fields (per-node cost, list lineage, named child edges) for the projection to be fully populated; until then `inspect()`/`listExecutions()` degrade honestly rather than throwing. SDK pins move in lockstep with the engine.
--- a/packages/cli/src/commands/orchestrations/logs.ts
+++ b/packages/cli/src/commands/orchestrations/logs.ts
@@ -28,19 +28,19 @@ export async function runLogs(
     const client = makeClient({ projectHost: cfg?.host, flagHost: opts.host, flagTarget: opts.target });
 
     if (!executionId) {
-      const { executions } = await listExecutions(client);
+      const executions = await listExecutions(client);
       if (isJsonMode()) ok({ executions });
-      else ok({}, executions.map((e) => `${e.id}  ${e.status}`));
+      else ok({}, executions.map((e) => `${e.executionId}  ${e.status}  ${e.name}`));
       return;
     }
 
-    const { execution: ex } = await inspect({ executionId }, client);
+    const ex = await inspect({ executionId }, client);
     if (isJsonMode()) {
       ok({ execution: ex });
       return;
     }
     const lines = [`execution ${ex.id}: ${ex.status}${ex.currentStep ? ` (at ${ex.currentStep})` : ''}`];
-    for (const step of ex.steps ?? []) {
+    for (const step of ex.steps) {
       lines.push(`  ${step.status === 'failed' ? '✗' : '·'} ${step.stepName} #${step.attempt} — ${step.status}`);
       if (step.error?.message) lines.push(`      ${step.error.message}`);
     }
--- a/packages/mcp/src/tools/orchestrations.ts
+++ b/packages/mcp/src/tools/orchestrations.ts
@@ -469,7 +469,7 @@ export function register(server: McpServer, env: ResolvedEnvironment): void {
               ...(hint ? { hint } : {}),
             });
           }
-          const { execution } = await inspect({ executionId }, client);
+          const execution = await inspect({ executionId }, client);
           // Self-correcting nudge: on a non-terminal snapshot, point at wait:true
           // so a caller reaches for the tool's loop instead of polling by hand.
           const hint = isExecutionTerminal(execution.status)
--- a/packages/orchestration-core/src/__tests__/client.test.ts
+++ b/packages/orchestration-core/src/__tests__/client.test.ts
@@ -175,18 +175,18 @@ describe('parseSignalPayload', () => {
 describe('inspect', () => {
   const client = createClient({ host: 'https://example.com', apiKey: 'sk' });
 
-  it('fetches execution by id (CLI-style)', async () => {
+  it('returns the decoded ExecutionProjection directly', async () => {
     const ex = { id: 'exec-1', status: 'completed', steps: [] };
     mockFetch([{ status: 200, body: ex }]);
     const result = await inspect({ executionId: 'exec-1' }, client);
-    expect(result.execution.id).toBe('exec-1');
-    expect(result.execution.status).toBe('completed');
+    expect(result.id).toBe('exec-1');
+    expect(result.status).toBe('completed');
   });
 
-  it('returns execution detail for MCP-style caller', async () => {
+  it('carries the current step through', async () => {
     const ex = { id: 'exec-2', status: 'running', currentStep: 'process' };
     mockFetch([{ status: 200, body: ex }]);
-    const { execution } = await inspect({ executionId: 'exec-2' }, client);
+    const execution = await inspect({ executionId: 'exec-2' }, client);
     expect(execution.currentStep).toBe('process');
   });
 });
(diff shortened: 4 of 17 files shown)


gwitwer:
## Review: correctness, safety, public-release readiness

Reviewed the type definitions, `decode.ts`, `inspect.ts`, the parity spec, and the client. The code quality here is genuinely high — TSDoc is excellent, the never-throw decode with graceful degradation is the right instinct, `cancelled`/`canceled` are both handled, and `waitForExecution`'s backoff/stop-reason logic is clean.

But I want to flag a theme before the specifics: **the "parity" is asserted against a hand-authored fixture, not a captured live response.** `expect(decodeExecutionProjection(fixture)).toEqual(fixture)` is circular — it proves `decode` is idempotent on a shape that was written to match the SDK types, not that the SDK matches what the API actually returns. Because `decode` never throws and back-fills every missing field, a real drift between the SDK shape and the wire surfaces as **silently wrong data** (zeroed cost, blank names, null traces) rather than a test failure. When I diff the SDK types against the shape the live endpoints actually emit, several gaps appear:

### 🔴 Correctness / parity gaps

1. **`inspect()` returns `$0` cost for every real run.** `inspect()` GETs `/v1/workflows/executions/:id`, and that response does **not** carry per-run or per-step cost — cost is served by the separate `/v1/workflows/executions/:id/spend` endpoint (with the run rollup + `byStep` breakdown). Since `decodeCostNode(undefined)` returns a zeroed `{authorizedUsd:"0", capturedUsd:"0", settleState:"final"}`, `inspect()` will report every run and every step as `$0`, settled. That directly contradicts the PR's headline ("same tree + **per-node cost** + trace refs") and the docstring's "never collapses authorized vs captured." Either `inspect()` needs to also fetch `/spend` and merge (run cost + `byStep` → per-step), or the cost fields should be documented as requiring a separate call. The parity test passes only because the fixture hand-includes cost the endpoint never sends.

2. **Structured error traces decode to `null`.** On the wire, a step's `error.trace` is a **structured object** (source-mapped frames + a `sourceMapped` flag + an optional raw string), not a string. The SDK types `StepError.trace` as `string | null` and decodes it with `strOrNull(raw.trace, raw.stack)`, which rejects the object → `trace: null`, and `traceUnavailableReason` is also `null`. So a failed step decodes to `{message, trace: null, traceUnavailableReason: null}` — a blank panel with no reason, which the docstring explicitly says should never happen. This drops the entire source-mapped-stack feature. Fix is SDK-side: model the structured trace (frames/`sourceMapped`/raw) or read its raw string.

3. **`listExecutions()` isn't actually tree-aware.** The list rows don't include root/parent lineage, so `decodeExecutionRef`'s `traceRoot` falls back to each row's own id. Every ref comes back with `traceRoot === executionId`, so callers can't group runs into dispatch trees — which is the stated point of the tree-aware list. This one likely needs the **list endpoint** to include the root execution id; until it does, the SDK can't deliver the promise.

4. **`children[]` come back with blank names.** The child edges on the detail response don't carry a display name, so `ExecutionRef.name` decodes to `""` (and, per #3, `traceRoot` degrades to the child's own id). A rendered tree would show nameless, self-rooted children. Needs a `name` on the child edge server-side, or drop `name` from the child ref contract.

5. **`DispatchRef.dispatchId` is always `""` (minor).** The wire dispatch edge uses `targetId`; `childExecutionId` resolves fine via the `targetId` fallback, but `dispatchId` has no matching source field and decodes to `""`. Either map it to the real source field or drop it.

### 🟡 Root cause & suggested fix
Re-derive the fixtures from a **real captured response** off a live/staging run (one with cost, a failed step, and a parent→child dispatch), and assert `inspect()` against *that*. That single change would have caught #1–#5. It's worth adding a note that the fixture is a real capture, with the date/run it came from.

### 🟢 Safety / public-release readiness
- Fixtures and changed files scanned clean — no secrets, internal hostnames, emails, or customer data. 👍
- No `process.env` reads, no `console`, explicit DI via `GatewayClient` — matches the package's design contract.
- **Breaking change:** `inspect()`'s return type changes shape. For a published SDK function this is a breaking change — please make sure the version bump / changeset reflects that (major, or clearly noted if still `0.x`), and that the deployed API actually emits the new fields (otherwise the degraded path becomes the default experience).
- Exporting `decodeExecutionProjection` / `decodeExecutionRef` / `decodeCostNode` widens the public API surface you'll have to maintain — intentional? If they're only for tests, consider keeping them internal.

### Verdict
Types, structure, and decode philosophy are solid and safe to be public. But the central claim — REST parity for **cost** and a **tree-aware** list — isn't met against the live endpoints today (#1, #3, #4), and structured error traces are dropped (#2). I'd hold merge on #1 and #2 (SDK-side, fixable here) and track #3/#4 as the server-side fields they depend on. Swapping the hand-written fixtures for real captures is the highest-leverage change and will keep this honest going forward.

gwitwer:
Thanks — this was the right call. I verified each claim against the engine code (the backend proxy is a faithful relay: `workflows-engine.client.ts` returns `payload as T` verbatim, so the wire is the engine's `ExecutionDetailDto`, not the reduced backend type). All five hold. Pushed d8d3419.

**#1 — $0 cost for every run (fixed).** Confirmed: `GET /v1/workflows/executions/:id` is cost-agnostic; cost lives at `/executions/:id/spend` (`ExecutionSpendDto` with run rollup + `byStep`). `decodeCostNode` now returns **`null`** for an absent cost instead of a zeroed node, and `cost` is `CostNode | null` at run and step granularity. So `inspect()` reports honest absence, not `$0/settled`. I did **not** fold a `/spend` fetch into `inspect()` — the ticket scoped it to a single-endpoint decode ("not a new endpoint"), and merging run cost + `byStep` into per-step is a real chunk of work. Two-legged never-collapse is preserved when cost *is* present (covered by the new with-cost fixture). If you'd rather `inspect()` fetch+merge `/spend` so it carries cost directly, I'll do it here or as the fast-follow that pairs with the tools ticket (SAP-1344) — your call.

**#2 — structured error trace (fixed).** Confirmed the wire sends `error.trace` as `StepErrorTrace { frames[], sourceMapped, raw }`. Modeled it (added `StepErrorTrace` + `StepErrorFrame`); `StepError.trace` is now that structured object, decoded faithfully (also accepts a bare stack string → `{frames:[], sourceMapped:false, raw}`). No more blank-panel drop.

**#5 — phantom `dispatchId` (fixed).** Dropped it (no wire source); `DispatchRef` is now `{ childExecutionId, targetType, correlationId, status }` with `childExecutionId` mapped from the ledger's `target_id`.

**#3 / #4 — list not tree-aware / blank child names (tracked, server-side).** These need engine fields (`rootExecutionId` on list rows; `name` on child edges) — can't be honestly fixed SDK-side. I kept the graceful degrade but documented it explicitly on `ExecutionRef` (traceRoot→own id, name→"") and in the changeset rather than pretending it works. I'll file the engine follow-ups under the S5 epic. Also passed step `logs` through as the wire array instead of silently coercing to `null` (same class as #2).

**Fixtures (addressed).** Swapped the hand-authored fixture for one derived from the authoritative engine `ExecutionDetailDto` wire shape (structured trace, `target_id` dispatch edges, cost-agnostic detail) and **removed the circular `decode(fixture) toEqual fixture`** in favor of explicit assertions against real-wire values — which is what caught #1/#2/#5. A live prod capture can't ship in a public repo, so I cited the DTO as the code-authoritative stand-in and noted provenance in the spec header. Added `execution-projection.with-cost.fixture.json` for the never-collapse path.

**Breaking change / versioning (addressed).** Added a changeset: `@sapiom/orchestration-core` minor (0.x), cli/mcp patch, documenting the `inspect()`/`listExecutions()` return-shape change and the `StepError.trace` type change. Agree the deployed API must emit the new fields before the degraded path stops being the default — that's the lockstep-pin dependency (~SAP-1018), now called out in the changeset.

**decode exports (addressed).** Narrowed the public surface: only `decodeExecutionProjection` is re-exported (the reusable entry point tickets 2/4 need to re-decode after an SSE refetch); `decodeExecutionRef`/`decodeCostNode` are now module-internal.

Build/typecheck/lint/test green across orchestration-core (71), cli (19), mcp (68).

positiveblue (approved):

PR #165, opened 2026-07-07
feat(orchestration-core): watchExecution() SSE async-iterator + live waitForExecution (SAP-1342)

gwitwer:
## Summary

Adds a live streaming helper so the SDK/CLI reflect **live** run state instead of polling on a timer. Resolves [SAP-1342](https://linear.app/sapiom/issue/SAP-1342).

- **`watchExecution({ executionId }, client) -> AsyncIterable<SseEvent>`** (new `src/watch.ts`) over `GET /v1/workflows/executions/:id/stream` (run + subtree). `SseEvent` is **IDs only** (`{ type, executionId, traceRoot, nodeId }`) — no payload; consumers refetch the `ExecutionProjection` via `inspect()`, so SSE can never drift from the REST source of truth (Module A invariant).
- Mirrors the `@sapiom/sandbox` `execStream()` reader pattern (`fetch().body.getReader()` + `TextDecoder` + `async function*`), adapted to SSE framing (`data:`/`event:`/`id:` fields, `Last-Event-ID` resume). Heartbeat and malformed frames are filtered.
- **Clean teardown:** iterator `return`/`break` runs the generator `finally`, which aborts the underlying `fetch` via an `AbortController` — no leaked connections.
- **`GatewayClient.openStream()`** centralizes the `x-api-key` SSE handshake (engine sits behind the service-key proxy; the SDK just presents its key) and maps failures to the same `OrchestrationError` shape as `request()`.
- **`waitForExecution()` is now SSE-driven:** it wakes on each event and refetches `inspect()`, bounded by aborting the stream at the deadline (heartbeats never wake it). On **any** SSE failure/drop it reverts to the original poll loop (×1.5 backoff, capped 5s) for the remaining budget — **no functional regression**.

## Interface contract

Implements the `watchExecution` / `SseEvent` signatures from the live-observable-execution `interfaces.md`. Consumes the Module A SSE channel (SAP-1139, Done) and `inspect()` (SAP-1341, merged #160). Consumed by CLI `--follow` (SAP-1343) and the tools watch surface (SAP-1344).

## Acceptance criteria

- [x] `watchExecution()` yields live `SseEvent`s from the run stream; iterator teardown aborts the connection.
- [x] `waitForExecution()` reflects live state via SSE and reverts to polling on SSE drop with no regression (all pre-existing wait tests unchanged and green).
- [x] `SseEvent` carries IDs only; live consumers refetch the projection.
- [x] No internal hostnames/customer data in examples (public repo — tests use `example.com`).

## Verification

`pnpm build && pnpm typecheck && pnpm lint && pnpm test` all green in `packages/orchestration-core` — **88 tests pass**, including new coverage for: yielded events, frame reassembly across read chunks, heartbeat filtering, teardown-aborts-fetch, handshake error; and `waitForExecution` live-path, SSE-drop fallback, stream-ended fallback, iterator teardown on settle.

### DIFF
--- a/packages/orchestration-core/src/client.ts
+++ b/packages/orchestration-core/src/client.ts
@@ -76,6 +76,62 @@ export class GatewayClient {
   post<T = unknown>(path: string, body?: unknown): Promise<T> {
     return this.request<T>('POST', path, body);
   }
+
+  /**
+   * Open a Server-Sent Events stream and return the raw {@link Response} so the
+   * caller can read `body` as it arrives (see `watchExecution`). Auth is the same
+   * `x-api-key` presented on every request — the engine sits behind the
+   * service-key proxy, so the SDK just presents its key. Handshake failures map
+   * to the same `OrchestrationError` shape as {@link request} (never a bare
+   * fetch rejection or a non-ok Response the caller has to re-inspect).
+   *
+   * The body is NOT consumed here: on success the live stream is handed back
+   * open. `signal` lets the caller abort the connection (iterator teardown);
+   * `lastEventId` is forwarded as the resume cursor (`Last-Event-ID`).
+   */
+  async openStream(
+    path: string,
+    opts: { signal?: AbortSignal; lastEventId?: string } = {},
+  ): Promise<Response> {
+    const headers: Record<string, string> = {
+      'x-api-key': this.apiKey,
+      accept: 'text/event-stream',
+    };
+    if (opts.lastEventId) {
+      headers['last-event-id'] = opts.lastEventId;
+    }
+
+    let res: Response;
+    try {
+      res = await fetch(`${this.base}${path}`, { method: 'GET', headers, signal: opts.signal });
+    } catch (err) {
+      throw new OrchestrationError({
+        code: 'NETWORK',
+        message: `Could not reach ${this.base}.`,
+        hint: err instanceof Error ? err.message : String(err),
+      });
+    }
+
+    if (!res.ok) {
+      const text = await res.text().catch(() => '');
+      const data = text ? safeParse(text) : undefined;
+      throw new OrchestrationError({
+        code: `HTTP_${res.status}`,
+        message: messageFrom(data) ?? `Stream request failed (${res.status} ${res.statusText}).`,
+        hint:
+          res.status === 401 || res.status === 403
+            ? 'Check your API key (`sapiom login` or SAPIOM_API_KEY) and that it has access to this orchestration.'
+            : undefined,
+      });
+    }
+    if (!res.body) {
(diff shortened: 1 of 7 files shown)


gwitwer (commented):
**Reviewed in depth — LGTM, no correctness blockers.** This is a careful, well-tested implementation of the live SSE path (SAP-1342). I traced the streaming, teardown, and fallback logic specifically:

**Correctness — the things that usually break in SSE iterators, all handled:**
- **Teardown aborts the fetch.** `watchExecution` owns an `AbortController`; both iterator `return()`/`break` (generator `finally`) and an external `signal` abort it, and the external-abort listener is removed on teardown (no listener accumulation on a long-lived caller signal). Verified in both the settle path (generator torn down at the `yield`) and the deadline path (`abort()` → pending `reader.read()` rejects → thrown → caught).
- **Heartbeats don't cause spurious refetches.** `parseSseFrame` drops heartbeat/comment/`retry:` frames before yielding, so `waitForExecution`'s `events.next()` only resolves on real events — the "heartbeats never wake it" claim holds.
- **Unknown/foreign/malformed frames are dropped**, not trusted — `parseSseEvent` validates `type ∈ SSE_EVENT_TYPES` and a non-empty `executionId`, defaults `traceRoot`/`nodeId` to null. Good defense for a fire-and-forget channel.
- **Frame reassembly** across read chunks + multi-line `data:` + CRLF/LF boundaries all handled (`indexOfFrameBoundary`), and covered by tests.
- **Fallback is complete:** stream-ends (`next.done`) → poll; SSE throws → `catch` → poll; deadline → `abort()` → poll → immediate timeout. Every exit runs `finally` (clearTimeout + `events.return()`). The refactor into `settledResult()` keeps the live and poll paths evaluating identically — no behavior drift.
- `openStream` maps handshake failures to the same `OrchestrationError` shape as `request()`, with a 401/403 auth hint; no token leak in errors.

**One test-hygiene note (non-blocking):** the pre-existing `waitForExecution` tests (`polls until terminal`, the auto-resume pauses, `needs-signal`, `timeout`) don't inject `watch`, so they now default to the real `watchExecution` — which only falls back to polling because the hand-rolled fake client has no `openStream` method (it throws, gets caught). So they pass, but they're implicitly testing the SSE-drop→poll path via a missing method rather than the pure poll path they were written for, and they'd silently change behavior if `openStream` were ever added to the fake. Consider passing an explicit `watch: async function*(){}` (empty → stream-ends → poll) to those cases to make the poll path deterministic and intent-clear. The new tests already drive the live/fallback paths explicitly and thoroughly (88 green).

Nothing here blocks merge.

positiveblue (approved):

PR #170, opened 2026-07-08
feat(harness): telemetry pipeline — hooks, ingest, local store, collector batcher

gwitwer:
## Summary

Analytics side of the Sapiom Harness: Claude Code hooks → local ndjson store → batched forward to a remote collector, opt-in gated.

- `core/inject/claude-settings.ts` — `generateClaudeSettings()` writes a per-session `settings.json` registering all six Claude Code hooks (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop, SessionEnd), each pointed at a generated `emit.cjs`.
- `emit.cjs` (template string, zero deps, CommonJS, node builtins only) — reads the hook's stdin JSON, POSTs `{hookEvent, receivedAt, harnessSessionId, payload}` to `SAPIOM_HARNESS_INGEST_URL` with a bearer token, 200ms abort timeout, swallows every error, always exits 0. A dead/slow harness server can never slow down or break the agent.
- `core/collector/normalizer.ts` — maps hook payloads to the normalized `AnalyticsEvent` shape (`session.start`, `prompt.submitted`, `tool.call`, `turn.completed`, `session.end`); `PreToolUse` stays registered but has no dedicated event yet. Captures the Claude session id off `SessionStart` so it can be linked to the harness session.
- `core/collector/transcript.ts` — tail-reads the session transcript JSONL (capped ~2MB from the end) to backfill `turn.completed` with the last assistant message text, model, and token usage.
- `core/collector/store.ts` — append-only ndjson writer (crash-safe `fs.appendFile`, `mkdir -p` on first write).
- `core/collector/batcher.ts` — in-memory queue → POST batch every 5s or 50 events, 3 retries with exponential backoff then drop, hard cap on queue size so a long outage can't grow memory unbounded, gated by a `telemetryOptIn` flag, no-ops (with a debug log) when no collector URL is configured.
- `server/ingest.ts` — `createIngestRouter(deps)`, a self-contained Express router for `POST /ingest`: bearer-token check, 1MB body limit, responds 200 immediately and processes async, calls back into the session registry when `session.start` reveals the agent's session id.
- `scripts/mock-collector.mjs` — standalone (zero deps on the package) mock remote collector: listens on `:4199`, logs a one-line batch summary, appends raw bodies to `./mock-collector-received.ndjson`.

## Test plan

- [x] `pnpm --filter @sapiom/harness test` — 37 tests passing across normalizer, transcript tail-reading, ndjson store, batcher (opt-out/no-URL/retry-backoff/close-flush), settings generation, and the ingest router (auth, unknown-session drop, PreToolUse drop, transcript enrichment, session-id linking).
- [x] `pnpm --filter @sapiom/harness typecheck` — clean.
- [x] `pnpm --filter @sapiom/harness build:server` — clean.
- [x] `pnpm --filter @sapiom/harness sim:e2e` — full end-to-end simulation **without running `claude`**: generates a real `emit.cjs`, pipes fixture `SessionStart` + `UserPromptSubmit` hook JSON into it as separate child processes (exactly how Claude Code invokes hooks), against a real HTTP server mounting `createIngestRouter`, and a real `scripts/mock-collector.mjs` child process as the remote. Confirms both events land in `events.ndjson`, the session registry gets linked to the agent session id, and the mock collector receives a batch containing both events. All assertions passed:
  ```
  ok - exactly 2 events appended to events.ndjson
  ok - first event is session.start
  ok - session.start captured the agent session id
  ok - second event is prompt.submitted
  ok - prompt text made it through the pipeline
  ok - session registry callback linked agentSessionId
  ok - mock-collector received the right machineId
  ok - mock-collector received both events in one batch
  ok - mock-collector logged a batch summary

  PASS — analytics pipeline e2e simulation succeeded
  ```

## Integration notes for whoever wires the main server

- Mount: `app.use(createIngestRouter(deps))` — `deps` needs `ingestToken` (the per-boot secret), `resolveSession` (harnessSessionId → `{harness, userId, machineId, agentSessionId}` from the session registry), `onAgentSessionResolved` (registry write-back), `store` (`createEventStore()`), `batcher` (a `CollectorBatcher` instance), and `enrichFromTranscript: enrichTurnCompleted`.
- When spawning `claude`, pass `--settings <settingsPath>` from `generateClaudeSettings({ harnessSessionId })`, and set `SAPIOM_HARNESS_INGEST_URL` / `SAPIOM_HARNESS_INGEST_TOKEN` / `SAPIOM_HARNESS_SESSION_ID` in the child's env (per `shared/types.ts`'s `ENV` map) — `emit.cjs` reads those from its own `process.env`, inherited from the `claude` process.
- `CollectorBatcher` needs `setTelemetryOptIn()` called whenever `PATCH /api/settings` changes `telemetryOptIn`.
- Local dev "remote": `pnpm --filter @sapiom/harness mock-collector` (or `node scripts/mock-collector.mjs`), then point `SAPIOM_COLLECTOR_URL=http://localhost:4199`.

No changes to `shared/types.ts` — built entirely against the existing contract. No new dependencies were needed.


gwitwer:
Updated for the telemetry contract v1 delta (seq, tenantId, versioned collector envelope):

- `AnalyticsEvent` now carries `seq` (per-`harnessSessionId` monotonic counter, assigned server-side in `server/ingest.ts` via a new `core/collector/seq.ts` — never trusted from the hook script) and `tenantId` (sourced from the same server-side identity context as `userId`; `IngestSessionContext` gained a `tenantId` field).
- `CollectorBatcher` now builds the versioned envelope (`{batchId, schemaVersion, machineId, sentAt, context, events}`, `context: CollectorContext` passed into the constructor) and POSTs to `${collectorUrl}/v1/harness/events`.
- Retry semantics now branch on status: any 2xx = accepted, 4xx = drop without retry (permanent), 5xx/network = retry 3x with exponential backoff then drop.
- Added `CollectorBatcher.setApiKey()` / constructor `apiKey` option — sends `Authorization: Bearer <key>` once a user is logged in, omitted for anonymous traffic.
- `toolResponseSummary` truncation cap raised to ~16KB (was the shared 4KB default; `toolInput` stays at 4KB).
- `turn.completed` payload field renamed `lastAssistantText` → `assistantText` to match the documented payload shape.
- `scripts/mock-collector.mjs` now serves `POST /v1/harness/events` and returns `202 {"accepted": n}` (was a bare `200 {"ok": true}` on any path).

Re-verified after merging the latest `feat/harness` (W2 SPA + W4 CLI landed): `pnpm --filter @sapiom/harness test` — 74/74 passing (44 mine + 30 from W2/W4), `typecheck` and `build:server` clean (across the full monorepo build, since `packages/mcp` needed its own deps built first — unrelated to this PR), and `pnpm sim:e2e` still passes end-to-end with the new envelope, seq assignment, and tenantId propagation verified against a live `mock-collector.mjs` process.

No further contract-change requests — built entirely against `shared/types.ts` as merged.

PR #175, opened 2026-07-08
feat(analytics-core): zero-dep usage analytics emitter (identity, consent, batching)

evtran0209:
## Summary

New package `@sapiom/analytics-core`: the walking skeleton of the shared usage-analytics emitter for Sapiom SDK packages — `createAnalytics()` → `track()/flush()/shutdown()` with persisted anonymous identity, layered consent, batched delivery, and hard never-throw/never-block guarantees. Zero runtime dependencies (Node built-ins only). No existing package is touched; the only out-of-package change is the pnpm-lock workspace importer entry.

Fixes SAP-1406 — https://linear.app/sapiom/issue/SAP-1406

## Changes

- **Emitter** (`analytics.ts`): `createAnalytics(config)` returns `{ track, flush, shutdown, enabled, anonymousId, sessionId }`. `track()` is a synchronous enqueue that never throws; `flush()`/`shutdown()` never reject; every public entry is wrapped, with an optional `debug` hook for observing internal failures. `sessionId` is one uuid4 per process.
- **Consent** (`consent.ts`): precedence `disabled: true` → `SAPIOM_TELEMETRY_DISABLED=1` → `DO_NOT_TRACK=1` (both accept `1`/`true`) → injected `consentProvider` → default on. Resolved once and gating at the `track()` entry: disabled means zero disk writes and zero network calls.
- **Identity** (`identity.ts`): `~/.sapiom/analytics.json`, mode `0600`, created lazily on first use; corrupt files silently regenerated; unwritable HOME degrades to `anonymous_id: null` without breaking emission. Holds the anonymous machine uuid4 + first-run-notice marker; a one-line stderr notice prints exactly once per machine across all packages.
- **Batching** (`batch-queue.ts`): flush every 3s, at 20 events, or best-effort on `beforeExit`; all timers `unref()`ed so the queue never holds the process open; at most one jittered retry per batch, then silent drop.
- **Delivery** (`http-sender.ts`): native `fetch` POST `{events: [...]}` to the public collector, optional `x-sapiom-api-key` header, request timeout so a hung collector can never wedge a flush; 429/5xx/network → single retry, other 4xx → immediate drop.
- **Envelope** (`envelope.ts`, `data.ts`): `event_id`/`anonymous_id`/`session_id`/`event_timestamp`/`source`/`event_type`/`sdk_name`/`sdk_version`/`schema_version: "1"`/`data`, `user_id` only when provided; per-event overrides restricted to known envelope keys; per-field ~16KB cap with `data._truncated: true`, unserializable values (circular/BigInt) replaced instead of thrown on.
- Package mechanics copy `packages/tools` (dual CJS/ESM build, jest + ts-jest, eslint, MIT), README stub with the consent/opt-out summary, and a `minor` changeset for the new package.

## Test plan

48 tests across 5 suites, isolated in a temp HOME (with a tripwire that fails any test escaping the sandbox) and injected fetch — no real collector traffic except one deliberate real-`fetch` ECONNREFUSED test against a closed local port.

- Consent: each env opt-out and programmatic `disabled` → zero fetch calls, no identity file; provider precedence; hostile provider tolerated
- Identity: file created lazily with mode 0600, reused across instances, corrupt/wrong-shape regenerated; notice printed exactly once across two instances; unwritable HOME keeps emitting with `anonymous_id: null`
- Batching: 20-event immediate trigger, 3s timer (fake timers), `beforeExit` best-effort flush, listener removed on shutdown, idempotent shutdown
- Fault injection: collector down (rejecting fetch + real closed port), slow collector (hot path stays synchronous), 500/429 → exactly one retry then silent drop, 400 → no retry; endpoint resolution and api-key header

```
pnpm --filter @sapiom/analytics-core build   # exit 0 (cjs + esm + esm-pkg)
pnpm --filter @sapiom/analytics-core test
  Test Suites: 5 passed, 5 total
  Tests:       48 passed, 48 total

pnpm build && pnpm typecheck && pnpm lint    # repo-wide, all exit 0
pnpm test                                    # repo-wide, exit 0
  packages/core:           132 passed | packages/tools: 322 passed
  packages/analytics-core:  48 passed | all other packages green
```

### DIFF
--- a/.changeset/analytics-core-walking-skeleton.md
+++ b/.changeset/analytics-core-walking-skeleton.md
@@ -1,0 +1,10 @@
+---
+"@sapiom/analytics-core": minor
+---
+
+Introduce `@sapiom/analytics-core` — a zero-dependency usage analytics emitter shared by Sapiom SDK packages.
+
+- `createAnalytics(config)` returns `{ track, flush, shutdown, enabled, anonymousId, sessionId }`; `track()` is a synchronous enqueue that never throws, and `flush()`/`shutdown()` never reject.
+- Consent precedence: programmatic `disabled: true` → `SAPIOM_TELEMETRY_DISABLED=1` → `DO_NOT_TRACK=1` → injected consent provider → default on. When disabled, nothing is written or sent.
+- Anonymous machine identity persisted at `~/.sapiom/analytics.json` (mode 0600, created lazily, corrupt files silently regenerated), plus a one-time first-run notice on stderr.
+- Batched delivery: flush every 3s, at 20 events, or best-effort on process exit; at most one retry per batch (jittered), then silent drop; per-field ~16KB cap flagged via `data._truncated`.
--- a/packages/analytics-core/.eslintrc.js
+++ b/packages/analytics-core/.eslintrc.js
@@ -1,0 +1,23 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: 'tsconfig.json',
+    tsconfigRootDir: __dirname,
+    sourceType: 'module',
+  },
+  plugins: ['@typescript-eslint'],
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
+  root: true,
+  env: {
+    node: true,
+    jest: true,
+  },
+  ignorePatterns: ['.eslintrc.js', 'dist', 'node_modules'],
+  rules: {
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_' }],
+  },
+};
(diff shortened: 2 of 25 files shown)


evtran0209:
Internal claude review pass complete (this repo has no CI reviewer round). **Verdict: MERGE-READY — zero blockers.**

Strengths verified line-by-line: fail-closed consent resolution (resolution error → disabled, stronger than spec); allow-listed per-event overrides (server-owned fields like org identity cannot be injected client-side); identity store 0600 on create AND rewrite with sticky degradation to anonymous null; retry-outcome mapping exactly per contract (429/5xx/network → single retry, other 4xx → drop); all timers unref'd; every public entry never-throw-wrapped including the debug hook itself.

Notes (no change requested):
1. Consent freezes per instance — dynamic consent (harness UI toggles) must gate at the harness server layer or recreate the instance. That's the intended design; the harness-server ticket will carry it.
2. Two racing first-ever processes could both print the first-run notice (read-check-write, no lock) — best-effort is acceptable; file locking would be over-engineering.
3. Oversized non-string data fields truncate to JSON strings (type morph) — acceptable and flagged by `data._truncated`.

gwitwer (approved):

PR #208, opened 2026-07-08
feat(harness): canvas style contract for generated workflow visualizations

gwitwer:
## Summary
- New `CANVAS_STYLE_GUIDELINES` constant (`src/profiles/canvas-guidelines.ts`), ~19 lines of guideline text (well under the ~30 line prompt-sized budget): self-contained HTML only (no external stylesheets/scripts/CDN fonts, no build step), the harness's actual dark-theme palette hexes (cribbed from `web/src/styles.css`'s `[data-theme="dark"]` tokens — `#0f0f0f` bg / `#1a1a1a` panel / `#2e2e2e` border / `#fafafa` text / `#6be195` accent-success / `#f59e0b` escalation / `#f87171` failure), monospace typography, workflow-graph conventions (rounded nodes, curved arrowed edges, subtle glow), color-coded terminal outcomes, a stats-header pattern, a legend-footer pattern, and a "prefer clarity over decoration" closing note.
- Modeled on what `scripts/seed-example.mjs`'s reference canvas already renders structurally (stats row, rounded SVG nodes with arrowhead markers, terminal-outcome coloring, legend footer) — didn't touch that seed script itself, only described its pattern.
- **Note on colors:** the seed script's own hardcoded palette (`#0b0e14` bg, `#5b8cff` accent, etc.) predates the current design-token pass and doesn't exactly match `web/src/styles.css`'s current tokens. Per the brief, I cribbed the *exact* hexes from `styles.css` (the live, current palette) rather than the seed's own values, and kept the seed's *structural* conventions. Flagging this discrepancy in case the seed script itself is worth a follow-up refresh.
- Injected the guidelines in both surfaces: appended to the visualize macro's inject text (`src/core/macros.ts`) after the existing `{{subject}}` line, and embedded in the system prompt's existing "Canvas convention" paragraph (`src/profiles/default.ts`), so both the one-click macro and freeform "show me X on the canvas" asks carry the same contract.

## Test plan
- [x] `pnpm --filter @sapiom/harness test` — 275/275, including a new `canvas-guidelines.test.ts` (guidelines are prompt-sized, carry the exact palette hexes, cover node/edge/legend/stats/clarity conventions, and are embedded verbatim in both the system prompt and the visualize macro's text) and an updated `server/macros.test.ts` assertion (its exact-match check on the visualize macro's injected text was intentionally outdated by this change)
- [x] `pnpm --filter @sapiom/harness lint` — clean
- [x] `pnpm --filter @sapiom/harness build` — clean
- [x] `pnpm --filter @sapiom/harness e2e:live` — macro phase (`POST /api/macros/visualize/run`) still green
- [x] `pnpm --filter @sapiom/harness test:ui` — 25/25 Playwright

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/harness/scripts/e2e-live.ts
+++ b/packages/harness/scripts/e2e-live.ts
@@ -400,16 +400,6 @@ async function testCoreFlow(): Promise<void> {
     assert(canvasRes.status === 200, "GET /canvas/:id/ serves the written index.html");
     assert((await canvasRes.text()).includes("e2e"), "served canvas content matches what was written");
 
-    // --- 10. run the visualize macro (inject kind) — proves the macro engine reaches the pty ---
-    const macroRes = await fetch(`${baseUrl}/api/macros/visualize/run`, {
-      method: "POST",
-      headers,
-      body: JSON.stringify({ harnessSessionId: sessionId, subject: "the e2e proof" }),
-    });
-    assert(macroRes.status === 200, "POST /api/macros/visualize/run returns 200");
-    const macroBody = (await macroRes.json()) as { ok: boolean };
-    assert(macroBody.ok === true, "macro run responds { ok: true }");
-
     // --- 11. the launch directory's sapiom.json (written before startServer) was scanned at boot ---
     // Note: WorkflowInfo.name comes from package.json (or the directory's own
     // basename) — the sapiom.json marker itself only carries definitionId —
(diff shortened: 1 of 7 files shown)


gwitwer:
**Amendment (per Dave, via team lead):** visualize is now a one-click render/re-render command with no free-text subject — dropped `{{subject}}` entirely and flipped `requiresWorkflow: true`. The inject text is now a complete, self-sufficient prompt: render the bound workflow at `{{workflow.path}}` (steps, control flow, and interconnections — see `.sapiom/harness-context.json` for the full workflow list) to `{{canvas.path}}`, following the style contract.

- w2 is removing the client-side subject-entry dialog separately.
- w3 is enriching `harness-context.json` with the full workflow list so "how they interconnect" has real data.
- **Flagging for follow-up:** `web/src/lib/mock-data.ts`'s `MOCK_MACROS` still has the old `{{subject}}` text and `requiresWorkflow: false` for visualize — that's mock-mode-only data, unaffected by this backend change, but should get the same update once the subject dialog is removed client-side.
- Reordered the `e2e-live.ts` macro-run assertion to run after the workflow-bind step (visualize now needs one) — this also mirrors the real one-click flow more closely and reuses the existing bound-workflow fallback.

Full bar re-verified on top of the amendment and after merging forward past #210 (action-rail retirement, unrelated/no conflicts): typecheck (server + strict web), 276 unit tests, lint, build, e2e:live, test:ui (30/30).

PR #220, opened 2026-07-08
feat(harness): canvas kit — LLM writes data, not HTML

gwitwer:
## Summary

One-click Visualize was slow (~55s, Claude writing ~11KB of hand-rolled HTML per run) and always dark regardless of the app's theme, because the LLM was doing presentation work every single time. This replaces that with a **canvas kit**: a prewritten, self-contained template (both light and dark palettes baked in, a vanilla-JS renderer) that the agent only ever feeds a small JSON data block.

- **`src/core/canvas-template.ts`** (new): `renderCanvasHtml(data)` is the single source of truth for the document. Theme is read client-side from `?theme=light|dark` with a `prefers-color-scheme` fallback — this turned out to already match exactly what PR #217 (canvas-follows-app-theme, merged into this branch) wired the iframe to pass, so no coordination gap. The renderer computes its own layered top-to-bottom layout from edge topology (robust to diamonds/reconverging branches, defensively terminates on a cycle), and follows the exact conventions `profiles/canvas-guidelines.ts` already documents for the freeform path: straight sequential / curved branching edges colored by the target node's outcome, a dashed border for `pause` nodes, a **uniform glow on every node** (not just entry/terminal), and a legend with a visually distinct marker per kind. `ensureCanvasTemplate(cwd)` is backfill-only — never clobbers a file an earlier session or an agent already populated.
- **`SessionManager`** gained an `ensureCanvasTemplate` hook (mirrors the existing `writeWorkspaceContext` pattern), called from both `create()` and `resume()`, so the canvas pane is never a blank iframe.
- **The visualize macro is unbound-friendly again** (`requiresWorkflow: false`): its prompt no longer references `{{workflow.path}}` at all — it tells the agent to read `harness-context.json` and either represent the bound workflow or the whole workspace's interconnections, editing **only** the JSON data block and leaving the rest of the file byte-identical. `CANVAS_STYLE_GUIDELINES` now only rides in the system prompt's freeform path (custom canvas requests the template's model can't represent); the system prompt's canvas paragraph now explicitly distinguishes the two.
- **`scripts/seed-example.mjs`** now renders through the exact same `renderCanvasHtml()`, prefilled with order-triage's real step graph, instead of its own hand-rolled HTML — fixes the palette drift from the style contract I'd flagged in an earlier PR, and means the seed and the kit can never diverge from each other again.
- **New `e2e/` Playwright suite** (`pnpm test:canvas`) — headless, no dev server needed since the template is fully self-contained (`page.setContent()` renders it directly). Screenshots both themes, and proves the exact contract the macro's prompt relies on: editing only the raw JSON substring between the `canvas-data` script tags via literal string surgery, leaving every other byte untouched, still renders correctly.
- **`e2e-live.ts`**: asserts the template is already on disk the instant a session is created (before any visualize run), and that visualize succeeds both unbound (workspace-overview mode) and once a workflow is bound.

**Bug found and fixed via a real render, not just review**: edge labels on two branches diverging from the same node were both positioned at their path's geometric midpoint, which put them at nearly the same coordinates and made long labels visually overlap into one unreadable string. Fixed by anchoring each label near the branch point instead, offset and text-anchored toward its own destination — confirmed via screenshot before/after.

**Screenshots**: both themes render correctly and consistently (I inspected them directly — full page renders showing the order-triage workflow with the intake→classify→route branch-to-auto_resolve/escalate graph, an interconnections panel, and a legend with distinct markers per kind, in both dark and light). I can't attach the binary PNGs through the CLI without a hosting step, but they're fully reproducible locally via `pnpm --filter @sapiom/harness test:canvas` (outputs to `e2e/test-results/canvas-order-triage-{light,dark}.png`, gitignored) or `pnpm --filter @sapiom/harness seed-example <dir>` for the real seed output.

**Known follow-up, flagged for whoever owns `web/`**: `web/src/lib/mock-data.ts`'s `MOCK_MACROS` still carries the pre-kit visualize text and `requiresWorkflow: true` for mock mode — didn't touch it since it's outside my lane.

## Test plan
- [x] `pnpm --filter @sapiom/harness typecheck` (server + strict web)
- [x] `pnpm --filter @sapiom/harness test` — 322/322
- [x] `pnpm --filter @sapiom/harness lint` — clean
- [x] `pnpm --filter @sapiom/harness build` — clean
- [x] `pnpm --filter @sapiom/harness e2e:live` — new assertions: canvas template exists at session create, visualize succeeds unbound and bound
- [x] `pnpm --filter @sapiom/harness test:ui` — 37/37 (confirms compatibility with #217's theme passthrough)
- [x] `pnpm --filter @sapiom/harness test:canvas` — 5/5 new headless render tests, both themes
- [x] Ran `seed-example.mjs` end to end against a scratch dir and visually inspected the real output

**Unrelated flake spotted during verification (not caused by this PR):** intermittent `ENOENT` on `workspace-context.ts`'s atomic rename, already tracked as a P0 elsewhere — reproduced identically with none of this PR's code in the stack trace.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/harness/e2e/.gitignore
+++ b/packages/harness/e2e/.gitignore
@@ -1,0 +1,2 @@
+playwright-report/
+test-results/
--- a/packages/harness/e2e/canvas-template.spec.ts
+++ b/packages/harness/e2e/canvas-template.spec.ts
@@ -1,61 +1,143 @@
 import { test, expect } from "@playwright/test";
-import { renderCanvasHtml, type CanvasData } from "../src/core/canvas-template.js";
+import { TEMPLATE_HTML, renderCanvasDocument } from "../src/core/canvas-template.js";
 
-const ORDER_TRIAGE: CanvasData = {
-  version: 1,
-  graphs: [
-    {
-      id: "order-triage",
-      title: "order-triage",
-      subtitle: "Support-ticket triage: intake -> classify -> route, then auto-resolve or escalate.",
-      badges: ["standalone workflow"],
-      stats: [
-        { label: "steps", value: 5 },
-        { label: "terminal outcomes", value: 2 },
-        { label: "branch points", value: 1 },
-      ],
-      nodes: [
-        { id: "intake", kind: "entry", label: "intake", sublabel: "receive + log order" },
-        { id: "classify", kind: "step", label: "classify", sublabel: 'category ?? "general"' },
-        { id: "review", kind: "pause", label: "review", sublabel: "waits for a human tag" },
-        { id: "route", kind: "step", label: "route", sublabel: "branch on category" },
-        { id: "auto_resolve", kind: "terminal-success", label: "auto_resolve", sublabel: "terminate({resolved:true})" },
-        { id: "escalate", kind: "terminal-warn", label: "escalate", sublabel: "terminate({escalated:true})" },
-      ],
-      edges: [
-        { from: "intake", to: "classify", kind: "sequential" },
-        { from: "classify", to: "review", kind: "sequential" },
-        { from: "review", to: "route", kind: "sequential" },
-        { from: "route", to: "auto_resolve", kind: "branching", label: "category != billing" },
-        { from: "route", to: "escalate", kind: "branching", label: "billing_dispute" },
-      ],
-    },
-  ],
-  interconnections: [
-    { from: "external", to: "order-triage.intake", kind: "signal", label: "an order object enters here" },
-    { from: "order-triage.escalate", to: "external", kind: "handoff", label: "routes to a human out-of-band" },
-  ],
-  note: "Static preview — regenerate after the workflow changes.",
-};
+/**
+ * A representative filled-in canvas body — the shape an agent following the
+ * visualize macro's prompt produces after cloning `_template.html` and
+ * using its documented node/edge patterns. Kept independent of
+ * scripts/seed-example.mjs (which has scaffold side effects on import) but
+ * intentionally mirrors the same order-triage workflow for consistency.
+ */
+const ORDER_TRIAGE_BODY = `
+<section class="canvas-panel">
+  <header class="canvas-header">
+    <div class="canvas-title-row">
+      <h1 class="canvas-title">order-triage</h1>
+      <span class="canvas-badge">standalone workflow</span>
+    </div>
+    <p class="canvas-subtitle">Support-ticket triage: intake -&gt; classify -&gt; route, then auto-resolve or escalate.</p>
+    <div class="canvas-stats">
+      <div class="canvas-stat"><span class="canvas-stat-value">6</span><span class="canvas-stat-label">steps</span></div>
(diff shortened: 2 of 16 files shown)


gwitwer:
**Course correction (per Dave, via team lead):** the JSON-data + JS-renderer approach in the original commit is out — Dave wants the middle ground: a prewritten **template the agent clones and fills with real markup**, not a data schema a renderer interprets. Pushed a full rewrite as a second commit.

What changed:
- `core/canvas-template.ts` is rewritten: `renderCanvasDocument(bodyHtml)` is now just the shared shell (CSS + the `?theme=` switch script) — no JSON, no layout-computing renderer. The template ships one example of every node kind (`node--entry/step/pause/terminal-success/terminal-warn`) and edge kind (base `.canvas-edge` = sequential straight; `--success`/`--warn` = branching curved, colored by outcome; `--cross` = cross-workflow signal/handoff, dashed) inside an inert `<template id="canvas-patterns">` block — real parsed DOM, never rendered (verified: `.canvas-node` count is 0 on the live page, only reachable via `.content`), and deliberately not an HTML comment, since the double-dash BEM class names would risk being misread as a comment terminator.
- `ensureCanvasTemplate(cwd)` now backfills **two** files: a pristine `_template.html` (written once, clone source) and the live `index.html` (same initial content). Both independently backfill-only.
- The visualize macro's prompt: "clone `_template.html` to `index.html`, keep the `<style>` block and structural/pattern classes untouched, do not write new CSS, fill in the content using the template's patterns." Still `requiresWorkflow: false` and still avoids `{{workflow.path}}` entirely, so the same static prompt works bound or unbound.
- `scripts/seed-example.mjs`'s order-triage canvas is now hand-authored markup (exactly what an agent following the macro's prompt would produce) through the same `renderCanvasDocument()` shell — writes both files, matching the real pairing.
- `e2e/canvas-template.spec.ts` rewritten for the new shell: screenshots a filled instance in both themes (attached below), a multi-graph workspace overview, and proves the template block is genuinely inert.
- `e2e-live.ts` updated for the two-file pairing, plus a fix for a pre-existing readiness-ordering gap I found while re-verifying (confirmed it predated my changes by reproducing it on the unmodified merge base) — PR #223 landed the identical fix independently upstream; resolved as a straightforward merge, and caught/fixed a duplicate-declaration mess the auto-merge left behind.

**Screenshots** (both themes, order-triage with the full node-kind set including `pause`):
- I inspected these directly during verification; both render correctly and consistently, matching the palette, universal glow, straight/curved/dashed edge conventions, and distinct legend markers from the earlier style-contract work.

Full bar green on the merged state: typecheck (server + strict web), 352 unit tests, lint, build, `e2e:live`, `test:ui` (38/38), `test:canvas` (5/5).

Known follow-up, still unaddressed and outside my lane: `web/src/lib/mock-data.ts`'s `MOCK_MACROS` visualize entry is stale again (references `{{workflow.path}}`, `requiresWorkflow: true` — predates even the earlier one-click amendment).

PR #231, opened 2026-07-08
fix(harness): persistent workflow actions panel, no more hover-expand

gwitwer:
## Summary

The docked workflow action strip had two problems the owner flagged:

1. **Bug**: the hover-expanded panel floated over the terminal as an absolutely-positioned overlay, and its background let terminal text bleed through — looked broken.
2. **UX reversal**: hiding actions behind hover-to-reveal made Visualize (and everything else) feel "removed." The ask was a persistently visible panel.

(A smaller opacity-only patch, #233, briefly landed on `feat/harness` as an interim fix for problem 1 alone — it's superseded here since the persistent column structurally can't bleed the way the old floating overlay did.)

## What changed

- Replaced `WorkflowActionStrip` (icon-only at rest, hover/`:focus-within`-expands to a floating panel) with `WorkflowActionsPanel` — a real fixed-width (168px) grid column between the rail and the terminal, not an overlay. It can't ever sit on top of or bleed into the terminal at any window width, because it always takes its own layout space.
- Icon + label are always visible for every action, stacked vertically. Disabled actions show their gating reason (e.g. "Not deployed yet") inline as a caption, always — not only on hover. Gating logic itself (`macroDisabledReason`) is unchanged.
- **Dropped the row-anchoring/notch trick** (the panel used to slide to vertically align with the selected rail row, with a border "notch" fusing the two). A full-height persistent column can't sensibly float at a row's y-offset, so per the owner's fallback guidance it's replaced with a small header inside the panel: "Actions · `<workflow name>`". Selecting a different workflow just relabels the header — the panel itself never moves or resizes.
- Deleted now-dead code: the old strip component, the hover-expand/notch CSS, the `useElementTopOffset` hook (only consumer was the strip), and the `onSelectedRowElement` ref plumbing through `App.tsx`/`WorkflowsRail`. Reverted the `.workflow-item.is-selected` negative-margin "bleed into the notch" trick since there's no notch to bleed into anymore.
- `.app`'s grid-template-columns swapped the old `32px` strip track for a named `ACTION_PANEL_WIDTH` (168px) constant in `use-pane-widths.ts` — still a fixed, non-draggable track; the rail↔terminal↔canvas resize handles are unaffected.
- Reused the existing design-fidelity tokens (`--accent-dim`, `--bg-hover`, `--focus-ring`, `--radius`, etc.) so the panel matches the rest of the app in both themes.

### Reconciled with #230 (deterministic canvas), merged after this branched

- #230 split the old single "Visualize" macro into two: `visualize` (deterministic, zero-LLM render — now `requiresWorkflow: false`, action kind `render-canvas`) and a new `ai-visualize` (the old LLM-authored fallback, also `requiresWorkflow: false`). The panel's macro list is generic over `MacroDef`, so both new buttons show up automatically — no per-kind branching needed in the component.
- Fixed a gap this exposed: the panel used to gate its *entire* macro list behind a truthy `workflow`, which was equivalent to the old per-macro gating when every macro required a workflow. Now that `visualize`/`ai-visualize` don't, the panel always renders the full macro list and lets `macroDisabledReason` gate each button individually (workflow-requiring macros show "Select a workflow first" when unselected; the two Visualize variants don't).
- Added `Wand2` to `Icon.tsx`'s curated icon map — `ai-visualize`'s icon was added by #230 but the icon component didn't have it yet, so it was silently falling back to the generic help-circle glyph.
- Removed the now-unused `.action-panel-empty` "select a workflow" empty state, since the panel no longer has a truly-empty state (the macro list itself is now the source of truth for what's actionable).

## Playwright

Rewrote the `docked workflow action strip` describe block (now `persistent workflow actions panel`) to match the new behavior, and added coverage for the post-#230 macro set:
- panel visible at rest with all labels, no hover/focus step
- both `macro-visualize` and `macro-ai-visualize` render and stay enabled (gated only on session presence, not workflow selection)
- disabled reason always visible inline
- panel doesn't move/resize on selection change, just relabels its header
- panel occupies its own column and never overlaps the terminal, even at a narrow (900px) viewport
- updated the narrow-viewport canvas-header test's math/viewport for the new (larger, since it's now a real persistent column) fixed-width floor

## Test plan

- [x] `pnpm --filter @sapiom/harness build` — green
- [x] `tsc --noEmit` (web) — green
- [x] `pnpm --filter @sapiom/harness test` — 405 passed
- [x] `pnpm --filter @sapiom/harness test:ui` — 45 passed
- [x] Verified locally: panel in light theme, dark theme, disabled-reason state, both Visualize buttons with distinct icons, and at a narrow viewport with no terminal overlap (screenshots reviewed, not committed — `web/e2e/screenshots/` is gitignored)


gwitwer:
Superseded — scope changed to a minimal opacity/z-index fix instead of restructuring into a persistent column. Closing this in favor of a smaller follow-up PR.

gwitwer:
Reopening — after further review, the persistent-column approach is the definitive fix (a real grid column can't bleed the way the transparent hover overlay did, and it matches the always-visible-actions ask). Reconciling with #230 (deterministic canvas), which merged after this branched and now conflicts.

PR #236, opened 2026-07-08
feat(harness): first-run welcome + onboarding experience

gwitwer:
## What

A real welcome experience for new harness users, on two surfaces:

### 1. First-run welcome panel (SPA)

On a machine that has never run the harness (no recent directories recorded before boot), the CLI now sets `AppState.firstRun` and skips the auto-created boot session — so instead of landing cold in a bare terminal, the user gets a welcome card in the terminal slot:

- one short paragraph on what the harness is,
- **Start a new project** → the existing new-session/directory-picker flow,
- **Run the sample project** → `POST /api/sample-project` seeds the bundled `order-triage` example into `~/.sapiom/harness/sample-project` (idempotent — an existing copy is reused, user edits intact) and opens a session bound to it; the seeded canvas renders immediately,
- a compact hint row (Visualize / Run local / Deploy chips + ⌘K) and a quiet "Skip for now".

Returning users never see it. Styled entirely from the existing component recipes (#228): badge chips, button variants, radius/shadow/focus-ring scale, light mode default — no new design system.

The seeding logic moved out of `scripts/seed-example.mjs` into a shared `src/core/example-seed.ts`; the demo script is now a thin force-reseed wrapper over the same module, so the in-app sample and the demo seed can't drift.

### 2. Agent first-reply orientation

The injected default profile (#192) now asks the agent's first reply to also orient the user — 2-4 sentences on what it can do here (visualize on the canvas, run locally, deploy) plus one concrete suggested first step picked from the workspace context (the bound/listed workflow by name — e.g. the seeded sample — or scaffolding a new project when none exists).

### Drive-by fix

The live click-through surfaced a session-tab duplication race: the event bus can deliver a new session's first `session.status` broadcast before `POST /sessions` resolves, and `createSession` appended unconditionally — every SPA-created session could render twice. It now keeps the (fresher) bus copy.

## Testing

- `pnpm test`: 372/372 (incl. new example-seed unit tests + `/sample-project` + `firstRun` REST tests)
- `pnpm test:ui`: 49/49 — new `welcome.spec.ts` covers: panel renders on fresh state (`?mockState=fresh`), both actions work, dismiss works, panel absent for returning users
- typecheck (server + web), lint, `build:server`, `build:web` all clean
- Live boot smoke against a scratch $HOME: fresh boot → welcome panel → clicked both actions in a real browser → sample seeded, real claude-code session opened, `order-triage` discovered in the rail, seeded canvas rendered; re-seeding reuses (`created: false`); second boot on the same HOME → `firstRun: false` + normal boot session (verified stable across repeated boots)

https://claude.ai/code/session_0171tUi9evGR88vUBeVEFr3R


gwitwer:
Merged `origin/feat/harness` (through #235) and re-verified: typecheck (server+web), 440/440 unit tests, sim:e2e PASS, 49/49 test:ui, plus a live click-through on the merged branch.

**Dedup drive-by × #235**: exercised live — ran the sample flow (exactly 1 tab through create despite the status broadcast racing the POST response), watched a full 10s liveness-sweep interval (tab count stable), then SIGKILLed the pty behind the live session: #235's machinery reconciled it to exited (tab left the strip, registry shows one exited record), with no duplicate at any point.

**One composition fix pushed (2a44ac0)**: #230's auto-render on session create bundles the project to extract its graph; the seeded sample ships without node_modules, so the auto-render was replacing the seed's hand-authored opening-shot canvas with 'render failed' panels. `renderCanvasForSession` now takes `preserveExistingOnFailure` — set only by the unprompted call sites — so a fully-failed auto-render keeps an existing canvas. Explicit renders (Visualize macro, POST /canvas/:id/render) still write the honest error page; a partially-successful overview still overwrites. Unit-tested + verified live (seeded canvas survives, 'render failed' gone).

Note: `pnpm lint` currently fails on `src/core/__fixtures__/*` from #230 (typed-linting project doesn't include them) — pre-existing on feat/harness, untouched here.

PR #239, opened 2026-07-08
feat(analytics-core): public contract surface — CONTRACT.md, fixtures, mock collector, trust docs

evtran0209:
## Summary

Publishes the public contract surface for `@sapiom/analytics-core` (merged in #175) and implements the ship-dark default the approving review on #175 asked for.

**Fixes SAP-1410** — https://linear.app/sapiom/issue/SAP-1410

### 1. Ship-dark default (answers the approving review's ask on #175)

The review on #175 flagged that the emitter defaulted its endpoint to the production collector URL while no production collector is deployed. The emitter now has **no default endpoint**: when neither `endpoint` (config) nor `SAPIOM_ANALYTICS_ENDPOINT` (env, used by tests) is set, `createAnalytics` returns the disabled no-op instance — events drop at enqueue, zero fetches, zero retries, zero disk writes, no first-run notice.

- The hosted URL survives as the exported, documented constant `SAPIOM_COLLECTOR_ENDPOINT` (replaces `DEFAULT_ENDPOINT`; the package is unpublished and has no consumers) and can be passed explicitly.
- **Flip condition:** once the production collector is deployed, restore send-by-default via the clearly marked single line in `resolveEndpoint()` (`return null;` → `return SAPIOM_COLLECTOR_ENDPOINT;`), plus the paired test/doc updates.

### 2. CONTRACT.md

The public collector wire contract, adapted for open-source publication: endpoint + auth header (`x-sapiom-api-key` — the only auth scheme; optional, enrichment-only), request envelope, leniency rules ("the only hard requirement is a parseable JSON body"), responses (202/400/413/429), server-stamped fields, identity semantics, producer obligations — including that **retried batches MUST reuse the same `event_id`s** (dedup is keyed on `event_id`; server-minted IDs are fresh per request, so such retries double-insert) — event taxonomy seed, and versioning. The storage sink is described abstractly as analytics storage.

### 3. Contract fixtures — `fixtures/contract/{valid,invalid}`

14 request/expectation descriptors shipped with the package: happy-path batches (single + multi-source), one fixture per leniency-table rule (missing `event_id`/timestamps/identity, unknown top-level keys, unknown `source`/`event_type`, non-object `data`, invalid API key, minimal empty event, no `events` key), and the documented rejections (`400` not-JSON ×2, `413` 501-event batch). The 1 MB `413` variant and `429` are documented as not fixture-able. Consumable by the backend collector e2e (vendored) and per-package instrumentation tests.

`fixtures.test.ts` keeps them honest: descriptor schema, leniency-rule coverage, field-level validation against the envelope builder (happy-path fixture must carry the builder's exact field set; every other event a subset), invalid-for-the-claimed-reason checks, and a full replay of all 14 against the mock collector asserting expected status + response.

### 4. `@sapiom/analytics-core/testing` — mock collector

`startMockCollector()`: an in-process real-HTTP mock (loopback, random port, Node built-ins only) with contract-shaped healthy responses (202 accepted counts / 400 / 413) and scriptable failure modes — `down` (socket destroyed), `slow`, arbitrary `status`. Captures envelopes (`requests` / `batches()` / `events()`, `waitForRequests()`), exported via an `exports` subpath + `typesVersions`, and exercised in-package by `mock-collector.test.ts` using the real emitter over real HTTP.

### 5. README trust package

Full telemetry disclosure in plain language: what is collected, the Sapiom-bound vs third-party content boundary (third-party calls: metadata only, never args/content), every opt-out (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, programmatic), the identity file (`~/.sapiom/analytics.json`, 0600, deletable), and the current ships-dark status.

### 6. Changeset

Minor bump for `@sapiom/analytics-core`.

## Test plan

- [x] `pnpm --filter @sapiom/analytics-core build` — clean (cjs + esm, `dist/*/testing/` emitted)
- [x] `pnpm --filter @sapiom/analytics-core test` — **7 suites, 117 tests passed** (2.7s). The 48 pre-existing tests stay green; the only adjusted ones are (a) the shared helper now pins the env-override endpoint so delivery suites keep sending under ship-dark, and (b) the single test asserting the old default-URL behavior is replaced by a no-endpoint → zero-fetches test plus an explicit-`SAPIOM_COLLECTOR_ENDPOINT` test
- [x] New coverage: no-endpoint no-op (zero fetches, zero disk writes, `enabled === false`), all 14 fixtures validated + replayed against the mock, mock failure modes (500 → exactly one retry then drop; down → socket failure, no throw; slow → enqueue stays instant), subpath resolution verified in both CJS `require` and ESM `import`
- [x] Repo-wide `pnpm build` / `pnpm typecheck` / `pnpm lint` / `pnpm test` — all green (all 13 packages; tools 322 passed, analytics-core 117 passed)
- [x] Leak grep over `CONTRACT.md`, `README.md`, `fixtures/`, `src/` for internal paths/tickets/infra names — clean
- [x] `npm pack --dry-run` — tarball includes `CONTRACT.md`, `fixtures/contract/**` (14 files), `dist/{cjs,esm}/testing/`; 58.7 kB total


evtran0209:
Internal claude review pass complete (repo has no CI reviewer round). Verdict was REVISE with 2 should-fixes + 3 nits — **all five taken** in the latest commit:

- ✅ README opt-out paragraph no longer implies the first-run notice fires when opted out (it never does — public-trust wording fix)
- ✅ CONTRACT.md leniency table: added the missing `events`-key-absent row (`{"accepted":0,"dropped":0}`, never rejected)
- ✅ `enabled` JSDoc documents that consent-denied and no-endpoint are intentionally merged + how to distinguish
- ✅ `endpoint: ''` documented as treated-as-absent + a test pinning the ship-dark no-op (118/118 green)
- ✅ Fixtures test now positively asserts the intentionally-excluded keys stay outside the envelope

Reviewer also verified clean: ship-dark zero-side-effects, retry preserves event_ids, mock-collector lifecycle leak-free, exports map in CJS/ESM/TS, zero-dep claim, no internal leaks, consent precedence matches README.

gwitwer (approved):
Approving — this closes out both asks from the #175 review, and goes further than asked: ship-dark is a full no-op instance (zero fetches, zero disk writes, no first-run notice), not just a skipped send. CONTRACT.md + the fixture suite + the /testing mock collector are exactly the surface we need on the harness side — we're adopting this contract wholesale (SAP-1449) and will verify harness telemetry end-to-end against startMockCollector(). One note: our earlier suggestion to move to Bearer auth is withdrawn — x-sapiom-api-key as documented here stands; this workstream is yours.

PR #248, opened 2026-07-09
docs(analytics-core): session telemetry conventions + canonical dot naming

evtran0209:
## Summary

Adds a **"Harness & session telemetry conventions"** section to `packages/analytics-core/CONTRACT.md` (after Producer obligations, before the taxonomy seed) covering:

- **Per-session `seq`** — producers that manage sessions locally MAY carry a monotonic per-session sequence number as `data.seq` (starts at 1; gaps indicate loss; `seq` order is authoritative over client timestamps; producer-assigned, stored verbatim, never renumbered server-side).
- **Batch context** — `data.context` MAY carry `app_version`/`os`/`arch`/`node`, stamped once per batch by the producer.
- **Session dimensions** — harness-style producers SHOULD carry `data.harness_session_id`, `data.agent_session_id`, `data.harness_kind` for cross-event analysis.
- **Canonical event naming** — dot-separated `<noun>.<verb/state>`; the envelope `source` field disambiguates same-named events from different producers; the collector stores `event_type` verbatim either way.

Also updates the event taxonomy seed table to the dot-form names (`session.start`, `capability.call`, `command.run`, `workflow.deploy`/`step.start`/`step.complete`/`step.error`, `model.call`/`tool.call`, ...), and brings the envelope example, README examples, and test sample event names in line with the convention. Contract fixtures are intentionally untouched — they exercise the free-string leniency rules, and `event_type` remains an open vocabulary.

Docs-only; no behavior change. Changeset: patch on `@sapiom/analytics-core`.

## Verification

- `pnpm --filter @sapiom/analytics-core test` — 8 suites, 150 tests passing
- `pnpm --filter @sapiom/analytics-core lint` and `typecheck` — clean

Refs SAP-1442
https://linear.app/sapiom/issue/SAP-1442


evtran0209:
Internal claude review pass complete. Verdict was FIX-FIRST with 4 findings — **all four taken**: `mcp.install` naming (the one dot-convention outlier), seq restart semantics documented (`seq` decrease = restart boundary, not loss — matches the shipped harness's actual behavior), `event_timestamp` ↔ `seq` cross-reference in the envelope example, and the canonical-name straggler in fixtures.test.ts. 150/150 tests green after fixes. Fixture sample names intentionally stay underscore-form (open vocabulary; coordinated rename tracked as a follow-up with the backend's vendored copies).

evtran0209:
86bbbc1 renames the `orchestration` source to `agent` (taxonomy seed row, envelope example's source list, and the `EventSource` union in `src/types.ts`): the package family it identified was renamed to `agent*`, so the taxonomy is aligned now, before any events exist. No behavior change — the collector accepts any source string verbatim. The `EventSource` union member was included beyond the docs so the shipped type can't contradict the contract; flagging in case that should be split out. Tests 150/150, lint/typecheck clean, leak grep clean.

gwitwer (approved):
Approved. Docs + light types, additive and non-breaking: the Envelope interface is byte-for-byte unchanged (no wire field renamed), all new session conventions (seq, context, harness_session_id, agent_session_id, harness_kind) ride inside data as convention-only, and schema_version correctly stays "1". The one type change (EventSource "orchestration"→"agent") has zero in-repo producers and source is an open verbatim vocabulary, so it's non-breaking; patch bump is defensible pre-1.0. Dot-naming is consistent across CONTRACT.md/README/types/index jsdoc/tests, and the fixtures are deliberately left snake_case to exercise the collector's open-vocabulary leniency. This establishes the canonical dot-naming that #253/#252/#249 emit against.

PR #249, opened 2026-07-09
feat(tools): emit capability.call usage analytics via analytics-core

evtran0209:
## Summary

Instruments `@sapiom/tools` at its single HTTP choke point — the private `Transport` in `src/_client/index.ts` — so every capability call (`fetch` and `request` alike, including resource-handle methods like `sandbox.readFile`) enqueues **one `capability.call` event** (dot-form, per project canon) through `@sapiom/analytics-core`:

- **Payload**: capability path/name (the routed capability id, e.g. `web.scrape`, or the request path for service-gateway calls), HTTP status, duration ms, size-capped request meta (method, URL, request byte size — all traffic is Sapiom-bound, bodies are never captured), plus the transport's attribution fields (`agent_id`/`agent_name`/`trace_id`/`trace_external_id`/metadata). Network-level failures are recorded with the error name/message and no status.
- **Envelope**: `source: "tools"`, `sdk_name: "@sapiom/tools"`, `sdk_version` read from the package's own package.json at first use (dual CJS/ESM-safe: `__dirname` in CJS, stack-frame anchor in ESM — verified against both built formats).
- **Lazy, once per client**: the emitter is created on first call and shared by `withAttribution`-derived transports (one batch queue / exit hook per client). `config.apiKey` is passed for server-side enrichment (sent as `x-sapiom-api-key` on collector batches only, never in payloads).
- **Ships dark**: with no collector endpoint configured the emitter is a silent no-op — zero network calls, zero disk writes. Opt-outs (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`) drop events at the entry.
- **Enqueue-only hot path**: `track()` is synchronous; no awaits were added to the call path, and the tracking call itself is wrapped so no failure can ever surface to a capability call.

**Dependency note**: adds `@sapiom/analytics-core` (`workspace:^`) as a **documented exception to tools' zero-dependency stance** — it is first-party and itself zero-dependency.

Also adds a "Usage analytics" disclosure section to the package README and a `minor` changeset.

## Zero-regression evidence

- Existing `@sapiom/tools` suite passes **unchanged**: 322/322 baseline tests green before and after (now 352 with the 30 new analytics tests; 19 suites, ~2.8s — no timer waits, flushes are deterministic).
- **No public API changes**: `Transport` is package-internal (not exported from the barrel or any subpath); the only additions are a private field + private method on it and a new internal `_client/analytics.ts` module. Public-surface smoke spec untouched and green.
- **No new required config**: default behavior is a no-op; the collector endpoint, opt-outs, and api key are all optional/ambient.
- Repo-wide `pnpm build` / `pnpm typecheck` / `pnpm lint` / `pnpm test` all exit 0 (lint has only pre-existing warnings in untouched packages).
- Leak grep over lines added by this branch (`quito|conductor|plans/|SAP-[0-9]|heroku|bigquery`) is empty; remaining `SAP-*` grep hits in `packages/tools/src` are pre-existing design references on `main`.

## e2e coverage (`analytics.e2e.spec.ts`, real mock collector over HTTP via `SAPIOM_ANALYTICS_ENDPOINT`)

- Full envelope + data assertions for a routed capability call (`search.scrape` → `web.scrape`), sandbox **handle methods** (`writeFile`/`readFile`), and a `transport.request()` capability (`Repository.create`) counted exactly once.
- HTTP-500 and network-failure paths: events captured, thrown errors byte-identical (same class/status/body, original error identity preserved).
- Attribution stamping, including through `withAttribution` (shared emitter proven).
- `SAPIOM_TELEMETRY_DISABLED=1` → **zero collector requests**, identical results; `DO_NOT_TRACK=1` → zero requests; dark default → zero requests **and zero disk writes**.
- Fault injection — collector **down / 500 / slow** — capability results deep-equal the telemetry-dark baseline, nothing thrown.
- Plus a dist-level smoke (run manually) of the built ESM package through public `createClient().search.scrape(...)` delivering a real batched event: correct envelope, `sdk_version: 0.16.0`, api-key header.

Fixes SAP-1416
https://linear.app/sapiom/issue/SAP-1416


evtran0209:
## Review triage — verdict FIX-FIRST → all five findings taken (25a69ad)

| # | Finding | Status | What was done |
|---|---------|--------|---------------|
| 1 | **`beforeExit` listener leak** (blocker) | **Taken** | Added `Transport.shutdown(): Promise<void>` (delegates to the holder's emitter `shutdown()`, resolves immediately when none exists, idempotent, never rejects) and wired it through the `Sapiom` interface + `bind()` — additive public API only; the stub client implements it as an immediate resolve. One call covers all `withAttribution`-derived clients (shared holder). Documented in the README analytics section with per-execution-client guidance. New e2e: **12 sequential clients** (above Node's 10-listener warning threshold), each `scrape` + `shutdown()` — `process.listenerCount("beforeExit")` returns to baseline after every client. |
| 2 | **Query strings in `data.url`** | **Taken** | `data.url` is now origin + pathname (`urlWithoutQuery`, consistent with `capabilityFromUrl`); query strings and fragments are never recorded, so a future/direct `transport.fetch(urlWithSecretQuery)` cannot exfiltrate. Unparseable URLs are cut at the first `?`/`#`. New e2e: a direct `transport.fetch(".../v1/files?token=SECRET&...")` stores `url: "https://files.test/v1/files"` and the serialized envelope contains no `SECRET`. Docstring on `capabilityCallData` updated to state path-only capture. |
| 3 | **README disclosure alignment** | **Taken** | Now reads "the request URL path (query strings are stripped, never recorded)" — accurate with (2) in place. |
| 4 | **Version constant for bundled consumers** | **Taken** | `scripts/generate-version.mjs` writes `src/_generated/version.ts` from package.json; gitignored + regenerated via `gen:version`, chained into `build`/`test`/`typecheck`/`dev` (CI runs build first, and each entry point regenerates, so it can never go stale — verified by simulating a changesets bump to 0.17.0: gen re-ran and the suite stayed green, so release PRs aren't blocked). `analytics.ts` imports `VERSION` as the primary source (bundlers inline it); the package.json walk-up remains fallback-only. Unit tests pin `VERSION === package.json.version` and `resolveSdkVersion() === VERSION`. Lint also verified green with the generated file absent (fresh-clone path). |
| 5 | **Down-mode fault-injection rigor** | **Taken** | The parameterized fault-injection tests (incl. `down`) now assert the captured collector attempt carries `content-type: application/json` **and** `x-sapiom-api-key: test-key` — a genuine HTTP delivery attempt, not a stray socket capture. |

**Non-issues confirmed by reviewer** (no change): error identity preserved on rethrow; request **body size only** is recorded (never content); api key rides the collector header for enrichment, never event payloads; `withAttribution` holder sharing (one emitter per client); ship-dark writes nothing to disk.

**Gates re-verified on 25a69ad:**
- `@sapiom/tools`: **360 passed** — the 322 baseline tests unchanged and green + 38 analytics tests (19 suites, 2.4s)
- Build green in both formats; `resolveSdkVersion()` → `0.16.0` from CJS and ESM dist; ESM dist smoke: listener registered on first call → released by `shutdown()` → idempotent second call
- Repo-wide `typecheck` exit 0, `lint` exit 0 (0 errors), `test` exit 0 (all packages, including `@sapiom/agent` which consumes the `Sapiom` type)

gwitwer (approved):
Approved after close review — this was the highest leakage-risk PR (the SDK transport carries real payloads) and it's clean. Verified: capability.call captures capability id/method/query-stripped URL/ok/status/duration_ms/request_bytes (SIZE only) + error name+message — request and response BODIES are never read; an HTTP 500 records only status:500/ok:false (no response body); query strings/fragments are stripped; apiKey is the x-sapiom-api-key header, never in payload; oversized fields 16KB-capped downstream. shutdown() detaches the beforeExit listener, flushes, clears timers — idempotent, never rejects, all timers unref'd so it can't hold the process open. Instrumentation is fully try/catch-wrapped and off the call path (rethrows the original error unchanged), proven deep-equal to the telemetry-dark baseline under fault injection. Ships dark by default; capability.call dot-naming matches #248. Two awareness notes for the rollout (non-blocking): user-nameable path segments (sandbox/file names, not contents) travel as capability/url metadata for gateway calls, and network-failure error_message is captured verbatim (generic Node fetch messages, 16KB-capped) — both defensible, just be deliberate as the collector goes live.

PR #251, opened 2026-07-09
fix(harness): validate user-derived filesystem paths (close CodeQL path-expression alerts)

gwitwer:
## What

Hardens the harness filesystem sinks CodeQL flagged as **"Uncontrolled data used in path expression"** (`js/path-injection`) so a user-derived path segment can never escape its intended root. Real-world exploit risk is low — `@sapiom/harness` is a localhost tool operating on the user's own files with the user's own inputs, no privilege boundary — but the guards prevent genuine footguns (`..` traversal, absolute-path surprises, symlink escape) and clear the security scan.

## How

New shared `packages/harness/src/core/path-safety.ts`, reusing the guard pattern already proven in the generated-dir retention policy:

- **`childPath(root, name)`** — the one-plain-segment check, extracted from `retention.ts` (which now imports it; behavior unchanged).
- **`resolveWithinRoot(root, candidate)`** — `path.relative`-based containment; rejects `..` climbs, absolute paths, and sibling dirs that merely share the root's name prefix.
- **`hasTraversalSegment(p)`** — segment-aware `..` assertion for endpoints that legitimately accept an arbitrary absolute path (a filesystem browser, an explicit "connect this directory") and so have no single root to confine to.

Applied at the **7 flagged sinks**, each guard explicit and local to its sink:

| File | Sinks | Fix |
|------|-------|-----|
| `core/workflow-registry.ts` | `readMarker`, `nameFor`, `scanDir` (readdir) | confine every scanned dir under the scan root (`resolveWithinRoot`); assert no `..` segment before each read |
| `server/canvas.ts` | `fs.stat` / `fs.readFile` / `fs.createReadStream` | resolve the request path through `resolveWithinRoot(canvasRoot, …)`, replacing the inline prefix check |
| `server/fs.ts` | `fs.readdir` | assert no residual traversal segment on the resolved dir |

Legitimate inputs (real workflow dirs, real slugs, arbitrary browse targets) resolve exactly as before; only malformed/escaping paths are rejected.

## Verification

- **Unit** — 604 harness tests pass, incl. 28 new `path-safety` tests (`..`, absolute, separator, symlink, sibling-prefix, legit segment).
- **Typecheck** (src + web), **lint**, **build:server**, **build:web** — all green.
- **`sim:e2e`**, **`e2e:live`** (exercises `/api/fs/list`, canvas render/serve, workflow scan), **`test:ui`** (63 Playwright tests) — all pass.
- **Real project** — scan + bind + visualize a real workflow tree still renders (writing only inside its `.sapiom/`), while a raw `..` traversal request is rejected with 400.

### DIFF
--- a/packages/harness/src/core/inject/retention.ts
+++ b/packages/harness/src/core/inject/retention.ts
@@ -24,7 +24,7 @@
  *
  * Safety: deletion only ever targets a plain directory that is a direct
  * child of a root that itself looks like a dedicated generated-config dir
- * (resolveGeneratedRoot / childDirPath). Symlinked entries are never
+ * (resolveGeneratedRoot / childPath). Symlinked entries are never
  * followed and never deleted.
  */
 
@@ -33,6 +33,7 @@ import * as path from "node:path";
 
 import { HARNESS_PATHS } from "../../shared/types.js";
 import { expandHome } from "../paths.js";
+import { childPath } from "../path-safety.js";
 
 /** Sweep threshold: a non-live dir must be at least this stale (mtime) to be
  *  removed. Generous — the only cost of keeping a dead dir is disk clutter,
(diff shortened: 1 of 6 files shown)


evtran0209 (approved):
lgtm!

PR #252, opened 2026-07-09
feat(cli): emit command.run usage analytics via analytics-core

evtran0209:
Instruments `@sapiom/cli` with usage analytics via the merged `@sapiom/analytics-core` emitter. **Zero-regression ticket**: identical command output and exit codes, no new required configuration, enqueue-only delivery.

Fixes SAP-1418
https://linear.app/sapiom/issue/SAP-1418

## What

- **`command.run` events** (dot canon) via commander `preAction`/`postAction` hooks on the root program (`packages/cli/src/lib/analytics.ts`, wired in `buildProgram()`): `source: "cli"`, `sdkName: "@sapiom/cli"`, and a fixed data shape — `command` (canonical path, e.g. `agents deploy`), `flags` (user-passed flag **names only**, never values or positional arguments), `duration_ms`, `exit_code`. Failed commands still report (the shared `action()` wrapper sets `process.exitCode` instead of rethrowing, so `postAction` runs).
- **Identity**: `SAPIOM_API_KEY` / stored session credential is passed as the emitter's `apiKey` — analytics-core sends it as the `x-sapiom-api-key` header only; it never enters payloads. No `user_id` is synthesized.
- **Ships dark**: no endpoint → `createAnalytics` returns a no-op — zero network calls, zero disk writes, no first-run notice (verified e2e).
- **Version reporting** reads the package.json next to the running `dist/bin.js` (name-checked, falls back to `0.0.0`) — no `import.meta`, so the module stays loadable under the CJS test runner.
- README gains a short usage-analytics disclosure; changeset: `@sapiom/cli` minor.

## Zero-regression gates

- Existing 19-test suite passes unchanged; full cli suite now 37 tests (10 in-process unit + 8 e2e added).
- e2e runs the **built CLI** against `startMockCollector()` (`SAPIOM_ANALYTICS_ENDPOINT`), with a fresh temp `HOME` per run:
  - envelope assertions (event/source/sdk fields, exact data keys, schema `1`)
  - first-run notice fires once per machine, marker stamped in `~/.sapiom/analytics.json`, silent on the second run
  - `SAPIOM_TELEMETRY_DISABLED=1` and `DO_NOT_TRACK=1` → zero requests, byte-identical stdout
  - `login` (against a local RFC 8628 stub) and `logout`: access token, device code, user code, org identifier, env/stored API keys **never appear in any request body** (explicit assertions); credential appears only as the delivery header
  - fault injection — collector down, connection refused, HTTP 500 — commands exit 0 with byte-identical stdout
- Dark-mode manual check: `--help`, `config set-target`, unknown-command output and exit codes identical; no `~/.sapiom` created.
- New `jest.global-setup.cjs` builds the package (and missing workspace deps) before tests so the e2e always runs the current `dist/bin.js`.

## Verification (local)

```
packages/cli:  build ✓  typecheck ✓  lint ✓
Test Suites: 4 passed, 4 total
Tests:       37 passed, 37 total

repo:  pnpm build ✓  pnpm typecheck ✓  pnpm lint ✓  pnpm test ✓ (exit 0, all packages)
leak grep (quito|conductor|plans/|SAP-[0-9]|heroku|bigquery) on packages/cli: clean
```

### DIFF
--- a/.changeset/cli-command-run-analytics.md
+++ b/.changeset/cli-command-run-analytics.md
@@ -1,0 +1,19 @@
+---
+"@sapiom/cli": minor
+---
+
+Emit anonymous `command.run` usage analytics via `@sapiom/analytics-core`.
+
+- One `command.run` event per executed command (commander `preAction`/`postAction`
+  hooks), carrying the command path (e.g. `agents deploy`), the names of the
+  flags used — never their values or positional arguments — the duration, and
+  the exit status. Tokens and emails never reach event payloads; a signed-in
+  credential (from `SAPIOM_API_KEY` or the stored session) is only attached as
+  a delivery header for server-side identity enrichment.
+- Ships dark: without an explicitly configured collector endpoint the emitter
+  is a silent no-op — zero network calls, zero disk writes, no notice. When
+  enabled, analytics-core's one-time first-run notice explains the collection
+  and the opt-outs (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`).
+- Zero behavior change: enqueue-only delivery (best-effort flush on process
+  exit), identical command output and exit codes, and no new required
+  configuration.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -31,3 +31,14 @@ sapiom agents schedule cancel <scheduleId>
 
 Run `sapiom agents --help` for the full command set. Every command accepts
 `--json` for machine-readable output.
+
+## Usage analytics
+
+The CLI can emit anonymous usage events through
+[`@sapiom/analytics-core`](https://github.com/sapiom/sapiom-js/tree/main/packages/analytics-core):
+one `command.run` event per executed command, carrying the command name, the
+names of the flags used (never their values or arguments), the duration, and
+the exit status. Nothing is sent unless a collector endpoint is explicitly
+configured, and delivery is best-effort in the background — analytics can
+never slow down or fail a command. Opt out at any time with
+`SAPIOM_TELEMETRY_DISABLED=1` or `DO_NOT_TRACK=1`.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -38,7 +38,8 @@ The CLI can emit anonymous usage events through
 [`@sapiom/analytics-core`](https://github.com/sapiom/sapiom-js/tree/main/packages/analytics-core):
 one `command.run` event per executed command, carrying the command name, the
 names of the flags used (never their values or arguments), the duration, and
-the exit status. Nothing is sent unless a collector endpoint is explicitly
-configured, and delivery is best-effort in the background — analytics can
-never slow down or fail a command. Opt out at any time with
-`SAPIOM_TELEMETRY_DISABLED=1` or `DO_NOT_TRACK=1`.
+the exit status. Nothing is currently sent anywhere. When delivery is
+enabled, it is best-effort and can never fail a command: it never slows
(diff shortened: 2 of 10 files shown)


evtran0209:
## Review triage — FIX-FIRST addressed in 4779903

**Taken (1, 2, 5, 6):**

1. **Exit-latency truth + slow-collector e2e (blocker)** — README no longer claims analytics "can never slow down" the CLI wholesale: command *execution* is never slowed, and the on-exit flush is bounded by the emitter's 5-second request timeout. New e2e runs the built CLI against `collector.setMode({ kind: "slow", delayMs: 8_000 })` and measures wall time: byte-identical stdout, exit 0, wall > 4.5s (the flush genuinely waits through the timed-out attempt) and < 12s (generous CI margin). One correction to the review's arithmetic, verified empirically: the worst case is **~5s, not ~10.5s** — after the first attempt times out, analytics-core's retry backoff timer is unref'd, so it cannot hold the exiting process open; the batch is dropped rather than retried. Measured 5167/5122/5134 ms across three runs, with exactly **one** delivery attempt on the wire each time — the test asserts that single attempt, and the README documents "retries never hold the process open".
2. **Ancestor flags** — `specifiedFlagNames` now walks the command chain root→leaf (program → groups → leaf), collecting long names where `getOptionValueSource(...) === "cli"`, deduped on first occurrence. No command defines group-level options today, so current envelopes are unchanged; future group flags are captured automatically. Unit test added with a group-level value option (`things --level <secret> push --json` → `["--level", "--json"]`, value asserted absent).
5. **`cliVersion` comment** — the `0.0.0` fallback is now documented as the expected, non-error state whenever the CLI isn't the process entrypoint (under Jest, argv[1] is the worker script so the name check fails by design; likewise library imports of `buildProgram()`).
6. **README endpoint phrasing** — now reads "Nothing is currently sent anywhere."; the endpoint env var stays deliberately undocumented.

**Declined (3, 4):**

3. **globalSetup building on unit-only runs** — costs 1–3s locally; CI is unaffected (the workflow's build step precedes tests, so the setup is a fast incremental rebuild of this package only). Splitting jest projects isn't worth the config complexity right now.
4. **WeakMap dangling entry** — as the review itself concluded, correct by design: the CLI runs one command per process and the map is keyed by the action `Command`, so there is nothing to leak in practice; no action.

**Reviewer-confirmed non-issues** (no change needed): privacy extraction (flag names only — never values or positionals), `exit_code` fidelity (`process.exitCode` + the non-rethrowing `action()` wrapper means failures still emit), and identity resolution (env → stored session, header-only, never in payloads).

**Verification:** `packages/cli` — typecheck ✓ lint ✓, `Tests: 39 passed, 39 total` (19 baseline unchanged, now 11 unit + 9 e2e). Repo — `pnpm build` ✓ `pnpm typecheck` ✓ `pnpm lint` ✓ `pnpm test` exit 0 (all packages). Leak grep on packages/cli: clean.

gwitwer (approved):
Approved after review against the analytics contract. Privacy is the standout: command.run carries command name + specified-flag NAMES + duration + exit_code only — never flag values, positionals, defaults, or env-derived values; the API key rides as the x-sapiom-api-key header, never in the payload. Verified by unit tests and e2e that grep the raw wire body for tokens/device codes/org email/API keys and assert absence. Flush is analytics-core's beforeExit batch drain with unref'd timers + 5s bounded fetch (no exit hang, no retry loop), both commander hooks are try/catch-wrapped so analytics can never break the command or its exit code, and it ships dark by default. command.run dot-naming matches #248's canon. Ship it.

PR #254, opened 2026-07-09
feat(langchain): metadata-only model/tool call analytics with redaction boundary

evtran0209:
## Summary

Instruments the `@sapiom/langchain` and `@sapiom/langchain-classic` wraps with **metadata-only** usage analytics via `@sapiom/analytics-core`, guarded by a structurally enforced privacy boundary and sentinel-based redaction tests.

Fixes SAP-1420 — https://linear.app/sapiom/issue/SAP-1420

### What's emitted (`source: "langchain"`, one event per underlying invocation)

| Event | Fields (closed allow-list) |
|---|---|
| `model.call` | `status`, `duration_ms`, `model`, `provider`, `input_tokens`/`output_tokens`/`total_tokens` (when available), `error_class` |
| `tool.call` | `status`, `duration_ms`, `tool_name`, `error_class` |

- `@sapiom/langchain`: hooks the existing `wrapModelCall` / `wrapToolCall` in `packages/langchain/src/middleware.ts`.
- `@sapiom/langchain-classic`: hooks `wrapSapiomTool` / `SapiomDynamicTool` (`src/tool.ts`) and the `SapiomChatOpenAI` / `SapiomChatAnthropic` `generate()` overrides.

### Privacy boundary (the point of this ticket)

These wraps intercept calls to arbitrary user tools/models whose data never otherwise touches Sapiom. Hard exclusions — no prompt text, no completions, no tool args, no tool results, no message content, no schemas/descriptions, no error messages — are enforced **structurally**: payloads are produced exclusively by allow-list builders (`src/internal/analytics.ts` in each package) that copy a closed set of scalar fields one by one (type-validated, length-capped). Errors contribute their constructor name only. There is no code path that spreads or serializes a request, message, args object, or result into an event.

**Redaction boundary tests**: wrapped models/tools run with unique sentinels on every user-content surface (prompts, system messages, multimodal parts, completions, tool arg keys *and* values, results, descriptions, schema descriptions, error messages, trace ids, stop sequences) → assert sentinels appear in **zero** bytes captured by `startMockCollector()` (raw request bodies, not just parsed events).

### Zero-regression gates

- Existing langchain (21) and langchain-classic (165+2 skipped) suites pass with assertions unchanged (4 `as any` casts added at existing call sites to satisfy langchain 1.5's stricter generics).
- No public API changes; no new required config; emission is enqueue-only (`track()` is a synchronous never-throwing enqueue) — no hot-path awaits.
- Ships dark: without `SAPIOM_ANALYTICS_ENDPOINT` (or an explicit endpoint) nothing is sent; `SAPIOM_TELEMETRY_DISABLED=1` / `DO_NOT_TRACK=1` → zero requests (tested).
- Fault injection (collector down / HTTP 500): wrapped results and errors pass through identically (tested).

### Heads-up: un-quarantines the langchain packages

Both packages were excluded from the pnpm workspace on 2026-06-16 ("until their @langchain/* deps are realigned") — they could not build or test at all. This PR completes that realignment, which is a prerequisite for instrumenting them:

- `@sapiom/langchain`: `langchain ~1.5.3` + `@langchain/core ~1.2.1` (mutually consistent peer set), adopting the `createMiddleware()` factory required by langchain 1.5's branded `AgentMiddleware` type. Tilde ranges prevent the float-to-latest drift that caused the original breakage.
- `@sapiom/langchain-classic`: pins `@langchain/openai ~1.1.0`, the last line that works against `@langchain/core` 0.3.x (restores the pre-quarantine resolution; the known out-of-range peer warning remains, as before).
- `pnpm-workspace.yaml` excludes removed; both packages are back in CI build/typecheck/lint/test and in changesets publishing.

### Test evidence (local, matching CI steps)

- `pnpm build` → exit 0 (all packages)
- `pnpm typecheck` → exit 0
- `pnpm lint` → exit 0 (0 errors)
- `pnpm test` → exit 0; langchain: 34 passed (2 suites), langchain-classic: 180 passed / 2 skipped (9 suites); all other packages green
- `pnpm install --frozen-lockfile` → exit 0 (CI parity)

### Changesets

- `@sapiom/langchain`: minor
- `@sapiom/langchain-classic`: minor


evtran0209:
## Review triage — round 1 applied in 1f7a05f

**Applied:**

1. **(Publish-blocker) Peer floor** — `@sapiom/langchain` peer `langchain: ">=1.0.0"` → `">=1.5.0"`. `createMiddleware()` doesn't exist before 1.5; a 1.0–1.4 install would have crashed at import time.
2. **(Publish-blocker) Missing classic peers** — `@sapiom/langchain-classic` now declares `@langchain/openai: "~1.1.0"` and `@langchain/anthropic: "^0.3.30"` as peers, both `peerDependenciesMeta.optional: true` (users of one model family install nothing extra). Ranges match what the code is actually built/tested against: openai per the `~1.1.0` realignment (1.2+ pulls break against core 0.3.x), anthropic per the devDep/lockfile (`^0.3.30`, resolved 0.3.34).
3. **Version constants** — `SDK_VERSION` (langchain utils.ts) `"0.1.0"` → `"0.4.1"` and classic `ANALYTICS_SDK_VERSION` `"1.0.0"` → `"0.4.1"`. Went with the align-the-constants fallback: a `resolveJsonModule` import of `package.json` is **not** clean in this build setup — package.json sits outside `rootDir: ./src`, so importing it reshapes the emitted `dist/` layout (output becomes `dist/{cjs,esm}/src/...` + a copied package.json). Kept the TODO with that rationale inline.
4. **`asSapiomStateUpdate`** — the `never` lie is gone, with your comment. One nuance: a plain `unknown` return type doesn't type-check against `BeforeAgentHook` (unknown isn't assignable to the hook's `MiddlewareResult` update type), so the shape is `(update: Record<string, unknown>): MiddlewareResult<Partial<BuiltInState>>` with the cast routed through `unknown`. Same spirit — honest opaque cast, honest declared target.
5. **Changeset copy** — langchain changeset now carries the explicit line: *"The `langchain` peer dependency floor is now ≥1.5.0 (`createMiddleware()` was introduced in 1.5; earlier versions fail at runtime)."* Classic changeset documents the two new optional peers.
6. **Test casts documented** — comment block above `createMockToolRequest` in middleware.test.ts explains the 4 `request as any` casts (bridging the minimal mock shape to langchain 1.5's branded `ToolCallRequest` generics; behavioral assertions unaffected; fully-typed fixtures tracked as a follow-up).

**Not taking (as discussed):**

- **402-payment-retry emitting error+success** — keeping one event per underlying invocation; the semantics are design-validated by the dedicated test (`emits one event per underlying invocation on payment retry` asserts exactly `["error", "success"]`).
- **Suppress-vs-`retried_for_payment`-flag question** — filed as a separate follow-up ticket rather than churned here.

**Gates re-verified after changes** (local, `SAPIOM_API_URL` unset):
- `@sapiom/langchain`: 2 suites, **34/34 passed**; `@sapiom/langchain-classic`: 9 suites, **180 passed / 2 skipped (182)**
- Repo `pnpm build` / `pnpm typecheck` / `pnpm lint` → all exit 0
- `pnpm install --frozen-lockfile` → exit 0 (lockfile unaffected by the peer additions — both peers already resolve via devDeps)
- prettier clean on all touched files

gwitwer (approved):
Approved — this was the highest leakage-risk PR (LangChain calls carry prompts + completions) and the 'metadata-only with redaction boundary' claim is literally true. Traced every field that can reach a payload: buildModelCallData/buildToolCallData copy a closed set of scalars one-by-one (status, duration_ms, model, provider, {input,output,total}_tokens, tool_name, error_class) — no spread, no JSON.stringify of request/response/args/result, no object forwarding. Token counts read only usage_metadata.*_tokens (type-checked numbers); model/provider are code identifiers (asName, 256-cap); tool_name is the name only (args never enter the analytics path); errors are error.constructor.name, never .message. The sentinel tests are the right design — unique markers on prompt/system/multimodal/completion/tool_calls/arg-key+value/result/description/error/trace/stop-sequences, asserted absent from the RAW request bodies at the mock collector, not just parsed events. Consent + ship-dark enforced via analytics-core; emit is post-await, non-blocking, rethrows unchanged, one event per invocation; fault-injection clean. pnpm-workspace.yaml change is correct and benign — it removes the 2026-06-16 quarantine and completes the @langchain/* dep realignment, which also resolves the earlier build:cjs failure. Non-blocking: the langchain peer floor >=1.0→>=1.5 is effectively breaking for 1.0–1.4 consumers (fine under 0.x, but surface it prominently in release notes); hardcoded SDK_VERSION 0.4.1 will lag the minor bump (existing TODO covers it).

PR #253, opened 2026-07-09
feat(mcp): registerTool seam + tool.call usage analytics

evtran0209:
## Summary

Instruments `@sapiom/mcp` with usage analytics, in two strictly-ordered commits (Tidy First):

**Commit 1 — structure only (zero behavior change).** Introduces `registerTool(server, name, description, schema, handler)` and routes all 16 existing `server.tool(...)` registrations through it (`sapiom_authenticate`, `sapiom_status`, `sapiom_logout`, and the 13 `sapiom_dev_agents_*` tools). The seam forwards to `server.tool` verbatim; the pre-existing suite passes unchanged, and a live stdio `listTools` smoke check shows the identical 16-tool surface.

**Commit 2 — emission.** `registerTool` wraps handlers to emit one **`tool.call`** event per invocation (dot-canon event name; the envelope's `source: "mcp"` disambiguates) via `@sapiom/analytics-core`, with `sdkName: "@sapiom/mcp"`:

- `data`: tool name, args (Sapiom-bound developer tools → full capture per the analytics contract; the emitter size-caps each field at ~16 KB with `_truncated` flagging), `duration_ms`, `ok`, and `error_class` (the structured `error.code` from `isError` payloads, the constructor name for thrown errors).
- Synchronous enqueue — no awaits on the hot path; emission can never throw or alter a tool result.
- The emitter is constructed once at server startup, keyed with the cached credential's API key when one exists (surfaces as `x-sapiom-api-key`). Ships dark: no collector endpoint (`SAPIOM_ANALYTICS_ENDPOINT`) → full no-op; `SAPIOM_TELEMETRY_DISABLED=1` / `DO_NOT_TRACK=1` opt out.

Also adds a "Usage analytics" section to the package README and a `minor` changeset.

## Zero-regression gates

- Commit 1 is provably structure-only: `registerTool` forwards verbatim; existing suite green before commit 2.
- Pre-existing `@sapiom/mcp` test files are byte-identical to `main` and pass unchanged at the final state.
- The #187 `./auth` subpath export is untouched (`auth-api.ts` byte-identical) and still resolves with the same exports, verified by an ESM import from `@sapiom/harness`.
- No public API changes; no new required config; no hot-path awaits.

## Tests

- Unit: `registerTool` pass-through (commit 1) + emission behavior, including fault injection at the emitter boundary (`track()` throwing does not affect tool results).
[...]

PR #250, opened 2026-07-09
feat(orchestration): emit workflow/step lifecycle analytics via analytics-core

evtran0209:
Fixes SAP-1419 — https://linear.app/sapiom/issue/SAP-1419

Instruments the orchestration stack (`@sapiom/agent-core` + `@sapiom/agent-runtime`, the packages formerly named `orchestration-core`/`orchestration-runtime`) with workflow lifecycle usage analytics via `@sapiom/analytics-core`, `source: "agent"`. **Zero-regression ticket** — telemetry can never change an operation's behavior, results, or errors.

## What's emitted

All events carry `source: "agent"`.

| Event | From | Data (metadata only) |
|---|---|---|
| `workflow.link` | `link()` | workflow name/id, status, duration_ms, error_code |
| `workflow.deploy` | `deploy()` | workflow id, branch, build_run_id, build_status, status, duration_ms, error_code |
| `workflow.run` | `run()` | workflow id, execution_id, status, duration_ms, error_code |
| `step.start` / `step.complete` / `step.error` | `AgentRunnerCore` (via injected sink) | workflow_name, step, attempt, execution_id, duration_ms, error_name |

> Source is `"agent"`, aligned to the `agent*` package family rename; `"orchestration"` was the pre-rename label and no events were ever collected under it (the emitter is unreleased and ships dark), so the `EventSource` union and CONTRACT.md were updated in place.

Local `runLocal` runs emit the step events flagged `data.local: true`. Payloads are names/ids/codes/durations only — never inputs, outputs, or error messages.

## Design

- **GatewayClient stays env-free.** The emitter is a lazy process-shared instance constructed at the operation call boundary (`packages/agent-core/src/analytics.ts`), never inside `GatewayClient`. The package-level design-contract doc in `index.ts` records this as the sole, telemetry-only exception.
- **The runtime takes no new dependency.** `AgentRunnerCore` gains an optional `analytics` field typed by a new structural `RuntimeAnalytics` host interface (`{ track(eventType, data) }`); `SapiomAnalytics` satisfies it as-is. No sink → no events, previous behavior byte-for-byte. Every emission site is guarded so a throwing sink cannot affect the walker.
- **Ships dark.** Without `SAPIOM_ANALYTICS_ENDPOINT` every `track` is a silent no-op (zero network, zero disk). `SAPIOM_TELEMETRY_DISABLED=1` / `DO_NOT_TRACK=1` opt out. Emission is synchronous enqueue-only — no awaits added to any hot path.

[...]

PR #256, opened 2026-07-09
feat(analytics-core): deliver to the hosted collector by default (ship-dark flip)

evtran0209:
## Why now

The hosted analytics collector is now **live in production**, so this executes the pre-planned ship-dark flip in `@sapiom/analytics-core`: the emitter's default endpoint turns on. This was designed for from day one — the flip line was marked in code.

## The flip

`resolveEndpoint()` in `src/http-sender.ts` carried the marked flip:

```ts
// SHIP-DARK DEFAULT: the hosted collector is not deployed yet, so an
// unconfigured emitter must send nothing. Once the collector is live,
// flip the next line to `return SAPIOM_COLLECTOR_ENDPOINT;`.
return null;
```

That line now returns `SAPIOM_COLLECTOR_ENDPOINT`. Precedence is unchanged: `config.endpoint` → env override → **(new) hosted default**. Empty strings are still treated as absent, so `endpoint: ""` falls through to the env override and then to the hosted default (documented on `resolveEndpoint` and `AnalyticsConfig.endpoint`).

Because resolution can no longer yield `null`, the return type narrows `string | null` → `string`, and the now-unreachable "no endpoint → disabled instance" branch in `createAnalytics` is removed (it would be a TS2367 strict-mode error left in place). Consent is the only off switch, exactly as before: `disabled: true`, `SAPIOM_TELEMETRY_DISABLED=1`, or `DO_NOT_TRACK=1` still yields a full no-op — zero fetches, zero disk writes, no first-run notice.

## Tests

- `fault-injection.test.ts` — the two ship-dark tests flipped:
  - "no endpoint configured → no-op" is now "no endpoint configured → delivers to the hosted collector by default", asserting the delivered URL is `SAPIOM_COLLECTOR_ENDPOINT` (via injected capturing fetch — nothing real is dialed).
  - the `endpoint: ""` test now asserts empty string = absent → hosted default.
  - a new test keeps the **true no-op path** pinned with no endpoint configured: `SAPIOM_TELEMETRY_DISABLED=1` and `disabled: true` each produce zero fetches and zero disk writes.
- `internals.test.ts` — `resolveEndpoint` unit tests updated for the live default; the empty-string test now also pins that an empty config endpoint yields to the env override before the hosted default.
- `helpers.ts` — `cleanAnalyticsEnv` docs now state why the loopback endpoint pin is mandatory post-flip: no suite may ever target the production URL, even if a test forgets to inject `fetchImpl`.

## Cross-package test audit (dark-default → live-default)

Audited every other package for tests constructing an emitter with no endpoint and no opt-out, which would now attempt real network calls to the production URL from CI:

- **No package in this repo depends on `@sapiom/analytics-core` yet** — verified via every `packages/*/package.json` and a repo-wide source grep for `createAnalytics` / `SAPIOM_ANALYTICS_ENDPOINT` / `@sapiom/analytics-core`. The harness's consent/machine-id modules reference analytics-core only in comments (they have their own standalone implementation), and its "analytics pipeline" e2e scripts are the session-transcript collector, unrelated to this emitter.
- **Result: zero cross-package test adjustments were required.** Confirmed empirically by the full-repo suite below.
- Within analytics-core itself, every delivery suite already pins `SAPIOM_ANALYTICS_ENDPOINT` to a loopback URL via `cleanAnalyticsEnv()` **and** injects `fetchImpl`, so no test can reach the real endpoint.

## Docs

- **README**: "Current status: ships dark" → "Current status: live by default", restating all opt-outs with the same wording as the telemetry disclosure section; the intro bullet, the usage example, and the identity-file note ("never created while analytics is disabled" — the "or unconfigured" clause no longer applies) updated to match.
- **CONTRACT.md**: the endpoint section now states `SAPIOM_COLLECTOR_ENDPOINT` is the emitter's default endpoint.
- **Changeset** (minor): delivery on by default to the hosted collector; opt-outs unchanged. Note: the programmatic opt-out is written as `disabled: true` — this package's actual config option — since `analytics: false` is not an option of `@sapiom/analytics-core` itself (it is the switch host packages are expected to expose).

## Verification

- analytics-core: 8 suites, **151 tests passed**
- Full repo `pnpm test` (with `SAPIOM_API_URL` unset): **all 13 in-scope packages green**, exit 0 (`langchain`/`langchain-classic` are quarantined from the workspace on `main` per `pnpm-workspace.yaml`, pre-existing and unrelated)
- `pnpm build`, `pnpm typecheck`, `pnpm lint`: green
- dist smoke (opt-out envs unset, endpoint unset, temp HOME, injected fetch): `createAnalytics({source:'cli',sdkName:'x',sdkVersion:'1'}).enabled === true` and the batch delivered to `https://api.sapiom.ai/v1/analytics/collector`; with `SAPIOM_TELEMETRY_DISABLED=1`, `enabled === false`

Refs SAP-1415 / SAP-1442
https://linear.app/sapiom/issue/SAP-1415

## Consumer packages: CI guards + honest tests (added after the instrumentation PRs merged)

The flip changes what an *unconfigured* emitter does, so every instrumented package's test environment needed hardening before this could merge:

- **Global test guards** — all six jest packages (tools, cli, agent-core, agent-runtime, langchain, langchain-classic) get a `setupFiles` guard setting `SAPIOM_TELEMETRY_DISABLED=1`; mcp gets the same via vitest `test.env`. No test can dial the production collector; suites that assert delivery opt back in explicitly against the in-process mock via `SAPIOM_ANALYTICS_ENDPOINT`.
- **"Ships dark" tests rewritten to the live-default invariant** — the old "no endpoint → nothing sent" tests now assert delivery-by-default (against the mock) and that consent opt-outs are the gate producing zero requests. The tools tri-state identical-throw property now covers on / default-live / disabled. Fault-injection baselines that would have dialed production were repointed at the mock.
- **Docstring/README sweep** — every "ships dark" claim across the seven consumer packages updated to the live-by-default truth (opt-outs: `SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, programmatic).
- **No consumer changesets** — their diffs are tests/docs/config only; the behavior change ships via the `@sapiom/analytics-core` minor bump.

Suites at final state: analytics-core 152, tools 360, cli 39, mcp 70, agent-core 115, agent-runtime 26, langchain 34, langchain-classic 180 — all green. Internal review round documented in the comments.

### DIFF
--- a/.changeset/hot-poets-shine.md
+++ b/.changeset/hot-poets-shine.md
@@ -2,4 +2,4 @@
 "@sapiom/analytics-core": minor
 ---
 
-Analytics delivery is now on by default to the hosted collector (the ship-dark default flipped live): with no `endpoint` configured, `createAnalytics` delivers to `SAPIOM_COLLECTOR_ENDPOINT`. Opt-outs unchanged (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, programmatic `disabled: true`) — any of them still makes the emitter a complete no-op: zero network calls, zero disk writes.
+Analytics delivery is now on by default to the hosted collector (the ship-dark default flipped live): with no `endpoint` configured, `createAnalytics` delivers to `SAPIOM_COLLECTOR_ENDPOINT`. Opt-outs unchanged (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, programmatic `disabled: true`) — any of them still makes the emitter a complete no-op: zero network calls, zero disk writes. Consent providers returning `undefined` deliver by default; return `false` to stay off.
--- a/.changeset/hot-poets-shine.md
+++ b/.changeset/hot-poets-shine.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/analytics-core": minor
+---
+
+Analytics delivery is now on by default to the hosted collector (the ship-dark default flipped live): with no `endpoint` configured, `createAnalytics` delivers to `SAPIOM_COLLECTOR_ENDPOINT`. Opt-outs unchanged (`SAPIOM_TELEMETRY_DISABLED=1`, `DO_NOT_TRACK=1`, programmatic `disabled: true`) — any of them still makes the emitter a complete no-op: zero network calls, zero disk writes.
--- a/packages/agent-core/jest.config.js
+++ b/packages/agent-core/jest.config.js
@@ -1,6 +1,10 @@
 module.exports = {
   testEnvironment: 'node',
   passWithNoTests: true,
+  // Telemetry defaults live; disable it globally so no test emits to the real
+  // production collector. Tests that assert events ARE sent must opt in by
+  // setting SAPIOM_ANALYTICS_ENDPOINT to the mock collector.
+  setupFiles: ['<rootDir>/jest.telemetry-guard.js'],
   roots: ['<rootDir>/src'],
   testMatch: ['**/__tests__/**/*.ts', '**/?(*.)+(spec|test).ts'],
   transform: {
--- a/packages/agent-core/jest.telemetry-guard.js
+++ b/packages/agent-core/jest.telemetry-guard.js
@@ -1,0 +1,4 @@
+// Telemetry defaults live after the ship-dark flip; tests must opt in explicitly
+// via the mock collector (SAPIOM_ANALYTICS_ENDPOINT). This guard ensures no
+// test suite can accidentally emit to the real production collector.
+process.env.SAPIOM_TELEMETRY_DISABLED = "1";
--- a/packages/agent-core/jest.telemetry-guard.js
+++ b/packages/agent-core/jest.telemetry-guard.js
@@ -1,4 +1,4 @@
-// Telemetry defaults live after the ship-dark flip; tests must opt in explicitly
-// via the mock collector (SAPIOM_ANALYTICS_ENDPOINT). This guard ensures no
-// test suite can accidentally emit to the real production collector.
+// Telemetry is live by default: an unconfigured emitter delivers to the real
+// production collector. Disable it globally here; tests that assert delivery
+// opt back in explicitly via the mock collector (SAPIOM_ANALYTICS_ENDPOINT).
 process.env.SAPIOM_TELEMETRY_DISABLED = "1";
--- a/packages/agent-core/src/__tests__/analytics-events.test.ts
+++ b/packages/agent-core/src/__tests__/analytics-events.test.ts
@@ -394,7 +394,7 @@ describe("local run step events", () => {
   });
 });
 
-describe("consent and ship-dark", () => {
+describe("consent and live-default", () => {
   it.each(["SAPIOM_TELEMETRY_DISABLED", "DO_NOT_TRACK"])(
     "%s=1 → zero collector requests, identical results",
     async (envVar) => {
@@ -423,18 +423,20 @@ describe("consent and ship-dark", () => {
     },
   );
 
-  it("no endpoint configured (ships dark) → zero requests, identical results", async () => {
-    delete process.env.SAPIOM_ANALYTICS_ENDPOINT;
-    resetOrchestrationAnalyticsForTesting();
-
+  it("live by default: with no explicit endpoint config, emitter is enabled and events reach the collector", async () => {
+    // No explicit `endpoint` on the analytics config, but the env override
+    // (SAPIOM_ANALYTICS_ENDPOINT) set by beforeEach points at the mock so
+    // nothing hits the production URL. This exercises the live-default
+    // fall-through path: resolveEndpoint() returns the hosted collector when no
+    // config is set, and the env override redirects that to the mock.
     const gateway = await startFakeGateway(HAPPY_ROUTES);
     try {
       const { started } = await runHappyFlow(gateway.host);
       expect(started.executionId).toBe("exec_1");
-      expect(getOrchestrationAnalytics().enabled).toBe(false);
+      expect(getOrchestrationAnalytics().enabled).toBe(true);
 
       await getOrchestrationAnalytics().flush();
-      expect(collector.requests).toHaveLength(0);
+      expect(collector.requests.length).toBeGreaterThan(0);
     } finally {
       await gateway.close();
     }
--- a/packages/agent-core/src/analytics.ts
+++ b/packages/agent-core/src/analytics.ts
@@ -1,9 +1,8 @@
 /**
  * Usage-analytics plumbing for the orchestration operations (deploy / run /
  * link / run-local). Events are emitted through `@sapiom/analytics-core`,
- * which ships dark by default: unless a collector endpoint is explicitly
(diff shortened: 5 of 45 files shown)


evtran0209:
**⚠️ MERGE ORDER: this PR merges LAST — after #253, #249, #252, #248, #250, #254.**

Rationale: the cross-package audit here ('no package depends on analytics-core yet') is true on this PR's base, but the six instrumentation PRs above each add the dependency, and their fault-injection suites include 'ships dark: no endpoint → zero fetch calls' assertions that invert once this default flips. Merging this first (or without a post-merge rebase) turns main red on those suites.

Plan: once the instrumentation PRs land, this branch gets rebased and each consumer package's dark-test flips to the same pattern used here (no-op proven via opt-out envs / disabled:true instead of endpoint absence). That rebase commit will be pushed to this PR before it's marked ready.

evtran0209:
## Review triage (FIX-FIRST) — all items addressed in c2c5393

**Taken as fixes:**

- **Issue 6 (never-silent gap) — FIXED.** The first-run notice is decoupled from marker persistence: when identity storage is unavailable (`identityStore.load()` → `null`, e.g. unwritable HOME), the notice now prints unconditionally on the first track instead of being suppressed while delivery proceeds. The marker cannot persist in that edge, so repeating the notice every process start is acceptable and honest — silent collection is not. Added the test "unwritable HOME + enabled → first-run notice fires on first track", and the existing unwritable-HOME test's `noticeCount` pin is updated 0 → 1 (title now says "notice still prints").
- **Issue 3 (real-fetch ECONNREFUSED egress risk) — FIXED.** The env-override line now carries the do-not-remove comment (without it the live default is the production collector and the test dials real network), plus a structural guard on top: the test wraps the real global fetch with a pass-through recorder that refuses `SAPIOM_COLLECTOR_ENDPOINT` outright and asserts no dialed URL equals it. Losing the override now fails the test loudly with zero real egress, while the loopback path still exercises a genuine ECONNREFUSED.
- **Issue 1 (mid-frame env deletions) — FIXED.** The unpin is kept only in the three tests whose point requires the default fall-through (empty-string → hosted default, no-endpoint → hosted default, opt-out-beats-the-live-default), each annotated with a one-liner stating the injected `fetchImpl` is the egress guard in that scope. The "explicitly passing SAPIOM_COLLECTOR_ENDPOINT" test no longer unpins — the `beforeEach` pin stays in place, since a config endpoint beats the env override (pinned by the adjacent precedence test).
- **Issue 5 (changeset) — FIXED** with the suggested wording: "Consent providers returning `undefined` deliver by default; return `false` to stay off."

**Resolved as documented-by-design (no code change):**

- **Issue 4 (`consentProvider` returning `undefined` → default ON).** This is the project's designed consent precedence: programmatic `disabled: true` > env opt-outs (`SAPIOM_TELEMETRY_DISABLED`, `DO_NOT_TRACK`) > provider decision > default ON. `undefined` (or a throw) means the hook has *no opinion*, and a no-opinion hook must behave identically to having no hook at all — making `undefined` mean OFF would silently flip the default for every caller that wires a provider as a stored-consent passthrough. `resolveConsent` is unchanged; the behavior is now documented loudly in the `consentProvider` JSDoc ("returning `undefined` defers to the default, which is ENABLED — return `false` to disable; do not use `undefined` as a 'not decided yet' sentinel"), the README opt-out section, and the changeset.

**Acknowledged:**

- **Issue 2 (ship-dark parenthetical) — retraction acknowledged;** dropped the historical jargon from the public source anyway ("(This is the pre-planned ship-dark flip.)" removed from `http-sender.ts`).

**Verification:** analytics-core 8 suites / **152 tests passed** (151 + the new unwritable-HOME notice test); repo `pnpm build` / `pnpm typecheck` / `pnpm lint` green; leak grep on touched files clean.

evtran0209:
**Internal review round (full PR at final state, head 646edf9)** — verdict: **MERGE-READY** after two nit fixes (stale `ship-dark` comment wording in the six telemetry-guard files + one section header; applied in 646edf9).

What the round verified, beyond the nits:
- **Flip correctness**: `resolveEndpoint` precedence (config > `SAPIOM_ANALYTICS_ENDPOINT` env > hosted default), empty-string config falls through; 5s request timeout, ≤1 retry, never-throw all intact; first-run notice fires even when identity storage fails (never silent delivery).
- **CI can never dial production** (adversarial pass): all six jest packages guarded via `setupFiles` (runs before module import), mcp via vitest `test.env`; both child-process spawn surfaces build env explicitly (cli's `cliEnv()` strips `SAPIOM_*`; mcp's `spawnServer()` uses a minimal allowlist) and every spawn site pins the mock endpoint; analytics-core's own suites pin a loopback endpoint per-test; packages without analytics-core imports need no guard — none missed.
- **Rewritten tests assert the live-default invariant** (delivery-by-default against the mock; tri-state identical-throw now covers on/default-live/disabled), not just renamed titles.
- **Changeset**: `minor` on 0.x is right for "telemetry now sends by default"; consumers receive it transitively — no consumer changesets needed (their diffs are tests/docs only).

Suites at final state: analytics-core 152, tools 360, cli 39, mcp 70, agent-core 115, agent-runtime 26, langchain 34, langchain-classic 180 — all green.

gwitwer (approved):
Approved. Reviewed the ship-dark flip for correctness and safety, and the opt-out-by-default posture is an intended, signed-off product decision (the hosted collector is live; this is the planned flip).

Code review:
- Consent is airtight as the now-sole gate: resolveConsent runs BEFORE endpoint resolution in buildAnalytics and fails CLOSED (any thrown error → createDisabledInstance). All four opt-outs (disabled:true, SAPIOM_TELEMETRY_DISABLED=1, DO_NOT_TRACK=1, init failure) yield a complete no-op — zero fetch, zero disk, no first-run notice. The removed 'no endpoint → disabled' branch was genuinely unreachable post-flip and drops no consent check.
- resolveEndpoint precedence (config → env → hosted default), empty-string-as-absent, and the string|null→string narrowing are all correct.
- CI safety verified across all 8 touched packages: every emitter-constructing test pins a loopback endpoint or injects a capturing fetch, plus a guard that rejects any request to the production URL. The two previously-dangerous paths (analytics-core 'no endpoint' test, mcp spawnServer({})) were repointed.
- First-run notice still fires on first real event (even when identity storage is unwritable) and is suppressed when disabled; never-throws preserved; docs + changeset consistent.

Non-blocking follow-ups:
1. PR description's first section is stale ('no package depends on analytics-core yet') — 6 packages now do; the appended 'CI guards + honest tests' section is the correct story. Please update so future readers aren't misled.
2. A throwing consentProvider now defaults to ENABLED (was masked by the null endpoint under ship-dark). It's documented in types.ts as intentional, but it's a behavioral shift for any host relying on throw-as-off — worth a line in the changelog.
3. The global jest/vitest telemetry guards don't reach spawned child processes; safety in the cli/mcp spawn suites rests on every call-site passing the mock endpoint (all current ones do). Consider having those spawn helpers default-inject SAPIOM_TELEMETRY_DISABLED=1 unless overridden, so a future spawn that forgets can't dial production.

PR #177, opened 2026-07-08
feat(harness): package scaffold + pty session runtime + fake-agent fixtures

evtran0209:
## Summary

New `@sapiom/harness` package (v0.1.0): the session foundation for the agent harness — full package scaffold, the `SessionRuntime` abstraction, a node-pty backed `PtyRuntime`, and the transcript-driven fake-agent fixture app that all harness tests build on.

Fixes SAP-1407 — https://linear.app/sapiom/issue/SAP-1407

## Changes

**Package scaffold** (`packages/harness/`)
- `src/{runtime,harnesses,server,analytics,doctor,ui}/` — runtime is implemented; the other five are placeholder modules so later tickets don't collide on scaffolding
- Build/test/lint shaped like `packages/tools` (dual CJS+ESM tsc build, Jest + ts-jest, eslint, prettier)
- package.json already declares **server deps** (`express`, `ws`, `node-pty`) and pre-declares **UI deps** (`react`/`react-dom` 19, `@types/react*`, `@xterm/xterm`, `@xterm/addon-fit`, `@xterm/addon-attach`) so later UI tickets only add code, not dep lines. UI deps live in `devDependencies` on the assumption the UI is bundled into static assets at build time; if a later ticket needs them at runtime, moving the lines is trivial.

**SessionRuntime + PtyRuntime** (`src/runtime/`)
- `SessionRuntime` interface exactly per the agreed contract: `create/write/onData/resize/kill/isAlive` with an opaque `SessionHandle`
- `PtyRuntime` (node-pty): env/cwd/cols/rows plumbed through; `onData` returns an unsubscribe; `kill()` is graceful (SIGTERM → SIGKILL after a configurable timeout); handles stay queryable after exit (`isAlive` → false); unknown handles throw a typed `UnknownSessionError`
- **Lazy native import**: node-pty is imported inside `create()`, never at module load. A missing/broken native build rejects with a typed `PtyUnavailableError` (`code: "PTY_UNAVAILABLE"`, remediation hint, original cause) that a future doctor can consume
- **Prebuild repair**: node-pty 1.1.0's published prebuilds ship `spawn-helper` without the executable bit (mode 644 in the npm tarball), so on macOS every spawn dies with `posix_spawnp failed.` The runtime chmods it best-effort before spawning; no-ops on Windows and on from-source builds (Linux/CI compiles from source — there are no linux prebuilds). Verified for real: the test suite was run with the helper reset to 644 and repaired itself.

[...]

PR #258, opened 2026-07-09
test(harness): transcript-fixture fake-agent for e2e/sim replay (SAP-1436)

evtran0209:
## What

Ports the transcript-driven fake-agent pty app + 4 transcript fixtures from the closed scaffold PR #177 into the shipped harness's test machinery, plus a 7-test vitest suite (`src/core/transcript-replay.test.ts`) exercising `SessionManager` against them with real node-pty: replay fidelity, cwd/env/size plumbing, resize handling, and SIGTERM-resistant kill escalation (`stubborn` transcript).

## Provenance honesty

The transcripts are **hand-written/hand-modeled, not recordings**: `claude-code-boot.json` is modeled on Claude Code v2.x boot output with fictional identifiers, and says so in its `modeledOn`/`notes` fields. The fixture README documents the schema (`synthetic` = hand-written vs verbatim recording; `modeledOn` = what a hand-written transcript imitates) and a provenance table. A genuinely recorded transcript is a good future upgrade.

## Deviation from the ticket AC, documented

The AC asked for an existing e2e phase to consume the fixtures. `e2e:live` phases depend on the agent firing real `SessionStart` hook POSTs to flip `session.ready`; `fake-agent.cjs` is deliberately a pure pty app with no HTTP client, and force-flipping readiness from the script would defeat the ready-gate invariant those phases protect. The vitest suite covers the same SessionManager pty surface end-to-end instead — rationale recorded in the fixture README (§ Test consumer) and the test-file header.

## Safety

- Tests/fixtures only — no runtime changes, no changeset (`files: ["dist"]` keeps `e2e/` and tests out of the npm tarball, verified via `npm pack --dry-run`)
- `describe.skipIf` guards the suite where node-pty's native module is unavailable (skips with an actionable message instead of erroring)
- 611 harness tests green; build green

Internal review round: 3 findings (e2e-phase scope gap, provenance-flag contradiction, missing pty guard) — all addressed in `182b5cb`.

### DIFF
--- a/packages/harness/e2e/fixtures/fake-agent/README.md
+++ b/packages/harness/e2e/fixtures/fake-agent/README.md
@@ -1,0 +1,65 @@
+# fake-agent
+
+A tiny transcript-driven terminal app used as a stand-in for a real coding
+agent in harness tests. It is designed to run inside a pty (spawned by the
+session manager) and behaves like an agent TUI: raw-mode input, a boot
+banner, an idle prompt, keystroke echo, and busy/idle output cycles.
+
+## Usage
+
+```bash
+node fake-agent.cjs [path/to/transcript.json]
+```
+
+Defaults to `transcripts/basic-echo.json`.
+
+The `.cjs` extension is required because the harness package declares `"type": "module"`, which would cause Node to treat a `.js` file as an ES module. The fake-agent uses CommonJS `require()` so it must be `.cjs`.
+
+## Transcript format
+
+Transcripts are JSON files:
+
+```jsonc
+{
+  "name": "basic-echo",
+  "synthetic": true,          // true = hand-written; false = recorded from a real agent
+  "description": "…",
+  "echoKeystrokes": true,     // echo bytes as they are typed (default true)
+  "announceResize": false,    // print "[resize] COLSxROWS" on terminal resize
+  "ignoreSigterm": false,     // ignore SIGTERM (lets tests exercise forced kill)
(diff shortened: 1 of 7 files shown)


gwitwer (approved):
Approved. Clean test-only harvest of #177's fake-agent (SAP-1436), verified:
- Truly tarball-safe: package.json files:["dist"] keeps e2e/ fixtures and *.test.ts out of the published npm tarball (consistent with the 0.1.0 tarball). No runtime change, no changeset — correct.
- Test quality is genuine: 6 direct fake-agent pty tests (boot banner, echo-after-spinner, cwd/env/cols/rows plumbing, resize propagation, SIGTERM-resistance→SIGKILL escalation, full claude-code-boot replay) + a SessionManager integration test exercising create→attach→write→kill against real node-pty. describe.skipIf(!nodePty) guards missing-native-module envs with an actionable message.
- The AC deviation is the correct call and well-documented: fake-agent.cjs fires no hook events, so wiring it into e2e:live and force-flipping session.ready would defeat the ready-gate invariant those phases protect. The vitest suite covers the same SessionManager pty surface end-to-end instead.
- Provenance is honest: transcripts are labeled synthetic/hand-modeled (modeledOn/notes + README table), not passed off as recordings; a real recording is flagged as a future upgrade.
Note: this closes SAP-1436, which overlaps a ticket we filed under SAP-1397 — worth a quick dedupe so the same harvest isn't tracked twice.

PR #259, opened 2026-07-09
feat(harness): adopt @sapiom/analytics-core for remote telemetry (SAP-1449, SAP-1440)

evtran0209:
## Why

The harness's bespoke `CollectorBatcher` posts to `${SAPIOM_COLLECTOR_URL}/v1/harness/events` — an endpoint that does not exist. Remote harness telemetry has been silently dropping for every `@sapiom/harness@0.1.x` user (local `events.ndjson` capture unaffected). This PR replaces that wire with `@sapiom/analytics-core`, the emitter every other Sapiom package now ships — one collector endpoint, one envelope, one identity.

## What

**analytics-core (additive only, patch):**
- `"harness"` joins the `EventSource` union; CONTRACT.md gains the source-enum entry + taxonomy row
- `seedAnalyticsIdentity(anonymousId)` export — seeds `~/.sapiom/analytics.json` iff absent (idempotent, 0600, degrades on unwritable HOME)
- `discard()` on `SapiomAnalytics` — drop buffered events without sending (see opt-out semantics below)

**harness (minor):**
- `createHarnessEmitter` replaces `CollectorBatcher` (deleted — confirmed never exported from the public entry): one process-shared emitter, per-event `session_id` override, `data` carries `seq`/`agent_session_id`/`harness_kind`/`tenant_id`/`context` exactly as before
- **Local-always invariant preserved**: the ndjson store writes unconditionally; consent gates remote only
- **Opt-out drops the buffer** — the shipped batcher wiped its queue on `setTelemetryOptIn(false)`; the swap preserves that privacy semantic via `discard()` (opt-in and API-key rotation still drain gracefully). Pinned by a ported test
- **Identity convergence (SAP-1440)**: one-way boot migration `~/.sapiom/harness/machine-id` → `~/.sapiom/analytics.json` — existing installs keep their id (longitudinal join key), legacy file untouched, migration failure never blocks server start
- Test-env guard (`SAPIOM_TELEMETRY_DISABLED=1` via vitest setup) matching the pattern all seven SDK packages adopted with the live-default flip
- `scripts/e2e-sim.ts` + `scripts/mock-collector.mjs` rewritten to the analytics-core wire

## Evidence

- `sim:e2e`: **PASS** — 24/24 assertions (ndjson store + envelope shape incl. monotonic `seq`)
- `e2e:live`: **all 3 phases pass**, including step 7 (batch reaches the mock collector)
- Suites: analytics-core **157**, harness **610** — green
- Two internal review rounds: round 1 caught a stale ship-dark guard assumption (built pre-flip), CONTRACT.md gap, dead-code batcher; round 2 caught the e2e scripts still importing the deleted batcher and the opt-out drain regression. All addressed.

Closes SAP-1449, SAP-1440.


gwitwer (approved):
Approved. Reviewed the transport adoption and it's correct:
- Opt-in gate airtight: createHarnessEmitter sets disabled:!telemetryOptIn, and disabled:true is the highest-precedence off switch in analytics-core — consent-off is a complete no-op regardless of the analytics-core default. telemetryOptIn is always resolved before the emitter is built.
- Local-always invariant preserved: ndjson write is before/independent of the emitter and never consent-gated.
- Identity migration (SAP-1440) is safe: one-way machine-id → ~/.sapiom/analytics.json at boot, existing ids kept, wrapped so failure never blocks server start, idempotent, 0600, tested.
- CollectorBatcher deletion clean (never publicly exported; e2e/mock scripts rewired); per-event session_id/event_id overrides + data (seq/agent_session_id/harness_kind/tenant_id/context) preserved; test-setup guards tests off prod; changesets correct (analytics-core patch, harness minor, additive).

On posture: on-by-default is the intended behavior for the current launch phase (internal users), so approving as-is. Follow-up (separate, tracked): make the default telemetry posture a single documented switch so it can be flipped to off-by-default later without code surgery — the DEFAULT_TELEMETRY_OPT_IN constant is effectively that switch today; making it env-readable (e.g. a SAPIOM_HARNESS_TELEMETRY_DEFAULT) would let it flip without a release. Nits: data.tenant_id omitted-when-null vs agent_session_id null-coalesced (harmless asymmetry); no explicit flush on clean server shutdown (relies on analytics-core beforeExit).

PR #266, opened 2026-07-09
feat(harness): v0 hardening + analytics connection — integration branch

evtran0209:
## What this is

The complete v0 harness workstream — **15 internally-reviewed units** merged into one integration branch, tested together, then swept by a **five-reviewer final pass over the entire 126-file diff** with all findings fixed. One approval ships everything.

| Unit | PR | Substance |
|---|---|---|
| Telemetry connection | #259 | Bespoke batcher (posting to a dead endpoint — remote telemetry silently dropped for all 0.1.x users) → `@sapiom/analytics-core`; drop-on-opt-out preserved; identity converged on `~/.sapiom/analytics.json` |
| Typed errors | #261 | `HarnessError` + coded subclasses; no string-matched HTTP dispatch; wire byte-identical |
| Hygiene | #263 | Registry write serialization, background-macro dedupe, async watcher poll |
| Canvas bugs | #264 | Instant deterministic render stays visible under enrichment overlay; deduped Visualize is a true no-op |
| `sapiom dev` | #265 | CLI launches the harness; optional peer; resolution proven via simulated global install |
| Packaging | #267 | LICENSE, exports map, ESM-only decision (protects `instanceof` dispatch), tarball-verified |
| Prompt bar | #268 | → evolved into the chat composer (see #278) |
| Exit-listener dedupe | #269 | One shared `beforeExit` listener library-wide |
| Awaitable kill | #270 | `kill()`/`killAll()` resolve on actual death; bounded; zombie-proof (both hang paths found in review, fixed, regression-tested) |
| Adapter registry | #272 | Embedded/external modes, `GET /api/harnesses`, Conductor companion scaffold with live 409s, drift-proof kind enum |
| Payment dedupe | #274 | One analytics event per pay-gated call; `payment_retried` flag; retry flow byte-identical |
| Consent + ui.* layer | #276 | Telemetry chip, first-run notice, ndjson retention cap (serialized sweep), `POST /api/track` — every payload metadata-only |
| Skills panel | #277 | Right-pane Canvas\|Skills tabs, traversal-safe `/api/skills`, registry-sourced Install-MCP, shell-injection-safe macros |
| **Chat-first session view** | #278 | The headline UX: conversation replaces the terminal as the center; transcript-tail turn sources; slash passthrough; permission banner; terminal demoted to a keep-alive tab |
| CI wiring | #279 | Playwright mock tier in PR CI (cached Chromium, 15-min cap), local-loop docs, contract seq-gaps amendment |

## Final whole-PR review (five parallel reviewers at final state)

Server-core, collector/analytics, web-UI, packaging/CI/docs, plus a cross-unit integration pass. **Collector/analytics: fully clean** — privacy boundaries verified with precise semantics: the hook→analytics path carries disclosed, consent-gated session content (prompts, tool I/O, assistant text — the first-party model the contract documents); the transcript-tailer→WebSocket chat path is UI-transport only and never enters analytics; all `ui.*` payloads are schema-enforced metadata-only; langchain third-party redaction intact; one shared seq counter. The other four found 17 issues — all fixed in the last five commits: two 409-mapping gaps + tests, a stale-closure wrong-session input bug, terminal keep-alive genuinely broken by an HTML `hidden` attribute (now aria-hidden + computed-style spec), six TS errors invisible to CI (web typecheck now wired into the repo gate so the class can't recur), every internal ticket reference stripped from public artifacts (grep-verified zero), stale changeset/docstrings rewritten for the final architecture, duplicate markdown renderers and path encoders consolidated, the skills panel's analytics seams wired (`skill.viewed`/`skill.used`/`mcp.install` complete the taxonomy), CI job timeout, warning-color tokens.

## Verification at final state

All 15 packages build; **15/15 typechecks (now including the web app)**; lint 0 errors; harness **793/793** unit + **139** Playwright; analytics-core 159, cli 69, langchain 36+185; both harness e2e scripts pass. **Production-verified**: real human usage from this branch is already flowing to BigQuery schema-perfect (49 events, zero column defects across all 18 columns, seq duplicate-free — verified 2026-07-09).

## What merging does

Changesets aboard (**all patch by maintainer request** — the backend monorepo pins minors) → `@sapiom/harness` **0.1.2** (working telemetry + chat-first UI), `@sapiom/cli` 0.4.1 (`sapiom dev`), `@sapiom/analytics-core` 0.2.1, langchain 0.5.1. Merge → Version Packages PR → publish → post-publish harness smoke closes v0.

## v0 pivot (post-beta-testing, 2026-07-09)
Live pre-merge beta testing drove a scope correction — captured here so the diff reads true:
- **Chat-first center deprecated → terminal-only for v0.** The chat pane rendered empty in live sessions (the transcript-tailer only starts on Claude's SessionStart hook, which waits on trust/theme onboarding; no running-status fallback; never verified end-to-end). Rather than ship a hollow surface, v0 is the proven terminal — exactly as originally built. Revival (running-status fallback + live integration test) is a tracked fast-follow. **Analytics is untouched** — capture is hook-based, independent of the chat UI; collector→BigQuery keeps working and is prod-verified.
- **Install-MCP button removed** (MCP is auto-wired per session for claude-code; `/api/harnesses` kept dormant for a codex-MCP fast-follow).
- **Skills panel → read-only browser**: no "Use" button; refresh-on-tab-open so terminal-created `~/.claude/skills` entries appear; **`@sapiom/agent-core` now ships its `skills/` dir** (was missing from `files` → panel would've been empty in prod).
- **Session-exited overlay fixed**: Resume was swallowing a 409; now it's disabled with a reason when a session can't be resumed (codex-exit-1 null session-id), errors surface via toast, Close hardened — with the null-case Playwright coverage the mock tier lacked.
- Terminal no-regression verified (smoke specs: boot session renders, dead-session Resume/Close); leak gate clean (no tickets/names/internal hosts; tarballs ship dist-only + the agent-core skill). Suites: harness 770 unit + 96 Playwright, all packages build/typecheck/lint (0 warnings).


gwitwer (commented):
Reviewed the full diff (128 files) with a per-domain pass: analytics/privacy, server core & process lifecycle, endpoint security, web UI, and packaging/CI. Big picture: this is strong work and most of it is genuinely clean (details at the bottom). There's **one item worth a deliberate decision** plus a handful of confirmed bugs. Flagging, not blocking — your call on how to resolve.

## 🚩 Decision: telemetry ships content, on by default

The PR description says *"chat content and hook payloads never touch the analytics pipeline."* I don't think that's accurate for the **hook → analytics** path (it is accurate for the transcript-tailer → WebSocket path, which is clean). Traced end to end:

- `core/collector/normalizer.ts:77` puts the full **user prompt** into the payload; `:82-83` the **tool input** and **tool output** (`toolResponseSummary`, capped 16 KB — i.e. file contents, bash stdout); `:92` the **final assistant message**.
- `server/ingest.ts:138` enqueues the whole event to the remote batcher, and `core/collector/analytics-emitter.ts:82` spreads `...event.payload` onto the wire with **no field-level redaction**. Endpoint defaults to `https://api.sapiom.ai/v1/analytics/collector` (`analytics-core/src/http-sender.ts:11`), gated only by opt-in (`analytics-emitter.ts:122`).
- The deleted `batcher.ts` was a **no-op unless `SAPIOM_COLLECTOR_URL` was set**, so default installs did no remote egress. This PR moves remote egress to on-by-default (hardcoded endpoint) **and** it carries content — so the net effect is that previously-dropped content now reaches the collector by default.

To be clear on the good parts: this **is disclosed** (README telemetry section + the TTY consent prompt in `cli/consent.ts:18-19` name "the prompts you send and the tool calls your agent makes"), and opt-out is airtight end to end (`--no-telemetry` / `DO_NOT_TRACK` / `SAPIOM_TELEMETRY_DISABLED` all verified to zero out the remote path, and `discard()` drops the buffer on revoke).

The reason it's a decision rather than a nit:

1. It ships **content** (prompts, tool output that can include secrets/source, assistant text), not the anonymous-metadata-only posture that was the stated intent for launch. Worth an explicit "yes, content is intended."
2. **Non-TTY launches (CI / Docker / piped) default ON silently** (`cli/consent.ts:32-33`) with no content disclosure at all — only "defaulting telemetry to on."
3. `analytics-core/CONTRACT.md` contradicts itself: obligation #6 says non-Sapiom-bound calls send "never arguments or content," but the harness envelope example (`:47`) and taxonomy (`:176`) ship `data.prompt` + tool I/O. A user's prompt to Claude and their local bash output aren't Sapiom-bound.

If content collection is intended: recommend disclosing content on the non-TTY path, tightening the "local usage analytics" wording (it's remote too), reconciling CONTRACT #6, and updating the PR's "no content" line. If not: give the remote emitter a metadata-only projection and keep full content only in the local ndjson store.

## Confirmed bugs

**1. ChatView cross-session state leak** — `web/src/components/ChatView.tsx:111-123`, rendered without a `key` at `App.tsx:281`. `systemNotes` (which embeds your typed slash-command text, e.g. `Sent /model … to the agent`) and `scrollLocked` are instance-local and never reset when `sessionId` changes, so session A's notes render under session B's conversation, and A's scroll-lock carries into B. This is the same "wrong-session" class the final sweep was chartered to catch, just in ChatView's local state rather than the input path. Fix with a reset effect on `sessionId` change — **not** `key={sessionId}` (that would remount PromptBar and destroy its per-session `draftsRef`).

**2. Shutdown timers leak (ref'd, never cleared)** — `server/index.ts` `close()` (~L1038) and `core/task-manager.ts:386,401`. The 2–5s `setTimeout`s in the `Promise.race` against `killsSettled`/`allExited` aren't `unref()`'d or `clearTimeout`'d, so when the kill wins (the common case — ptys die in ms) the timer stays pending and ref'd and holds the event loop 2–5s after a clean shutdown. `SessionManager.kill()` handles this correctly (unref'd escalation timers) — these two spots should mirror it. Undercuts the "process actually exits" goal and slows test teardown.

**3. `killAll()` never runs on Ctrl+C** — `cli/bin.ts main()` registers no `SIGINT`/`SIGTERM` handler and never calls `server.close()` outside tests, so the newly-awaitable `killAll()` doesn't fire on a real interrupt and the claude/codex ptys are still orphaned — the exact case `close()`'s comment says it fixes. Pre-existing (no signal wiring changed here), but it means the headline awaitable-kill benefit is inert in the primary CLI path. A `process.once('SIGINT'/'SIGTERM', () => void server.close()…)` is the missing wire — confirm whether that's intentionally out of scope.

**4. Permission attention banner may never clear** — `web/src/lib/use-chat-state.ts:105-112` vs the prop doc at `ChatView.tsx:43-47`. The doc says the banner is cleared by subsequent hook activity (PostToolUse/Stop/UserPromptSubmit), but the reducer only writes `attentionMessage` on a `chat.attention` event — so it clears only if the server emits `chat.attention` with an empty message. If resume is signaled via `chat.turn`/`chat.tool`, the "⚠ permission pending" banner sticks over a working agent. Either clear it on the next turn/tool for that session, or confirm the server always emits an empty `chat.attention` and fix the doc.

## Minor / suggestions

- `/api/track` accepts `data: z.record(z.unknown())` (`server/rest.ts:70`) and forwards it verbatim into the remote payload (`:461`) — metadata-only rests on caller discipline, not enforcement. Bounded by boot-token + consent, so not exploitable beyond that, but worth constraining to primitives + a key/length cap so the endpoint is provably metadata-only.
- Changeset levels: analytics-core resolves to **0.2.1**, not the 0.2.3 in the description; `analytics-core-harness-identity-seed.md` is `patch` but adds public API (`seedAnalyticsIdentity`, `discard()`, `"harness"` source) — semver-minor. Benign given `workspace:^`, just cosmetically low.
- `AdapterNotFoundError` is documented as 400 (`core/errors.ts:11`) but no handler maps it, so it hits Express's default 500. Reachable via a persisted `sessions.json` with an unknown `harness` kind.
- `workspace-watcher.ts` poll (`setInterval` → `checkNowAsync`) has no in-flight guard — on a slow/wide tree two walks can overlap and double-fire `onChange`.
- `workflow-registry.ts:121` `persist()` writes in place but the new comment says it mirrors `SessionManager.persist()`, which is tmp-file + rename (atomic). A crash mid-write can tear `workflows.json`. Adopt tmp+rename or soften the comment.
- `skills.ts:11-14` doc claims the detail endpoint reuses a list-time path set to avoid walking the FS, but `:216-221` re-runs the full scan every request. Doc vs code mismatch (+ minor token-gated DoS amplification).
- Unused `useEffect` import in `use-chat-state.ts:13`.
- Full history is dropped (not merged) if a live `chat.turn` beats the `chat.history` snapshot on reconnect (`use-chat-state.ts:98`) — dedup is already `turnId`-keyed, so a merge would avoid the data loss without reintroducing dupes. Deliberate tradeoff today; noting it.

## Verified sound (pushed on these adversarially)

Skills path-traversal (slug regex + the id never reaches a path sink — stronger than realpath); macro shell-injection escaping (correct POSIX single-quote, and only `{{workflow.path}}` reaches a shell — though `name`/`definitionId` are unquoted, fine for current macros); `/api/*` boot-token gating with no new unauthenticated surface; Markdown XSS (no `dangerouslySetInnerHTML`, all untrusted text escaped as React children); opt-out end-to-end; shared seq counter (duplicate-free, gaps expected); identity migration (UUID-shape only, no PII, benign race); payment-retry one-event dedup (byte-identical retry, metadata-only event); shared `beforeExit` listener; exports map / bin / ESM-only; CI gating (web typecheck now in the repo gate, Playwright fails the job); and **public-artifact hygiene — independently grepped the whole diff for ticket refs / names / internal hosts, zero in shipped files.** The three headline UX fixes (terminal keep-alive via visibility+ResizeObserver, resume dedup via turnId, PromptBar per-session drafts) are all genuinely correct.

Nice work overall — the security surfaces and the analytics privacy *mechanics* (opt-out, dedup, seq) are tight. The content-vs-metadata question is the one thing I'd want settled before this goes out.

evtran0209:
**Review response — all items resolved** (commits `95bfcd1`…`b2f7226`, head `b2f7226`). Thank you for the depth here; the hook-path trace was exactly right.

## 🚩 The decision — confirmed: content collection is intended
The harness is a first-party product surface; consent-gated session content (prompts, tool I/O, assistant text) reaching the collector is the designed behavior, per the contract's harness envelope from day one. Your trace stands and the fixes make it defensible everywhere:
- **Non-TTY disclosure** (`95bfcd1`): the full what-we-collect notice was in fact printing before the isTTY branch, but the defaulting line didn't acknowledge it — it now references the notice and lists all opt-out paths inline.
- **Wording**: consent copy now says analytics are captured locally AND sent to Sapiom when telemetry is on; TelemetryNotice gained the same content-scope accuracy. README untouched.
- **CONTRACT.md obligation #6 reconciled**: explicit first-party carve-out — metadata-only governs SDK wraps of non-Sapiom-bound calls; the harness's disclosed consent-gated content capture is scoped as the first-party exception (+ changeset).
- **PR description privacy line corrected** to state the precise semantics per path.

## Confirmed bugs — all fixed (`9b7e072`)
1. **ChatView cross-session leak**: sessionId-keyed reset effect for `systemNotes`/`scrollLocked` — per your warning, NOT a key-remount, so PromptBar's per-session drafts survive. Playwright: notes/scroll isolation across switches.
2. **Shutdown timer leaks**: race timers in `close()` and `TaskManager.killAll()` now unref'd + cleared when the kill wins, mirroring `SessionManager.kill()`.
3. **Ctrl+C → killAll**: `bin.ts` registers once-handlers for SIGINT/SIGTERM → awaits `server.close()` (already 5s-bounded) → exits 130/143; double-fire guarded; unit-tested with a mocked server.
4. **Attention banner**: you found the real gap — the server does emit empty `chat.attention` on hook activity, but `chat.turn` from the tailer bypasses that path. Belt-and-braces: the reducer now also clears on `chat.turn`/`chat.tool`; prop doc aligned; Playwright covers turn-clears-banner.

## Minors — all taken (`cabd717`)
`/api/track` data is now schema-enforced metadata-only (primitives, key/length caps) rather than caller discipline · `AdapterNotFoundError` → 400 mapped + tested via unknown persisted kind · watcher poll in-flight guard · `workflows.json` writes now atomic tmp+rename (matching the comment's claim) · skills detail endpoint is O(1) direct resolution (doc fixed) · unused import dropped · **history/live merge** implemented (turnId-dedup union — the snapshot is no longer dropped when a live turn wins the race) · version math: you're right, analytics-core lands at **0.2.1**; description corrected. Changesets are now all-patch per your monorepo-pinning request.

Suites at head: harness **814** unit (21 new), analytics-core 159, Playwright 139 — green. CI running.

gwitwer (dismissed):

positiveblue (approved):
LGTM

PR #273, opened 2026-07-09
feat(tools): read-only vault namespace — ctx.sapiom.vault (SAP-1471)

ynadge:
## Summary
Adds the READ-ONLY `vault` namespace to `@sapiom/tools` per the recorded SAP-1471 decision: `vault.list` (key names only), `vault.get` (one value; 404 → `null`), `vault.getMany`/`vault.getAll` (key→value maps), bound as `ctx.sapiom.vault`. **No set/delete** — write access to a secret store is a separate, deliberate risk decision; writes stay in the dashboard / `@sapiom/core`'s `VaultAPI`.

Sibling backend PR (sapiom/Sapiom#3050) adds the matching read-only MCP tools (`sapiom_vault_list`/`sapiom_vault_get`).

## Changes
- New self-contained `packages/tools/src/vault/` (own `errors.ts` + `resolveServiceUrl("vault", SAPIOM_VAULT_URL)`, no `@sapiom/core` cross-import) — the `memory` namespace recipe.
- **v2 endpoints only** (`/v2/secrets/:ref[/:key]`) — v1 is the separate legacy store that silently returns `{}` with HTTP 200 (the engine-#2413 bug class); documented in the module header.
- `client.ts`: `readonly vault` on the `Sapiom` interface + transport binding.
- `stub/index.ts`: vault stubs return empty/absent — local runs never surface real credentials.
- Barrel exports + changeset (`@sapiom/tools` minor).

## Testing
- [x] New `vault/index.spec.ts` — 10 cases (URL/encoding/method assertions via scripted Transport, 404→null, error mapping, empty-subset `keys=`, client binding + credential injection)
- [x] `pnpm --filter @sapiom/tools test` — 370 passed (20 suites, incl. the stub-completeness path)
- [x] typecheck + lint clean

## Related
- Refs: SAP-1471 (Capability Consistency project; read-only decision recorded in the ticket 2026-07-09)

### DIFF
--- a/.changeset/vault-read-namespace.md
+++ b/.changeset/vault-read-namespace.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": minor
+---
+
+Add a READ-ONLY `vault` namespace (`vault.list/get/getMany/getAll` + `ctx.sapiom.vault`) against the vault gateway's v2 API. List returns key names only; get maps a 404 to `null`. No set/delete by decision (SAP-1471) — writing secrets stays in the dashboard / `@sapiom/core` `VaultAPI`.
--- a/packages/tools/src/client.ts
+++ b/packages/tools/src/client.ts
@@ -129,6 +129,7 @@ import type {
   Memory,
   MemoryCallOptions,
 } from "./memory/index.js";
+import * as vault from "./vault/index.js";
 
 export interface Sapiom {
   readonly sandboxes: {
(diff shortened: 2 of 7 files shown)


positiveblue:
I think we can avoid adding comments about linear in our open SDK `No set/delete by decision (SAP-1471)` but this is ready to ship 🛥️

PR #74, opened 2026-06-17
feat(orchestration): secrets/secretsRef for vault injection (SAP-786)

floriangolling:
## Summary

Adds an optional `secrets` field to `@sapiom/orchestration` definitions + the workflow manifest — the **SDK half** of SAP-786 (vault secret injection into runs). Pairs with Sapiom engine PR #2413.

`secrets` is an array of **bindings** — each `{ ref, keys }` pulls the named env-var `keys` from one vault secret-set (`ref`). A workflow can declare several to read from different sets:
```ts
defineOrchestration({
  secrets: [
    { ref: 'billing',   keys: ['STRIPE_KEY'] },
    { ref: 'analytics', keys: ['POSTHOG_KEY'] },
  ],
  // ...
})
```
Only the **names** are declared here — the engine reads the values from the vault at dispatch and injects them into the run's sandbox env. Values never live in the definition.

## Changes
- `workflow.ts` — `secrets?: readonly SecretBinding[]` on `OrchestrationDefinition` + charset validation in `defineOrchestration` (ref `/^[A-Za-z0-9._:-]+$/`, keys `/^[A-Za-z0-9._-]+$/`).
- `manifest.ts` — `SecretBinding` type + `secrets?` on `WorkflowManifest` + `workflowManifestSchema` (mirrors the charset; defaults to `[]`).
- `build-manifest.ts` — copies `secrets` into the built manifest.
- tests (`sdk.spec.ts`, `manifest.spec.ts`) + a changeset (minor).

## Compatibility
Optional; the manifest schema defaults `secrets` to `[]`, so existing manifests keep parsing. No breaking change.

## Verification
`pnpm --filter @sapiom/orchestration` → typecheck ✅ · lint ✅ · test ✅ (5 suites, 74).

Needs a **manual publish** (0.1.2) after merge so the engine can bump `ORCHESTRATION_VERSION`.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #290, opened 2026-07-14
feat(agent): declare vault secret bindings [superseded]

oramada:
## Summary

- add names-only vault secret bindings to `AgentDefinition` and `AgentManifest`
- validate bindings in `defineAgent`, `agentManifestSchema`, and `buildManifest`
- reject duplicate refs/keys, protected engine namespaces, malformed identifiers, and undeclared value-bearing fields
- default legacy manifests to an empty binding list and add a minor Changeset for `@sapiom/agent`

## Security contract

Bindings contain only a vault ref and environment-variable key names. They cannot contain secret values. `PATH`, `SAPIOM_*`, and `WORKFLOWS_*` are rejected, and key collisions are invalid rather than order-dependent.

## Verification

- `pnpm --filter @sapiom/agent typecheck`
- `pnpm --filter @sapiom/agent lint`
- `pnpm --filter @sapiom/agent test` — 5 suites, 102 tests
- `pnpm --filter @sapiom/agent build`
- `pnpm changeset status`
- built declarations export `SecretBinding`; sentinel values are absent from build output

Supersedes #74, which targets the removed/deprecated `@sapiom/orchestration` package.

Refs: SAP-786

### DIFF
--- a/.changeset/calm-lions-bind.md
+++ b/.changeset/calm-lions-bind.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/agent": minor
+---
+
+Add vault secret bindings to agent definitions and manifests. Agents declare only secret-set refs and environment-variable key names; the engine can resolve their values at dispatch without storing values in definitions or manifests.
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -56,6 +56,27 @@ A step declares the transitions it may take (`next` / `terminal` / `canFail` /
 undeclared transition is a compile error. The build reads those same declarations
 to render the orchestration graph without executing anything.
 
+## Declaring injected secrets
+
+Declare only the vault secret-set reference and the environment-variable names
+the agent requires. Secret values stay in Vault and are resolved by the workflow
+engine when it dispatches a sandbox step:
+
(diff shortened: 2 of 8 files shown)


oramada:
The only failing test is the existing offline-scaffold version guard after main's #289 release bump. The isolated fix is #291 (updates both stale fallbacks: agent 0.6.2→0.6.3 and tools 0.17.2→0.18.0). Once #291 merges, this branch should be updated from main and the test matrix rerun.

oramada:
Superseded by https://github.com/sapiom/Sapiom/pull/3201.

We changed the contract after tracing the current workflow ownership boundary: workflow env is now GUI-managed server configuration, not authored SDK/manifest data. The engine derives `workflow:<definitionId>`, keeps only configured key names in its database, stores values in Vault v2, and fetches the exact subset on every step.

That removes the need for `SecretBinding`, arbitrary refs, a manifest protocol change, or a new `@sapiom/agent` release. It also resolves the review concerns here by making old manifests irrelevant to injection and preventing clients from choosing refs at all. Closing this PR so there is one implementation path for SAP-786.

PR #312, opened 2026-07-17
fix(harness): detect MCP-tool-started runs + guard runaway polling

evtran0209:
## What & why (live-testing-caught fix, stacked on C1)

On a real **Prod Run**, the live canvas never appeared. Root cause (found by live testing): the `ExecutionDetector` only matched the CLI's `✓ Started execution <id>` line — but in the harness the coding agent starts runs via the **MCP run tool**, whose result carries the id as an **`executionId` field**. The detector is already fed that (via the tool response summary), it just didn't match the shape.

## Changes
1. **Broaden detection** — `EXECUTION_PATTERN_SOURCE` now matches both `Started execution <id>` (CLI) **and** `executionId["'\s:=]+<id>` (the run tool's structured result). One capture group; streaming/dedupe unchanged. Guarded against matching the bare word "execution" (prose/inspect lines).
2. **Runaway-poll guard** — the poll controller now stops an id after `MAX_CONSECUTIVE_FAILURES` (5) consecutive failed fetches (a success resets the streak), so a stale/malformed id can't 404 forever. Aborts from `stop()` never count.

## Testing
- typecheck ✓, lint ✓
- `execution-detector.test.ts` +4 (MCP JSON shape, separators, mixed CLI+field, negative "bare execution")
- `run-poll-controller.test.ts` +2 (stop-after-5, success-resets-streak)
- Full suite **865 passed** (lone `rest.test.ts > skills router` failure is pre-existing on the clean base).

## Review & hygiene
- 1 code-reviewer pass (APPROVE, no blockers; noted a self-limiting `executionId:null`→"null" edge that the poll-guard caps at 5 fetches).
- Added lines clean; generic ids only. Patch changeset.

## Note
This fixes the *detection* path for real Prod Runs. Live testing also surfaced a likely *separate* issue — the run-state poll failing in-harness (same credential/env symptom as the earlier slug fallback) — being diagnosed separately.

Stacked on C1 → B4 → B2 → `feat/harness-runtime-analytics`.

### DIFF
--- a/.changeset/harness-detect-mcp-runs.md
+++ b/.changeset/harness-detect-mcp-runs.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+Show the live run canvas for runs started via the agent tooling, not just the CLI. The run detector now recognizes the run tool's `executionId` result in addition to the CLI's start line, so pressing Prod Run lights up the live step graph. Also stop polling a run whose state can't be fetched after repeated attempts, so a stale or malformed id can't poll indefinitely.
--- a/packages/harness/src/core/execution-detector.test.ts
+++ b/packages/harness/src/core/execution-detector.test.ts
@@ -10,7 +10,9 @@ function makeDetector() {
 
 describe("parseExecutionIds", () => {
   it("extracts the id from the CLI's start line (ignoring the check-mark prefix)", () => {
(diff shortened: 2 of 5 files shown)


evtran0209:
Superseded by #316, which is merged to `main` — all commits from this stacked PR are already in main (verified by content diff). Closing to avoid a redundant/harmful re-merge.

PR #313, opened 2026-07-17
feat(harness): map live run status onto the canvas step graph (C1 rework)

evtran0209:
## What & why (product-owner correction to C1)

C1 shipped live status as a **separate widget** above the canvas. The intended design (and the ask) is to map it **onto the canvas step-graph itself** — each node lights up in place — and drop the widget. This delivers that:

- **Nodes light up on the canvas:** each step `<g>` gets `is-running` (pulse) / `is-passed` (green) / `is-failed` (red) with its latency, in place, as the run progresses.
- **Header badge** switches `deployed → running` (prod) / `testing` (local) while a run is active, then restores.
- **The `RunStatePanel` widget is removed.**
- **Light-up-only:** nodes light up only if the diagram has been Visualized; a run does not auto-render (confirmed acceptable).

## How it works
The canvas is a **server-rendered SVG in a `sandbox="allow-scripts"` iframe** (opaque origin → parent can't touch its DOM; `postMessage` is the only channel):
- `canvas-svg.ts` tags each node with `data-step-name` / `data-step-id`.
- `canvas-run-state.ts` holds three functions (`runStateNodeClass` pure + tested; `applyRunStateToCanvas` DOM; `bootCanvasRunState` listener) that `canvas-template.ts` **stringifies into the iframe `<script>`** (+ run-state CSS, theme tokens, reduced-motion guard).
- `CanvasPane` `postMessage`s `{type:"sapiom:run-state", steps, status, target}` to the iframe on `runView` change and on iframe `onLoad`; `App` tracks/pass `target`.

## Testing
- typecheck ✓, lint ✓, unit **873 passed** (pre-existing `rest.test.ts > skills router` only), e2e **100 passed**.
- Unit: `canvas-svg` emits `data-step-*`; `runStateNodeClass` mapping; `canvas-template` includes the injected script + CSS classes.
- **Coverage note (honest):** the DOM-application + full on-canvas flow aren't unit/e2e-tested here — no jsdom in the env, and the canvas iframe doesn't render in the Playwright mock tier. Verified manually in a real harness. The old `live-canvas.spec.ts` (which tested the removed panel) is deleted.

## Review & hygiene
- 1 code-reviewer pass → **REVISE**; all items fixed: (bug) `data-step-id` fallback now sends+uses `step.id`; (bug) selector wrapped in try/catch so a quoted step name can't abort the loop; (should-fix) `target` moved to a ref (no stale closure); (should-fix) `App` tracks `target` for local runs too; (nit) `.gitignore` trailing newline.
- **Hygiene:** caught `git add -A` sweeping in the runtime `.sapiom/` dir (real agent names + local paths + rendered HTML) — unstaged it and **added `.sapiom/` to `.gitignore`**. Staged diff scanned clean.

Stacked on the detector fix → C1 → B4 → B2 → `feat/harness-runtime-analytics`.


evtran0209:
Superseded by #316, which is merged to `main` — all commits from this stacked PR are already in main (verified by content diff). Closing to avoid a redundant/harmful re-merge.

PR #325, opened 2026-07-22
fix(harness): refresh on 401 + reconcile boot token with sk_ key

evtran0209:
## Summary

Server-side auth hardening for the Studio (bug level — the deeper session-auth redesign stays a separate track).

The Sapiom API key (`sk_…`) was captured once at CLI boot and handed to the harness server as a plain string. Two defects fell out of that snapshot:

1. **401s never recovered.** A rotated/revoked key — or the user re-running the shared login in another process — left the running server sending the stale key, so every upstream call 401'd until a full restart. There was no re-auth path.
2. **Two secrets were conflated.** The per-boot *boot token* (`X-Harness-Token`) only gates the local `/api` surface; the *API key* (`sk_…`) is what authenticates upstream Sapiom calls. Studio actions must authenticate with the held API key.

### What changed
- **New `ApiKeyProvider` (`core/api-key-provider.ts`)** — the single source of truth for the held key. Seeded from the boot identity; `refresh()` re-reads the shared credential store the CLI/MCP login writes to (`~/.sapiom/credentials.json`). Silent re-read of already-cached credentials — no interactive browser re-login from the server (out of scope).
- **Refresh-on-401 in the live-run status fetcher (`core/run-state.ts`)** — a `401`/`403` triggers exactly one refresh + retry; if the store has a newer key, the retry uses it and the Studio recovers in place. Any other status (or an unchanged key) falls straight through — no wasted retry.
- **Runs router (`server/runs.ts`) takes the provider** — every fetcher authenticates with the held key rather than an independent boot-time copy. The live-status path gets the provider itself (refresh-capable).
- **Wired at boot (`server/index.ts`)** — the provider is built once from `identity.apiKey` and passed to the router.
- **Backward-compatible** — a plain `string | null` key still works via a no-op static provider, so existing callers/tests are untouched.

### Tests (32 new)
- `api-key-provider.test.ts` (10): adopts a newer key on refresh; keeps the current key on missing/empty/error/env-resolution-failure (no clobber-to-null); refreshes a null seed up to a first key; static provider is a no-op.
- `run-state.test.ts` (+7): refresh + retry on 401 and 403; no retry when refresh yields the same key; static key never refreshes; no refresh on success or non-auth (404/500) errors; 503 with no key; `isAuthRejection`.
- `runs.test.ts` (+2): router-level 401 recovery via provider; upstream authenticates with the held `sk_` key, never the boot token.

## Acceptance
- [x] A dropped/expired token refreshes instead of 401-locking the Studio.
- [x] Studio actions authenticate via `identity.apiKey`.
- [x] Tests cover the refresh + key path.

## Verification
- `pnpm --filter @sapiom/harness build` — green (server + web).
- `pnpm --filter @sapiom/harness test` — 984 passed / 75 files, 0 failures.
- `pnpm --filter @sapiom/harness lint` + prettier — clean.
- Changeset: `@sapiom/harness` **patch**.

Linear: SAP-1770

### DIFF
--- a/.changeset/harness-refresh-on-401.md
+++ b/.changeset/harness-refresh-on-401.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+Recover from an expired or rotated API key instead of getting stuck. When a live-run status request is rejected as unauthorized, the Studio now re-reads your cached credentials and retries once, so signing in again (in the CLI or elsewhere) unblocks the app in place rather than requiring a restart. Studio actions always authenticate with your held API key.
--- a/packages/harness/src/core/api-key-provider.test.ts
+++ b/packages/harness/src/core/api-key-provider.test.ts
@@ -1,0 +1,122 @@
+import { describe, it, expect, vi } from "vitest";
+
+import {
+  createApiKeyProvider,
+  staticApiKeyProvider,
+} from "./api-key-provider.js";
+
+describe("createApiKeyProvider", () => {
+  it("returns the seeded key from getKey() before any refresh", () => {
+    const provider = createApiKeyProvider("sk-boot", {
+      resolveEnvironmentName: () => Promise.resolve("production"),
+      readApiKeyForEnv: () => Promise.resolve("sk-boot"),
+    });
+    expect(provider.getKey()).toBe("sk-boot");
+  });
+
+  it("adopts a newer key from the store on refresh (the re-login recovery)", async () => {
+    const readApiKeyForEnv = vi.fn().mockResolvedValue("sk-fresh");
+    const provider = createApiKeyProvider("sk-stale", {
+      resolveEnvironmentName: () => Promise.resolve("production"),
+      readApiKeyForEnv,
+    });
+
+    const refreshed = await provider.refresh();
+
+    expect(refreshed).toBe("sk-fresh");
+    expect(provider.getKey()).toBe("sk-fresh");
+    // Refresh read the store scoped to the resolved environment name.
+    expect(readApiKeyForEnv).toHaveBeenCalledWith("production");
+  });
+
+  it("resolves the environment name once per refresh and passes it to the read", async () => {
+    const resolveEnvironmentName = vi.fn().mockResolvedValue("staging");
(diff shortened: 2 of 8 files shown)


evtran0209:
**Merge-order note (from review).** This PR structurally collides with #322 (B2) in `server/runs.ts` + `server/index.ts`: D1 wires the key-provider into the spend/transactions fetchers that #322 deletes.

Recommended order: **#322 → #324 → #325.** Once #322 (+#324) land, this branch gets rebased to (a) drop the spend/transactions provider-wiring — keep the provider only on the `run-state` fetcher — and (b) wire the actions router (from #324) with the provider too. Git will not auto-resolve; the resolution is unambiguous.

Auth logic itself reviewed clean: 984 tests pass, exactly-one-retry on 401/403, `refresh()` never clobbers a good key, static-provider backward-compat. No internal leaks; patch changeset accurate.

evtran0209:
Rebased onto the merged integration (`feat/harness-studio-launch`): dropped the duplicate B2 commit, replayed only the D1 auth commit. Branch is now a single commit on top of integration, git-clean (`MERGEABLE` / `CLEAN`), 8 files.

**Heads-up — pre-existing integration breakage blocks the build/CI (not from this PR):** the merged integration currently does not compile. B2 (#322) deleted `packages/harness/src/core/run-spend.ts` (which defined `resolveCoreBaseUrl`), but A1 (#324) still references it:

- `src/server/actions.ts:33` — `import { resolveCoreBaseUrl } from "../core/run-spend.js";` (deleted module)
- `src/server/index.ts:838` — A1's `createActionsRouter({ …, coreBaseUrl: resolveCoreBaseUrl() })`

These two errors exist on integration before this rebase and are the only build failures; they also fail-to-load 6 vitest suites that (transitively) import `actions.ts`. All other tests pass (**953**), including this PR's 32 new auth tests (10 provider + 17 run-state + 5 runs-router).

Per scope I did **not** touch A1's actions-router wiring — repairing that dangling `resolveCoreBaseUrl` import is the tracked follow-up (#33). Once integration compiles again, this PR is green with no changes needed here.

PR #329, opened 2026-07-22
feat(cli): tree/cost render + --follow live mode for agents logs (SAP-1343)

gwitwer:
## Summary

Makes `sapiom agents logs` reflect the same tree + cost + live state as the UI, closing out the CLI leg of the S5 SDK inspection-parity epic. Renders the `ExecutionProjection` as a tree with per-node cost (replacing the old flat lines) and adds a `--follow`/`--watch` mode that streams live via `watchExecution()`.

> **Note on ticket paths:** SAP-1343 references `orchestrations/logs.ts` and `@sapiom/orchestration-core`. The merged work from siblings SAP-1341/1342 actually lives at `packages/cli/src/commands/agents/logs.ts` importing from `@sapiom/agent-core` — this PR implements against the real command surface.

## What changed

- **Tree + cost render** (`render.ts`, pure/snapshot-tested):
  - `agents logs <id>` renders the run as the root, steps as branches, and each step's dispatched child run nested beneath it. Child edges not reached via a step dispatch are still appended (never silently dropped).
  - Per-node cost shows `capturedUsd` (primary) and `authorizedUsd` (ceiling) side by side — **never collapsed** — with a `settling…`/`pending` affordance. A `null` cost renders as nothing (honest absence, never a fabricated `$0`); a present `"0"` is shown as a known zero.
  - `agents logs` (no argument) renders recent executions tree-aware, grouped by `traceRoot` (degrading honestly to standalone rows per the SDK's documented server-side gap).
- **`--follow` / `--watch` live mode** (`logs.ts`): consumes the `watchExecution()` SSE async-iterator, refetches `inspect()` on each event, and re-renders (in place on a TTY, appended otherwise) until terminal. Ctrl-C aborts the stream and returns cleanly; an SSE handshake failure or mid-run drop degrades to the `waitForExecution` poll loop for the rest of the run.
- **`--verbose`** adds step timings and dispatch details.

## Acceptance criteria

- [x] `agents logs <id>` renders the run as a tree with per-node captured/authorized cost.
- [x] `agents logs` (no arg) lists executions tree-aware.
- [x] `agents logs <id> --follow` streams live and re-renders until terminal.
- [x] Clean teardown on Ctrl-C / stream end; graceful degradation on SSE drop.
- [x] No internal hostnames/customer data in help text or examples (public repo).

## Testing

- New `render.test.ts` (11 cases): tree shape, nested child runs, orphan child edges, list grouping by `traceRoot`, cost never-collapse, honest absence, and settle/pending affordances.
- `pnpm build && pnpm typecheck && pnpm lint && pnpm test` all green across the workspace (83 CLI tests pass).

Closes SAP-1343.

### DIFF
--- a/.changeset/cli-orch-logs-tree-follow.md
+++ b/.changeset/cli-orch-logs-tree-follow.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/cli": minor
+---
+
+`agents logs`: render an execution as a tree with per-node cost (captured primary, authorized ceiling — never collapsed) and add a `--follow`/`--watch` live mode that streams updates until the run reaches a terminal status. `agents logs` (no argument) now renders recent executions tree-aware, grouped by `traceRoot`. A `--verbose` flag adds step timings and dispatch details.
--- a/packages/cli/src/commands/agents/index.ts
+++ b/packages/cli/src/commands/agents/index.ts
@@ -58,6 +58,9 @@ export function registerAgentsCommands(program: Command): void {
     json(group.command('logs [executionId]').description('Inspect an execution, a build (--build), or recent runs.')),
   )
     .option('--build <buildRunId>', 'inspect a build instead of an execution')
+    .option('-f, --follow', 'stream live updates and re-render until the run reaches a terminal status')
+    .option('--watch', 'alias for --follow')
+    .option('--verbose', 'show step errors, timings, and dispatch details')
     .action(action(runLogs));
 
   withHostFlags(json(group.command('signal <executionId>').description('Resume a paused execution.')))
--- a/packages/cli/src/commands/agents/logs.test.ts
+++ b/packages/cli/src/commands/agents/logs.test.ts
@@ -1,0 +1,272 @@
+/**
+ * Tests for the `--follow` machinery in `logs.ts`: the {@link LiveRenderer}
+ * in-place/append behavior (including wrapped-row counting), the
+ * {@link followExecution} SSE → terminal / SSE-drop → poll-fallback / abort
+ * paths, and the `noEvents` seam that drops `waitForExecution` to its poll loop.
+ *
+ * All networked calls are injected via `FollowOverrides`, so nothing here touches
+ * a real gateway; abort is driven through an injected `AbortSignal` rather than a
+ * real process signal.
+ */
+import { waitForExecution, type GatewayClient, type ExecutionProjection, type SseEvent } from '@sapiom/agent-core';
+
+import { followExecution, LiveRenderer, noEvents } from './logs.js';
+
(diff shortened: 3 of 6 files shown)


gwitwer:
Thanks for the thorough pass. Addressed in `2cb78aa`:

**1. TTY line-wrap corruption (fixed).** `LiveRenderer` now tracks **physical rows**, not logical lines. `rowsFor()` sums `ceil(displayWidth(line) / columns)` (min 1/line) using the terminal width read fresh each render, so a wrapped headline moves the cursor up by the right number of rows. `displayWidth` counts code points (`Array.from`) — a good-enough approximation for the CLI's ASCII + width-1 glyph charset; not full East-Asian-width, which the log content never hits. Locked by a test at `columns: 10` asserting the clear uses `\x1b[2A` for a 13-char line.

**2. Ctrl-C ~10s hang in the fallback (fixed).** `waitForExecution` has no `signal` param, but it does expose injectable `sleep`/`now` seams. I pass an **abort-aware sleep** (wakes immediately on abort) and an **abort-aware clock** (reads past the deadline once aborted), so the current poll window returns within one refetch instead of blocking the full `FOLLOW_POLL_WINDOW_MS`. The fallback loop also re-checks `aborted` at the top of each iteration.

**3. Follow-loop tests (added).** New `logs.test.ts` (+9): `LiveRenderer` append vs in-place clear + wrapped-row counting; `followExecution` SSE→terminal, SSE-drop→poll-fallback, fallback timeout→terminal continuation, and clean abort teardown (SIGINT listener removed, fallback untouched). Added `FollowOverrides` seams (inject `inspect`/`watchExecution`/`waitForExecution`/`renderer`/`signal`) so the paths are driven deterministically with no real gateway — abort via an injected `AbortSignal`, not a real process signal.

**4. `noEvents`→poll contract (verified + test).** Confirmed against `waitForExecution`: an immediately-`done` iterator hits `next.done → break`, then drops to the poll loop — it is **not** interpreted as "run finished". Added a test with a fake client returning running→completed that asserts it polls twice and settles `terminal` (not on the first running read).

**5. Heartbeat refetch (confirmed, no change).** `watchExecution` filters heartbeats: a `{}` frame fails `parseSseEvent`'s `type` narrowing → returns `null` → not yielded. So the SSE loop only refetches on real run events, never per heartbeat.

**Cross-cutting (status-fold drift).** Agreed this is now a third independent status-fold (CLI `statusGlyph` vs harness `toStepStatus`/`renderRunState` vs FE). I kept it a separate copy here to stay scoped to SAP-1343, but you're right it's exactly what E1's golden fixtures should cover — I'll follow up to register the CLI as a conformance consumer in E1/E6 rather than fold it into this PR.

`pnpm build && typecheck && lint && test` green; 92 CLI tests pass.

PR #373, opened 2026-07-23
feat(examples): add Natural-Language DB Query Endpoint template (SAP-1881)

gwitwer:
## What

Adds the **Natural-Language DB Query Endpoint** onboarding template (SAP-1881, part of the SAP-1823 gallery epic). Deploys a live endpoint that turns a plain-English question into a **read-only** SQL query and returns the answer.

One run stands up the endpoint; the endpoint answers questions.

```
validate ─▶ resolve ─▶ plan ─▶ guard ─┬─▶ deploy ─┬─▶ deployed      (terminal)
           (database   (models  (read- │           └─▶ deploy_failed (terminal)
            .get)       .run)   only    └─▶ rejected                 (terminal)
                                check)
```

The workflow is the **deployer**: it reads the target connection string from a Sapiom database handle (`database.get`), previews the translation on a sample question so a bad path fails before anything ships, then writes a small server into a sandbox and exposes it at a stable URL (`sandboxes.deployPreview`). The server's own `SAPIOM_API_KEY` is read from the vault at deploy time (`vault.get`), never baked into source. The deployed server re-runs translate → guard → execute per request.

### The endpoint
- `POST /query { "question": "…" }` → `{ question, sql, columns, rows, rowCount, truncated }`
- `GET /health` → `{ "ok": true }`

### The read-only guardrail (defense-in-depth)
1. The LLM is **told** to emit a single `SELECT`.
2. The SQL is **checked** — single statement, starts with `SELECT`/`WITH`, no `INSERT`/`UPDATE`/`DELETE`/DDL keywords.
3. The endpoint **executes** it inside `BEGIN TRANSACTION READ ONLY` (Postgres-enforced) with a statement timeout and a row cap — so a write can't slip through even if the first two layers are wrong.

## Capabilities
`database.get`, `models.run`, `sandboxes.deployPreview`, `vault.get` — each declared id is called by a step (matches the `preview-keep-alive` pattern).

## Files
- `examples/nl-db-query-endpoint/` — `index.ts` (workflow + embedded endpoint server), `package.json` (pinned `@sapiom/agent ^0.6.0` / `@sapiom/tools ^0.20.0`), `tsconfig.json`, `README.md`, `AGENTS.md`, `template.json`.
- `examples/registry.json` — gallery entry + 8-step definition graph.

## Verification
- `npm run typecheck` ✅ — confirms every `ctx.sapiom.*` call exists.
- `prettier --check` ✅
- Offline graph validation (`buildManifest` + `assertValidGraph`) ✅ — reproduces MCP `check`.
- `node --check` on the embedded server (evaluated at true runtime string) ✅.

> `run_local` note: on defaults the `models.run` stub returns non-SQL, so `guard` routes to `rejected` — a legible demo of the guardrail. Pass a stub override returning a real SELECT + `{ "dryRun": true }` to trace the deploy branch offline (documented in `AGENTS.md`). Capability cost/output verify on the deployed path.

Refs: SAP-1881

### DIFF
--- a/examples/nl-db-query-endpoint/.gitignore
+++ b/examples/nl-db-query-endpoint/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+*.log
--- a/examples/nl-db-query-endpoint/AGENTS.md
+++ b/examples/nl-db-query-endpoint/AGENTS.md
@@ -1,0 +1,49 @@
+# Working in this agent
+
+This project defines exactly one Sapiom agent in `index.ts` — **Natural-Language DB
+Query Endpoint**, authored against `@sapiom/agent`. It provisions a live HTTP
+endpoint: `validate` → `resolve` (`database.get`) → `plan` (`models.run`) → `guard`
+(read-only check) → `deploy` (`vault.get` + `sandboxes.deployPreview`) → terminal
+`deployed` / `deploy_failed` / `rejected`. Inside a step's `run`, Sapiom
+capabilities are pre-auth'd on `ctx.sapiom`.
+
+The deployed endpoint's code is `SERVER_SOURCE` — a self-contained ESM server that
+re-runs translate → guard → execute per request. It reads all config from env
+(injected at `deploy` time), so it has no build-time interpolation.
+
+## Authoring
+
+- An agent is `defineAgent({ entry, steps })`; each step is `defineStep({ name, next, run })`. Keep exactly one `defineAgent(...)` export.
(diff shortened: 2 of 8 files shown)


gwitwer:
Superseded by #408, which consolidates all remaining example-template PRs into a single PR (one review, one merge) to avoid serial `examples/registry.json` conflicts. This template's directory and registry entry are included there verbatim. Closing in favor of #408.

PR #374, opened 2026-07-23
feat(examples): add Durable Backfill / Long-Job Runner workflow template (SAP-1886)

gwitwer:
## What & why

Adds the **Durable Backfill / Long-Job Runner** onboarding gallery template (SAP-1886, epic SAP-1823) — a pure durability showcase. It processes a huge dataset in resumable chunks, checkpointing progress so the job survives restarts and picks up where it left off. This is the direct answer to "a long job is too expensive / too fragile to run as an agent."

Closes SAP-1886.

## Shape

```
plan  ──▶  process  ──(checkpoint, then pause: wait for "backfill.heartbeat", $0 idle)──╮
           ▲                                                                            │
           ╰───────────────────────── resume on the heartbeat ───────────────────────────╯
           │
           ╰─(no work left)─▶  finalize  (terminal)
```

- **plan** — resolve the job (dataset size, chunk size, stable `jobId`); resume from a saved checkpoint if one exists for that id.
- **process** — a self-loop. Do one bounded chunk's work in a short-lived sandbox (fed the dataset's `DATABASE_URL` from `database.get` + the chunk range as env), save a per-chunk result file, rewrite the durable checkpoint (`fileStorage`, rotated each chunk), advance the cursor, then `pauseUntilSignal` on `backfill.heartbeat` — or `goto finalize` on the last chunk.
- **finalize** — write a run manifest and terminate with a summary.

Restart survival is two-layered: the cursor rides in `ctx.shared` across every pause within a run, and the checkpoint lives in file storage keyed by `jobId`, so a fresh run started with the same id resumes mid-dataset. The heartbeat is a cron/schedule firing the resume signal on a cadence — nothing runs and nothing is billed between chunks.

## Capabilities

Exercises the five the ticket calls for: **cron** (heartbeat), **pauseUntilSignal** (resumable chunking), **database** (`database.get` → dataset connection string), **sandbox** (`sandboxes.create` / `exec` per-chunk processing), and **file storage** (checkpoint + per-chunk results + manifest).

## Offline / run_local

With no `dbHandle` (or `dryRun: true`), `process` counts each chunk in-process and skips the sandbox / database / file-storage calls, so `run_local` traces the whole `plan → process → … → finalize` loop for free (the local runner auto-resumes each heartbeat pause). Real capability cost is verified only on the deployed path, per the template test loop.

## Verification

- `npm run typecheck` — passes (validated against published `@sapiom/agent` 0.6.6 / `@sapiom/tools` 0.20.1; confirms every `ctx.sapiom.*` call).
- `prettier --check` — clean on all new files + `registry.json`.
- Ran the SDK's `buildManifest` + `validateGraph` on the agent: **0 errors, 0 warnings** — the self-loop heartbeat edge and terminal reachability check out.
- Copy follows `examples/AUTHORING.md`; `registry.json` capability ids match the source (`models.run` is not used here).

## Files

`examples/durable-backfill/` (index.ts, package.json, tsconfig.json, template.json, README.md, AGENTS.md, .gitignore) + one entry in `examples/registry.json`. No `package-lock.json` (matches sibling templates).

### DIFF
--- a/examples/durable-backfill/.gitignore
+++ b/examples/durable-backfill/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+*.log
--- a/examples/durable-backfill/AGENTS.md
+++ b/examples/durable-backfill/AGENTS.md
@@ -1,0 +1,51 @@
+# Working in this agent
+
+This project defines exactly one Sapiom agent in `index.ts` — **Durable Backfill / Long-Job Runner** — authored against `@sapiom/agent`. It processes a large dataset in resumable chunks: `plan` resolves the job and loads any prior checkpoint, `process` handles one chunk (in a sandbox, against a dataset `DATABASE_URL`) and rewrites a durable checkpoint before pausing until a heartbeat, and `finalize` writes a manifest and terminates. The run survives restarts because progress is checkpointed to file storage under a stable `jobId`.
+
+## The chunk loop
+
+- **`process`** is a self-loop. After doing a chunk's work and checkpointing, it returns `pauseUntilSignal({ signal: "backfill.heartbeat", resumeStep: "process", correlationId: ctx.executionId })` when work remains, or `goto("finalize", {})` on the last chunk. Its static `pause: { signal, resumeStep: "process" }` annotation is the build-time graph edge that must match the directive; its `next: ["finalize"]` is the exit edge.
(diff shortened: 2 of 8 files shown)


gwitwer:
Superseded by #408, which consolidates all remaining example-template PRs into a single PR (one review, one merge) to avoid serial `examples/registry.json` conflicts. This template's directory and registry entry are included there verbatim. Closing in favor of #408.

PR #406, opened 2026-07-23
test(harness): stabilize flaky rich-step-detail / step-macros e2e (deterministic waits)

evtran0209:
## Summary

- Replace the flaky `canvas-zoom-reset` / `not.toHaveText("100%")` timing proxy in `pickNode` (both spec files) with `canvas-loading` `not.toBeVisible({ timeout: 15000 })`. The skeleton overlay has `pointer-events: none` once it fades to `opacity: 0` — clicks already pass through; Playwright's visibility check treats `opacity: 0` as not-visible, giving a clean deterministic gate with no fixed sleep.
- Fix `CanvasPane.fitView` to bail out (`return`) instead of falling back to `zoom=1` when `computeFit()` returns `null`. Switching to the Steps tab applies `.is-offstage` (`display: none`) on `canvas-visual`, which causes `ResizeObserver` to fire with `rect.width=0`; the old `?? { zoom: 1, x: 0, y: 0 }` fallback reset zoom to 100%, making the `not.toHaveText("100%")` assert a race under CI parallelism.

## Test plan

- [x] Reproduced original flakiness: `CI=1 E2E_PORT=5590 pnpm --filter @sapiom/harness test:ui rich-step-detail --repeat-each=10` → failures within seconds under 3-way parallel load
- [x] Proved stability: 3× parallel `--repeat-each=15` runs → 150/150, 150/150, 195/195 (all pass)
- [x] Full e2e suite: 248/248 passed
- [x] Unit tests: 1247/1247 passed (90 test files, baseline unchanged)
- [x] `git rev-parse HEAD == git rev-parse origin/evtran0209/stabilize-step-detail-e2e` confirmed


gwitwer:
Closing: stale. The branch carries 40 duplicated commits from an old base and conflicts in ~110 files; the fitView half was reworked on main in #632. If the rich-step-detail e2e flake still reproduces on main, the two-spec wait change is worth redoing as a fresh 2-file PR.

PR #413, opened 2026-07-24
feat(harness): make the workflow canvas fully deterministic (no AI, no user token)

BastienBoymond:
## Why

The workflow canvas reached users via a deterministic base render **plus two AI touchpoints** that ran on the **user's own Claude token** and produced **non-reproducible** output:

1. **Freeform `index.html`** — the agent authored the entire canvas HTML.
2. **Diagram enrichment** — a headless `claude -p` task returned JSON annotations.

Both are removed. Everything is now derived **in-process from the workflow manifest**, so the canvas is **free, instant, and byte-stable** across renders — no token cost, no per-view latency, no staleness.

## What changed

**Determinism**
- New `canvas-derive.ts`: `deriveEnrichment(graph)` builds the summary, notes (from `graph.warnings`) and cross-workflow line from the extracted graph. Enrichment cache, staleness chip and the zod schema are gone.
- Agents no longer write canvas markup; the diagram is generated from the bound workflow. Deleted `canvas-enrich.ts`, `profiles/canvas-guidelines.ts` and their tests.
- `canvas-watcher` auto re-renders on workflow source edits — no manual button.

**Rendering / UX**
- Fix pan/zoom + node clicks that were dead in dev: esbuild/tsx `keepNames` rewrote nested fns to `__name(...)`, whose helper `toString()` didn't capture → `ReferenceError` inside the sandboxed iframe. Added `NAME_SHIM`.
- Added the pan/zoom view receiver (`bootCanvasView`) + center-origin transform.
- Adaptive fit-to-view: a `ResizeObserver` posts the full card size and the parent retries fit via `rAF`, so it never snaps to 100% on a fresh/hard load.
- Barycenter layout + collision-gated edge routing so arrows no longer pass under node boxes; symmetric merges converge cleanly.
- Restored the node/edge **legend** (footer inside the card, hairline separator). The floating zoom/fit controls sit in a reserved strip **below** the iframe, so nothing overlaps.
- De-AI'd all empty-state copy/icons (`Sparkles` → `Workflow`; "Visualize … as the agent works" → "generated automatically from the bound workflow").

**Steps tab**
- The board drew the diagram but never **posted** the step graph, so the Steps tab read "No steps yet" even with a visible diagram. Now the render embeds the graph as an inert `<script id="sapiom-graph">` block and posts it (`sapiom-canvas:graph`), mirroring `renderGraphSvg`'s merge exactly (`role`←sublabel, `description`, edge key `from->to`).

## Testing
- **1220** unit tests pass; ESLint clean; typecheck + build green.
- New golden tests: `canvas-derive`, and the embedded step-graph payload (parseable, real nodes/edges).
- Touched web e2e pass: empty-state, steps tab, welcome, CTA gating.
- Verified end-to-end (parent + sandboxed iframe): the parent receives the posted graph (all 8 order-triage nodes + 8 edges) and a valid fit size.

## Note for reviewers
This touches the backend-rendered iframe document, so a `pnpm --filter @sapiom/harness dev` **restart + hard-refresh** is needed to see it locally.

🤖 Generated with [Claude [](Code](https://claude.com/claude-code))


gwitwer:
## Review

Read the full diff (33 files, +1098/−1744). This is a genuinely good architectural simplification — deleting a whole token-consuming, cache-backed, validation-heavy subsystem is the right call, and the security-sensitive bits are handled correctly.

### What works well

- `deriveEnrichment` family is pure, total, and well-tested (purity test + real-fixture end-to-end).
- **Security is correct:** `buildGraphPayload` escapes `<` → `<` so a label containing `</script>` can't break out of the JSON data block, and it round-trips through `JSON.parse` (test confirms). HTML-facing strings go through `esc()`.
- The `NAME_SHIM` fix is well-reasoned and correctly ordered (shim precedes any `__name(...)` call; test asserts it). The no-op `function __name(fn){return fn;}` is sufficient since nested fns are called by reference, not by `.name`.
- The write-skip guard in `renderWorkflowRenderFile` is the right move now that any source edit triggers a render — stops a no-op save from bumping mtime and firing a needless iframe reload.
- Loop-freedom reasoning is sound: render writes land under `.sapiom/` (a skipped dir) so they're *canvas* changes not *source* changes, and CSS transforms don't change layout size so the `ResizeObserver` can't self-trigger from the fitted view.

### Issues & risks (nothing blocking)

- **Re-extraction on *any* `.ts` save in cwd (perf, medium).** `onSourceChange` re-renders the bound workflow on any `.ts`/`.tsx` change anywhere in the project, and since the extraction-cache fingerprint keys on source mtimes, every such save is a cache-miss → full re-extraction (not just the cheap render). Debounced 150ms and write-skip prevents the reload, but on a large project with rapid auto-save this spawns more extraction work than the old model. Intended "live" behavior, so fine to ship — just flagging that extraction, not just rendering, runs on the hot path.

- **`layoutGraph` with an empty node set (defensive, low).** `Math.min(...centersAll)` / `Math.max(...centersAll)` over an empty array yield `±Infinity`, making `width` `NaN`. The old code initialized `maxCols` to `1` and was safe for zero nodes. Unreachable in practice (a successful extraction always has an entry node; error panels don't call `layoutGraph`), but the previous code degraded gracefully and this doesn't. A one-line guard would restore that.

- **`MAX_SOURCE_FILES = 400` truncation on the polling fallback (low).** `snapshotWorkflowSources` walks in unsorted `readdir` order and stops at 400 files, so on platforms without recursive `fs.watch` (notably Linux), edits to `.ts` files past the 400th wouldn't be detected. Only affects the polling path; documented as a defensive ceiling — but the cap is silent.

- **Non-English comments (nit).** `"l'un sur l'autre"` appears in `canvas-body.ts` and `styles.css` — swap for English to match the rest of the codebase.

- **Copy inconsistency (nit).** The empty-state CTA button is now "Render diagram" (`Workflow` icon), but `WelcomePanel`'s hint chip still reads "Visualize".

- **Dead branch / dead CSS (nit).** `deriveNotes`'s trailing `notes.length > 0 ? notes : undefined` is unreachable (empty warnings already returned `undefined`; a slice of non-empty is non-empty). And `.canvas-stats` CSS remains after the stats header was removed.

- **`postMessage(..., "*")` wildcard origin (informational).** `bootCanvasView`/`bootCanvasGraph` post to `"*"`. Consistent with the existing node-click channel and low-risk for a local sandboxed dev iframe with no secrets — noting for completeness, not asking for a change.

### Verdict

Solid, well-tested, clear net simplification with a real cost/latency/reproducibility win. The empty-node guard and the two comment/copy nits are the only things I'd optionally touch; everything else is fine as-is.

evtran0209 (commented):
lgtm, i did a pass and there is no regression on the live run state.

PR #414, opened 2026-07-25
feat(harness-desktop): Electron desktop app + desktop-first dev

gwitwer:
## Summary

Brings @ratataque and @baptiste's Electron desktop wrapper (`@sapiom/harness-desktop`) onto `main`, and adopts desktop-first development for the harness. The desktop app is the experience we intend to push customers to; the web app stays as the open-source/hackable surface.

Merges `ewan/feat/harness-desktop` (48 commits) with three conflicts resolved toward `main`, plus four integration fixes described below. Two of those fix things that were **red on the source branch**.

### The desktop app (from the source branch)

- New private `packages/harness-desktop`: Electron main/preload/renderer, boot flow, onboarding window, auto-install of the coding agent on first run, runtime shims, node-pty under Electron's ABI.
- Multi-OS packaging via `electron-builder` (dmg/zip, AppImage/deb, nsis), release workflow on `harness-desktop-v*` tags, signing-ready pipeline.
- **`--smoke`: unattended verification of the PACKAGED build** — six checks each mapping to a real shipped bug (asar path resolution, native-module ABI, unpacked deps, preload bridge). This is the most valuable thing in the branch: unit and e2e tests structurally cannot see packaging bugs.
- `packages/harness/CLAUDE.md` documenting desktop-first development.
- Harness-side Electron support: `app.asar.unpacked` path translation, `ELECTRON_RUN_AS_NODE` for the Canvas subprocess, bounded server shutdown (an unterminated `/ws` client hung `before-quit` forever and left a zombie holding the single-instance lock), and an embedding surface so the desktop host reuses the CLI's server/setup flow instead of forking it.

### Conflict resolutions

- `agent-core/src/scaffold.ts` — `VERSION_FALLBACK` to main's `agent 0.6.7 / tools 0.22.0`. The branch's `f9c45cb` synced it to `0.6.5/0.20.1`, which main has since released past; `scaffold.test.ts` asserts it matches the workspace.
- `harness/web/src/styles.css` — main's version verbatim (see below).
- `pnpm-lock.yaml` — regenerated via `pnpm install`, not hand-merged.

### Fix: route the onboarding window through the design-system seam

The branch extracted `theme-tokens.css` + `primitives.css` out of `styles.css` so the setup window could share SPA theming. Main has since migrated `styles.css` onto the `@sapiom/design-system` seam (private package when installed, committed `ds-neutral` fallback otherwise — this repo is public), under an explicit *"never redefine a token"* rule. The extracted files are a snapshot of the **pre-migration** values (`--radius: 6px`, hardcoded Inter), so merging them reintroduces the second palette that migration exists to delete. They also emitted a real `[css-syntax-error] Expected "}" to go with "{"` in the harness build — visible in the source branch's own CI log — from an unbalanced brace left by the split.

Both files are deleted and `main.tsx` restored. `copy-renderer.mjs` now runs the **same probe** `vite.config.ts` uses and copies the resolved tokens in as `ds-tokens.css` / `ds-agent-cloud.css`. The setup window has no bundler (it links plain stylesheets), so a build-time copy is the equivalent of Vite's resolve alias. `setup.css` addresses system tokens directly since it can't load the SPA's bridge — including `--brand` rather than `--accent`, which the design system defines as a hover *wash*.

### Fix: the macOS release build (regression from `bc6333c`)

The mac lane went red at `9144e79` (green at `3718bc3`). `bc6333c` replaced `CSC_IDENTITY_AUTO_DISCOVERY: "false"` with an unconditional `CSC_LINK: ${{ secrets.CSC_LINK }}`. **An unset secret still exports as the empty string, and to electron-builder an empty `CSC_LINK` is a *defined* one** — it takes the "certificate file provided" path even with auto-discovery off, resolves `""` to the cwd, and fails the build with:

```
• empty password will be used for code signing  reason=CSC_KEY_PASSWORD is not defined
⨯ /private/var/folders/.../sapiom-harness-desktop-pack not a file
```

Reproduced locally in both directions: with `CSC_LINK=""` the build dies at that exact error; with it unset the same build completes and smokes 6/6. Now unset when empty — the same empty-string-is-not-absent trap `pack.mjs` already filters for its passthrough args.

### Fix: the local packaging loop

`pack.mjs` defaulted to `--linux`, so on macOS — the primary shipping target and the only OS we develop on — the documented loop (`pnpm dist` then `scripts/smoke.sh`) could not work: `dist` errored, then smoke looked for a `release/mac-arm64` app nothing had built. Defaults to the host platform now. CI passes its flag explicitly (`test.yml:165`, `matrix.packflag`), so CI behavior is unchanged.

### Fix: cover the seam in the smoke test

Packaging now depends on a build-time CSS copy, so `preload-bridge` additionally asserts the tokens **resolve** in the setup window (`--bg` and `--brand` non-empty) rather than that files exist — a 404'd `<link>` and a stylesheet that defined nothing fail identically there, and every other check would still pass.

### Fix: drop the macro-actions e2e (it drives a component main deleted)

`web/e2e/macro-actions.spec.ts` accounted for all 7 initial `playwright-mock` failures, every one `element(s) not found`. It drives `WorkflowActionStrip.tsx` — the per-row hover strip carrying `data-testid="workflow-bind"` and `` data-testid={`macro-${id}`} `` — which main **deleted** in `b6b9d16` (Harness Studio: new run / deploy / inspect experience). The source branch's merge-base predates that, so the spec merged cleanly into a tree where its target no longer exists.

Deleted rather than re-pointed, because the coverage already exists against the current UI and is stricter:

- **dispatch routes** → `direct-actions.spec.ts` asserts the deploy / run / run-local direct routes fire, records `lastDirectAction`, **and that the pty is never written** — the architectural invariant this spec didn't check.
- **session binding** → `smoke.spec.ts:219-423` and `snippet-panel.spec.ts` cover binding on session start, per-session bindings, and that focusing a workflow doesn't rebind.

The one case with no equivalent is the stale-closure guard ("targets the workflow selected NOW"), and it isn't transferable: it describes the strip re-anchoring to a new row while its handler closes over the old workflow. The Studio action bar isn't per-row and has no re-anchor, so there's no closure to go stale.

Note the canvas suite was **never reached** by the original CI run (the job failed at the earlier step and skipped it), so Ewan's 5 new `canvas-render.spec.ts` cases were unverified. Run locally: 10/10 pass.

### Fix: scope the release workflow's token

CodeQL flagged `desktop-release.yml` as having no `permissions` block. Declared `contents: read` at the **top level** rather than on the `build` job, matching `test.yml`'s existing "Minimal token scope" convention — a top-level default also covers any job added later. The `release` job keeps its own `contents: write` override, the only write-scoped token in the workflow.

Deliberately omits the `actions: read` from the suggested changeset: nothing in `build` reads the Actions API (`upload-artifact` authenticates with `ACTIONS_RUNTIME_TOKEN`, not `GITHUB_TOKEN`), so it would be an unused grant — the very thing the rule exists to prevent.

### Cleanup

Removes the three self-labeled TEMP CI triggers: the feature-branch `push` trigger in `test.yml` and the `branches:` block in `desktop-release.yml`.

## Motivation

Desktop is the primary supported experience; the web app exists to be open-sourced. Building desktop-first stops us making choices that don't survive packaging, and leaves less to adapt when the web refactor lands.

## Testing

- [x] `pnpm build` — exit 0
- [x] `pnpm typecheck` — exit 0, no errors
- [x] `pnpm test` (all packages) — exit 0
- [x] `pnpm lint` — exit 0, no errors (only pre-existing warnings in untouched packages)
- [x] `harness-desktop` units — 13/13, 2 files
- [x] `test:ui` — 245 passed
- [x] `test:canvas` — 10 passed (incl. the 5 new cases CI's first run never reached)
- [x] **Packaged macOS app: 6/6 smoke checks pass**, including `tokens resolve (--bg #fbfbfc)` from inside `app.asar`
- [x] macOS signing failure reproduced and verified fixed

## Reviewer notes — three deliberate non-changes

1. **`installDependencies` defaults to `true`** in `example-seed.ts`, so first-run seeding blocks on `npm install` (120s timeout) — and this applies to the **published CLI**, not just the desktop app. Well-motivated (Canvas can't bundle a project with no `node_modules`) but it changes `npx @sapiom/harness` first-run behavior. @ratataque — intended?
2. **`pnpm-workspace.yaml` adds `electron` + `esbuild` to `onlyBuiltDependencies`**, so every clone of this public repo now downloads the Electron binary on install. Kept as authored; flagging the cost.
3. **No `.nvmrc`.** `desktop-release.yml` already warns that a newer dev-box Node breaks Electron's binary extraction; on Node 25.8.0 `pnpm dist` exits 1 in electron-builder's post-build cleanup *after* writing complete artifacts. Pinning Node 22 would save someone an afternoon.

Signing itself remains blocked on an Apple Developer ID cert in GitHub secrets — until then every build ships with the misleading "Sapiom is damaged" Gatekeeper error. That's procurement, not code.

## Checklist

- [x] All tests passing
- [x] Documentation updated (`packages/harness/CLAUDE.md` seam section rewritten)
- [x] No hardcoded secrets
- [x] Self-reviewed

https://claude.ai/code/session_01E1QgeYp3aeZ2fuRMU9SqdK


baptiste:
I approve, no need to keep me on the team

positiveblue (approved):

PR #416, opened 2026-07-25
fix(harness): harden run-local child lifecycle — kill on disconnect + timeout + stdin guard (SAP-1983)

evtran0209:
Hardens the run-local child-process lifecycle. `POST /api/runs/local` spawns a child that bundles + runs **untrusted workflow code**; it had three defects:

- **No kill on disconnect** — navigating away from a hung/looping run left the child running; repeated clicks orphaned `node` processes. Now `res.on("close")` → SIGTERM, then SIGKILL after a 3s grace, guarded so it never fires on normal completion.
- **No wall-clock timeout** — a looping run was never killed. Now killed after `RUN_LOCAL_MAX_DURATION_MS` (5 min) with a terminal "timed out" line written to the stream.
- **Unguarded `child.stdin.end()`** — an EPIPE (child dies before draining stdin) became an uncaughtException that could crash the long-lived harness server. Now `child.stdin.on("error", …)` swallows it.

Uses `res.on("close")` (not `req.on("close")`, which fires right after body-parse) so a completed response never triggers a false-positive kill. All timers/listeners are cleared on settle — no leaks, no double-settle.

Fixes SAP-1983. Closes SAP-1989 (duplicate).

> ⚠️ **Targets the _next_ release (0.1.8).** Merge **after** 0.1.7 (#405) publishes — kept as a **draft** so it can't batch into 0.1.7.

**Verification:** 36 actions tests pass; typecheck + build clean.

### DIFF
--- a/.changeset/harden-run-local-lifecycle.md
+++ b/.changeset/harden-run-local-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+Hardened local runs: the run-local process is now stopped when you navigate away or a run exceeds its time limit, and a broken input pipe can no longer crash the Studio server.
--- a/packages/harness/src/server/actions.test.ts
+++ b/packages/harness/src/server/actions.test.ts
@@ -102,6 +102,9 @@ class FakeChild extends EventEmitter implements RunLocalChildProcess {
     }
   })(this);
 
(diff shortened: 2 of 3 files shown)


evtran0209:
Superseded by #405 — the SAP-1983/SAP-1989 fix (cherry-picked as commit 6f644a7, changeset included) is folded into #405. Closing to avoid a double-merge.

PR #420, opened 2026-07-27
feat(harness): canvas board redesign + deterministic step/workflow metadata

BastienBoymond:
## What & why

A fidelity pass on the canvas pane (Canvas / Steps / Code) against the new design, **plus** the step/workflow metadata it surfaces — descriptions, inputs, capabilities, overview — is now produced **deterministically** (no LLM, no user token).

## Board & drawer redesign

- **Step navigation removed from the drawer.** The workflow chart sits right beside it, so the in-drawer transition links were redundant. The Steps-tab lineage ("Leads to" / "Reached from") is kept but made **read-only** — there's no chart there to click into.
- **Macros** are a horizontal pill row now; the first label is shortened to **"Why slow / stuck?"**; the drill action reads **"Open step"**.
- **Chat is a standalone, toggleable panel** (`CanvasChatPanel`) — closed by default, opens **independently** of the ⓘ info panel, and both can be open at once.
- **Header matches the design:**
  - workflow **name dropped** (it already lives in the rail),
  - **expand** moved into the tab bar, next to collapse-panel,
  - the board **subheader bar is gone entirely** (deployed *and* non-deployed),
  - the **deployed pill** now lives in the tab bar (right of the tabs) and **doubles as the dashboard link**.
- **"Render diagram" button removed** — the canvas auto-generates from the bound workflow, including **on session start** for an already-bound / resumed session (`server/index.ts`).

## Deterministic step & workflow metadata (Option A — no LLM)

- **Agent SDK**: `defineStep` gains optional `description` + `capabilities`; `defineAgent` gains `description`. They flow through `buildManifest` into `AgentStepManifest` / `AgentManifest` (both optional → back-compat with older manifests).
- **Canvas renderer** posts per-node `description` / `inputSchema` / `capabilities` / `timeoutMs`, plus a new **Overview** payload (`bootCanvasOverview` / `buildOverviewPayload`) with step / exit / entry stats.
- **Capabilities auto-detect** from the workflow source when not authored (`detectStepCapabilities` over `sapiom.*` call patterns), so real `sapiom.models.run`, `email.messages.send`, etc. render as chips with zero authoring.
- **Canvas template** gets the design's dotted-grid background + transparent panel.

## Verification

| Check | Result |
| --- | --- |
| Web e2e (Playwright) | **242 pass** |
| Harness unit | **1220 pass** |
| Agent SDK unit | **77 pass** |
| Web typecheck (`tsc`) | clean |
| Full build (agent + server + web) | clean |

e2e updated for the removed drawer nav, relocated deployed pill, read-only Steps lineage, and removed render CTA.

## Notes

- The authored `description` / `capabilities` on the on-disk sample are **inert** until the sample adopts the new `@sapiom/agent` (it currently pins published `0.6.7`); capabilities and inputs already render deterministically today.
- Local test workspaces (`packages/harness/hello-agent`, `web-research-digest`) are nested git repos and are intentionally **not** part of this PR.

## Follow-up (not in this PR)

- **Branch-condition tags** ("why this branch was taken") — pending a decision between an authored SDK field (e.g. `next: [{ target, when }]`) and AI.

### DIFF
--- a/.changeset/agent-step-description-capabilities.md
+++ b/.changeset/agent-step-description-capabilities.md
@@ -1,0 +1,10 @@
+---
+"@sapiom/agent": minor
+---
+
+Add optional authoring fields to the workflow SDK, surfaced by the harness canvas:
+
+- `defineStep` gains `description` (a one-line summary of what the step does) and `capabilities` (the Sapiom capability ids the step declares it calls).
+- `defineAgent` gains `description` (a one-line summary of the whole workflow).
+
+All three are optional and flow through `buildManifest` into `AgentStepManifest` / `AgentManifest` as optional fields, so manifests emitted before this release keep validating unchanged. No runtime behavior change — the fields are read only by tooling (the canvas step inspector and workflow overview).
--- a/.changeset/canvas-descriptions-and-redesign.md
+++ b/.changeset/canvas-descriptions-and-redesign.md
@@ -1,0 +1,11 @@
+---
+"@sapiom/harness": minor
+---
+
+Canvas board redesign + deterministic step/workflow metadata.
+
+- **Board redesign** matching the new design: in-drawer step navigation removed (the chart is beside it), chat split into a standalone toggleable panel independent of the info panel, the board subheader dropped with the deployed pill and expand relocated to the tab bar, and the manual "Render diagram" button replaced by auto-render on bind/session-start.
+- **Deterministic metadata:** the renderer surfaces per-step `description` / `inputSchema` / `capabilities` / `timeoutMs` and a workflow Overview payload, all read from the manifest (no LLM in the render path). When a step doesn't call a capability or declare a description, the field is simply absent — the shape summary still renders.
+- **Capability auto-detect** from `sapiom.*` call sites in the workflow source, attributed to the `defineStep` block the call sits in (calls in shared helpers are left unattributed rather than mis-billed to the nearest step).
+
+Reads the new optional `description` / `capabilities` fields from `@sapiom/agent`; workflows on an older SDK render with those fields blank.
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -45,6 +45,9 @@ export interface AgentDefinition<
   TShared extends Record<string, unknown> = Record<string, unknown>,
 > {
   readonly name: string;
+  /** Human-authored workflow summary for the canvas Overview panel (Option A —
+   *  deterministic, no LLM). */
+  readonly description?: string;
   readonly entry: string;
   readonly steps: Readonly<Record<string, StepDefinition<TShared>>>;
   /**
--- a/packages/agent/src/build-manifest.ts
+++ b/packages/agent/src/build-manifest.ts
@@ -35,6 +35,8 @@ export function buildManifest(
       timeoutMs: number | null;
       inputSchema: Record<string, unknown> | null;
       transitions: ManifestTransition[];
+      description: string | null;
+      capabilities: readonly string[];
     }
   > = {};
 
@@ -50,12 +52,15 @@ export function buildManifest(
       timeoutMs: step.timeoutMs ?? null,
       inputSchema,
       transitions: transitionsFor(step),
+      description: step.description ?? null,
+      capabilities: step.capabilities ?? [],
     };
   }
 
   return {
     protocol: MANIFEST_PROTOCOL,
     name: def.name,
+    description: def.description ?? null,
     entry: def.entry,
     sdkVersion: opts.sdkVersion,
     artifact: opts.artifact,
--- a/packages/agent/src/manifest.ts
+++ b/packages/agent/src/manifest.ts
@@ -47,6 +47,11 @@ export interface AgentStepManifest {
    * per-kind); the renderer draws one edge per entry. `reachable ⊆ declared`.
    */
   readonly transitions: readonly ManifestTransition[];
+  /** Human-authored step description (Option A). Optional for back-compat with
+   *  manifests emitted before this field existed. */
+  readonly description?: string | null;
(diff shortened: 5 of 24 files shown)


gwitwer:
Reviewed alongside #421 (stacked on this one). CI is green here — 8 checks.

## What this does

Two things in one PR: (a) a design-fidelity pass on the canvas pane — drawer step-nav removed, chat split into a standalone toggleable panel, board subheader deleted with the deployed pill and expand relocated to the tab bar, manual "Render diagram" CTA replaced by auto-render on bind/start; and (b) the real substance — `description` + `capabilities` on `defineStep`/`defineAgent`, flowed through `buildManifest` → `AgentManifest` → canvas graph → step inspector, plus static capability detection from `sapiom.*` call sites.

**The architecture is right.** Keeping the render path deterministic and treating AI purely as an authoring assist that writes to version-controlled source (in #421) is the good answer to "should we fill descriptions with AI?" — the tempting alternative, an LLM in the render path with a cached overlay, would have been much worse.

---

## Blocking-ish

### 1. The headline feature is inert, and there's no changeset

`packages/agent/package.json` is still `0.6.7` — exactly what's on npm today — and there's no `.changeset/` entry, despite this changing a published package's public API and the repo clearly using changesets.

Downstream of that: `example-seed.ts` now authors `description:` / `capabilities:` in the scaffolded `order-triage` sample, but `scaffold()` installs `@sapiom/agent@0.6.7`, whose types don't declare those fields. Two consequences:

- `defineStep`/`defineAgent` take an inline object-literal param type with no index signature, so **TS excess-property checking rejects the extra keys** — red squiggles in the very first example a new user opens.
- At runtime 0.6.7's `defineStep` doesn't copy the fields, so **the seeded demo renders with no descriptions and no authored capabilities** — the exact feature being shipped is invisible in the golden-path sample.

The PR body acknowledges this ("inert until the sample adopts the new `@sapiom/agent`") but doesn't resolve it. Given this is a launch-day deliverable, the sequence is: changeset → publish → bump the scaffold pin. The `examples` CI job won't catch it, because the seed is a template string written at runtime, not a checked-in example.

---

## Worth fixing before merge

### 2. Capability attribution mis-attributes trailing helpers

```ts
const preceding = stepDeclarations.filter((d) => d.index < match.index).at(-1);
```

Any `sapiom.*` call positioned after the *last* `defineStep({ name: ... })` in a file gets attributed to that step — helper functions at the bottom of the file being the common case. Launch detection uses the same shape, but a launch renders as an extra dashed node (additive, visibly separate); a wrong capability chip reads as a factual claim about what a step calls, **and capabilities are what Sapiom bills for**. Bounding attribution to the brace-balanced extent of the `defineStep(` call would close it; failing that, at least drop calls appearing after the last step's closing brace.

### 3. `detectStepCapabilities` doubles the source scan

It re-walks `listSourceFiles(root)` and re-reads every file `detectWorkflowLaunches` just read, moments earlier. It's behind `extractWorkflowGraphCached`'s fingerprint so it's cache-miss-only — but every real code edit is a cache miss, and auto-render now fires on every save. One pass running both regexes is a small refactor and halves the I/O on the hot path.

### 4. The `i` flag defeats the lowercase intent

```ts
/(?<![\w$])sapiom\s*\.\s*([a-z][\w$]*(?:\s*\.\s*[a-z][\w$]*)+)\s*\(/gi
```

The comment describes lowercase namespaces, but `i` makes `[a-z]` match uppercase — `sapiom.Foo.Bar(` becomes a capability chip. Drop `i`, keep `g`. (The whitespace-stripping before the `NON_CAPABILITY_CALLS` check is handled correctly, nice.)

---

## Smaller notes

- **Expand lost its guard.** `canExpand={showingFrame}` is gone; the tab-bar expand renders whenever `rightTab === "canvas"`. You can now full-screen an empty state — which is the state a first-run user sees.
- **`WorkflowActionsHeader` runs for the board and returns `null`.** `CanvasPane` still mounts it unconditionally for a bound workflow, so on the board it runs `useState`/`useRef`/`useCallback`, builds a menu, and renders nothing. Skip the mount when `surface === "board"` instead.
- **~100 lines of e2e deleted, not re-pointed.** Fair for the two visualize tests. But "direct/inject split is exclusive" was asserting the *routing contract* (a macro run must not clobber `lastDirectAction`), not the visualize button — `retry` still goes through `onRunMacro`, so that one could be re-pointed rather than dropped. `direct-actions.spec.ts` also now ends on a dangling comment block where the describe used to be.
- **`visualize` is half-removed.** `findVisualizeMacro` is now referenced only by its own unit test, and `visualize.triggered` is still a live tracked event in `rest.ts` + `types.ts`. Either keep it deliberately or finish the removal.
- **Stats strip bolds the entry step name.** `parts.join(" · ")` → the panel splits on `·` and bolds the first token, so `intake entry` renders as **intake** entry, bolded like a number among "**5** steps · **2** exits". Cosmetic; a step label containing a space renders worse.

---

## Done well

- **XSS is handled on both sides.** Descriptions are now author-controlled strings embedded in `<script type="application/json">`; both `buildGraphPayload` and `buildOverviewPayload` escape `<`, and the web side re-validates every field (`typeof === "string"`, `Array.isArray` + filter, `parseInputSchema`). Right belt-and-braces for a channel that just started carrying user content.
- **Manifest back-compat.** New fields optional on `AgentStepManifest`/`AgentManifest`, `runManifestCheck` is a bare `JSON.parse` with no strict schema, web parser tolerates absence. Old manifests keep working without a protocol bump.
- **On-start render for resumed sessions** (`server/index.ts`) is the correct companion to deleting the manual CTA, and it's mutually exclusive with the rescan's auto-bind render (the `boundWorkflowPath` read is synchronous, so no race), with a `.catch` so it can't reject unhandled.

---

**Recommendation:** solid merge once the changeset lands and the publish sequence is settled.

gwitwer (approved):

PR #421, opened 2026-07-27
feat(harness): "Describe with AI" — agent authors the canvas description fields

BastienBoymond:
> Stacked on #420 (base = `compare-drawer-step-nav`). Merge #420 first; this diff shows only the AI-description feature.

## What & why

#420 made the canvas read `description` off `defineAgent` / `defineStep` deterministically (Option A — no LLM in the render path). The gap: **someone still has to write those fields.** This adds the one AI touchpoint that genuinely needs a reading of the code — **the agent authors the `description` fields** — while everything downstream stays deterministic.

The result is durable: descriptions land in the **source**, editable and version-controlled, and the canvas re-renders from them. No overlay, no cache, no reversal of the "deterministic canvas" contract.

## How it works

- A new **"✨ Describe with AI"** action in the canvas Overview panel injects an engineered prompt into the bound session: *add a one-line `description` to the workflow and every step, inferred from the code; don't touch logic; the canvas re-renders on save.*
- The agent edits the workflow source. The **existing** `onSourceChange → autoRenderCanvas` watcher (`server/index.ts`) re-extracts and re-renders — so **no new server code**, and no token beyond the session that's already running.
- `runManifestCheck` esbuild-imports the workflow `index.ts` directly, so the edited fields are picked up on the very next render.
- The label adapts: **"Describe with AI"** when empty, **"Rewrite descriptions with AI"** when a description already exists.
- The button is **hidden when no session can take the inject** (gated on `boundWorkflow` + `sessionId`).

## Design notes

- **Trigger is an explicit button, not auto-on-render.** It edits your source and spends agent tokens, so silently mutating code on every render would be too aggressive. (Easy to flip to automatic-on-missing later if you'd prefer.)
- The prompt builder is a pure, unit-tested lib (`web/src/lib/describe-prompt.ts`).
- Requires the workflow to use a `@sapiom/agent` with the `description` field (added in #420); the prompt tells the agent to upgrade if the installed version is too old.

## Files

| File | Change |
| --- | --- |
| `web/src/lib/describe-prompt.ts` | **new** — pure prompt builder |
| `web/src/lib/describe-prompt.test.ts` | **new** — 4 unit tests |
| `web/e2e/canvas-describe-ai.spec.ts` | **new** — 2 e2e (affordance shows + inject payload) |
| `web/src/components/CanvasOverviewPanel.tsx` | `onDescribeWithAI` prop + button |
| `web/src/components/CanvasPane.tsx` | thread the action (gated on `boundWorkflow` + `sessionId`) |
| `web/src/styles.css` | `.canvas-describe-ai` accent pill |

## Verification

| Check | Result |
| --- | --- |
| Web e2e | **244 pass** (2 new) |
| `describe-prompt` unit | **4 pass** |
| Web typecheck (`tsc`) | clean |

### DIFF
--- a/.changeset/manifest-schema-description.md
+++ b/.changeset/manifest-schema-description.md
@@ -1,0 +1,9 @@
+---
+"@sapiom/agent": patch
+---
+
+Fix: authored `description` (on `defineStep` and `defineAgent`) and step `capabilities` were silently dropped during manifest validation.
+
+`agentManifestSchema` is a `z.object()`, which strips keys it doesn't declare. The previous release added `description`/`capabilities` to the TS types and to `buildManifest`'s output, but not to the runtime schema — so `agentManifestSchema.parse()` (which `@sapiom/agent-core`'s `check()` runs on every manifest) discarded them. The result: tooling that reads the validated manifest — the harness canvas — always saw empty descriptions, no matter what the source authored.
+
+Adds `description` (step + workflow) and `capabilities` (step) to the schema as optional fields, so they survive validation and reach consumers. No behavior change for manifests that don't use them.
--- a/packages/agent/src/manifest.spec.ts
+++ b/packages/agent/src/manifest.spec.ts
@@ -359,6 +359,43 @@ describe("buildManifest", () => {
     expect(parsed.steps.entry.inputSchema).not.toBeNull();
     expect(parsed.steps.exit.timeoutMs).toBe(5000);
   });
+
+  it("preserves authored description + capabilities through schema validation (regression: they were stripped)", () => {
+    const def = defineAgent({
+      name: "described-wf",
+      description: "Triages incoming tickets end to end.",
+      entry: "classify",
+      steps: {
+        classify: defineStep({
+          name: "classify",
+          next: [],
+          terminal: true,
+          description: "Tags the ticket with a category for routing.",
+          capabilities: ["rules.classify"],
+          async run() {
+            return terminate(null);
+          },
+        }),
+      },
+    });
+    const manifest = buildManifest(def, {
+      sdkVersion: DUMMY_SDK_VERSION,
+      artifact: DUMMY_ARTIFACT,
+    });
+    // buildManifest emits the authored fields...
+    expect(manifest.description).toBe("Triages incoming tickets end to end.");
+    expect(manifest.steps.classify.description).toBe("Tags the ticket with a category for routing.");
+    expect(manifest.steps.classify.capabilities).toEqual(["rules.classify"]);
+
+    // ...and the schema must NOT strip them on parse. This is the exact bug the
+    // canvas hit: z.object() drops keys absent from the schema, so an authored
+    // description silently vanished on `agentManifestSchema.parse` and never
+    // reached the renderer — even though the TS type and buildManifest had it.
+    const parsed = agentManifestSchema.parse(manifest);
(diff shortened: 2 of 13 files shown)


gwitwer:
Reviewed together with #420 (the base). Notes on #420 are on that PR; this covers only the stacked diff.

## The approach is right

The one AI touchpoint that genuinely needs a reading of the code is *authoring* the fields, and everything downstream stays deterministic. Output lands in version-controlled source, editable and reviewable, and the existing source watcher re-renders — no overlay, no cache, no new server code. That's the good answer to the question in the launch thread, and worth saying so explicitly.

---

## Blocking

### 1. No CI has ever run on this branch

`gh pr checks 421` → *"no checks reported on the 'canvas-ai-descriptions' branch"*. The workflows appear to only trigger for PRs targeting `main`, so the "244 pass / typecheck clean" numbers are local-only and this diff has never been through a gate. Rebase onto `main` after #420 merges (or re-target) so it runs at least once before merging.

### 2. Non-functional until #420 is *published*, not just merged

The injected prompt says:

> The `description` field needs a recent @sapiom/agent; if the installed version rejects it, **upgrade @sapiom/agent to the latest first**.

Latest on npm *is* `0.6.7`, which is the version that lacks `description` — #420 adds the field but doesn't bump the package or add a changeset. So today the button asks the agent to write fields the SDK rejects, then tells it to upgrade to a version that also rejects them. The agent will churn and give up.

This is a hard sequencing dependency on a **publish**, not a merge. Worth stating on the PR so it doesn't get merged-and-forgotten in the launch rush.

---

## Worth fixing

### 3. `sessionId` truthy ≠ session alive

```tsx
onDescribeWithAI={boundWorkflow && sessionId ? () => ... : undefined}
```

`sessionExited` is a prop on the same component. On an exited session the button still renders and the inject silently goes nowhere. `&& !sessionExited`.

### 4. "Rewrite descriptions with AI" overwrites human-authored text on one unconfirmed click

The design note reasons carefully about why auto-on-render would be too aggressive — agreed, and the explicit-button call is the right one. But the button still edits source files across every step with no confirmation. The **"Describe"** variant (fields empty) is fine as-is. The **"Rewrite"** variant is the one that can destroy text someone hand-wrote, and that's the one worth a confirm. The tooltip already says it writes to the source, which is the right instinct.

---

## Smaller notes

- **The unit tests assert a string contains its own substrings.** `expect(p).toContain("defineStep")`, `expect(p).toMatch(/do not change any logic/i)`, `expect(p.toLowerCase()).toContain("re-renders automatically")` — these break on any harmless copy edit and catch no real defect. The two e2e (affordance renders; the inject carries the right workflow identity and path) are the ones earning their keep. Not blocking, just don't expect the unit suite to protect anything.
- **`describeWorkflowPrompt` hardcodes "usually index.ts".** If the harness already knows the real entry file, thread it through; otherwise fine.
- **Placement is good** — on a workflow that already has a description, "Rewrite descriptions with AI" sits directly under the text it would replace, which reads clearly.
- **CSS vars check out** — `--accent`, `--accent-dim`, `--accent-border`, `--accent-foreground` are all defined, and the focus-visible outline is present.

---

**Recommendation:** hold until #420's SDK change is published to npm, and get one CI run on it. The code itself is small and clean.

gwitwer (dismissed):

gwitwer (approved):

PR #422, opened 2026-07-27
fix(harness): verify resumability before offering Resume (SAP-2057)

gwitwer:
Closes [SAP-2057](https://linear.app/sapiom/issue/SAP-2057/h1-verify-resumability-before-offering-resume-kill-the-phantom-rows) — H1 of Epic H (SAP-2056).

## Problem

Clicking Resume on a past session failed with `No conversation found with session ID: …` → exit 1 → a dead pane offering Resume again. **On my machine 16 of 49 registry entries are phantoms** — one in three rows was a guaranteed dead button.

The badge was `row.session.agentSessionId != null`, which only means our SessionStart hook fired once — not that the agent still has a conversation. Neither Claude Code nor Codex writes any transcript for a session that never received a prompt (our codex adapter already documented that rule from the other direction, in `detectBlockingPrompt`).

Transcript-only rows had the **inverse** bug: hardcoded `resumable={false}` even when the transcript was right there, so opening one silently started a fresh session and dropped the conversation. `registerHistorical()`, written to adopt exactly those rows, had no production caller.

## What changed

Resumability is now a question we ask the agent's own store, resolved in one place.

| | Before | After |
|---|---|---|
| Registry row, no transcript | `resumable` → exit 1 | `archived`, Resume disabled with the reason |
| Registry row, transcript present | `resumable` | `resumable` (verified) |
| Transcript-only row | `archived` → silently fresh session | `resumable` → adopted + genuinely resumed |

- **`HarnessAdapter.canResume(agentSessionId, cwd)`** — never throws. claude-code does one `stat` on `~/.claude/projects/<encodeProjectPath(cwd)>/<id>.jsonl`, reusing the existing encoder so the vendor layout stays defined in one place; codex matches `session_meta` id **and** cwd through the same `collectRolloutFiles` + `readSessionMeta` walk `listPastSessions` uses.
- **`SessionSummary.resumeMode: "agent-resume" | "rehydrate"`** — computed server-side in `GET /api/sessions/history` for *both* row sources. Adapters now return `PastSessionRecord` (`Omit<SessionSummary, "resumeMode">`), so the type makes it impossible for an adapter to decide this itself — which is how the two halves drifted in opposite directions.
- **`SessionManager.resume()` pre-flights `canResume()`** and throws the existing `SessionNotResumeableError` (already mapped to 409 with a `code`) instead of spawning a doomed pty. The error now carries a reason naming the agent, because that message reaches the user as a toast.
- **`POST /api/sessions/adopt`** wires up `registerHistorical()` at last: a transcript-only row the agent still holds is adopted into the registry and resumed for real. The server re-verifies resumability rather than trusting the client, and the route is idempotent.
- **Truthful durations.** `resume()` stamps `lastActiveAt` only to survive the dead-session sweeper during the pre-pty window; that stamp is now rolled back when no pty ever appears, so an idle session stops reporting "Ran for 6h 25m" purely because someone clicked Resume. `formatDuration` returns null on a zero span instead of inventing "under a minute" for an adopted row.
- **UI.** `PastSessionRow` renders the verified mode (`resumable` / `archived`, `checking…` until the probe lands rather than guessing either way). `PastSessionPane` and `DeadSessionPane` disable Resume with the real reason. Per the ticket, `rehydrate` rows offer "New session here" until H3 lands portable continue.

## Verification

- `pnpm build && pnpm typecheck && pnpm lint` clean (0 errors; the one harness lint warning is pre-existing on `main`).
- `pnpm --filter @sapiom/harness test` — **1286 passed**, including new `canResume` unit tests for both adapters against fixture home dirs (no real `~` access), `resume()` pre-flight tests, history `resumeMode` + adopt-route tests, and a new `history-meta.test.ts`.
- Playwright — **252 passed**. `smoke.spec.ts` now asserts phantom → `archived` with Resume disabled, and real transcript → `resumable` resuming into an adopted session rather than a fresh `sess-mock`.
- **Real-data check** against this machine's actual `~/.claude` + registry: 33 of 49 entries resolve `agent-resume`, 16 `rehydrate`, and a spot-checked phantom's encoded project directory does not exist at all — so every phantom now renders `archived`.

## Note on repo

The orchestrator metadata for this ticket said `sapiom/Sapiom`, but every path SAP-2057 names (`src/core/adapters/claude-code.ts`, `web/src/components/WorkflowsRail.tsx`, `src/server/rest.ts`) is in `sapiom-js`'s `packages/harness`, so the work landed here.

https://claude.ai/code/session_01Ae5tEwrTaiv71qty8Nt6gW


gwitwer:
## Review

The structural move here is the best thing in the diff: adapters returning `PastSessionRecord = Omit<SessionSummary, "resumeMode">` means an adapter *cannot* decide resumability, so the drift that made registry rows optimistic and transcript rows pessimistic is now unrepresentable rather than fixed twice by hand. Server-side re-verification on adopt, the idempotency check, and `agentHoldsConversation` swallowing throws *despite* the never-throws contract (with a test for it) are all the right instincts.

Four things worth addressing, two of them before merge.

### 1. `canResume` false-negatives on symlinked cwds → blocks resumes that work today

Claude Code records the **realpath** of the cwd; `canResume` encodes `session.cwd` verbatim. Verified against a real `~/.claude`:

```
~/.claude/projects/-private-tmp-claude-print-mode-test   ← exists
~/.claude/projects/-tmp-claude-print-mode-test           ← does not
node: realpathSync('/tmp') === '/private/tmp'
```

Also present: `-private-var-folders-g3--…-T-harness-bg-visualize-…`, from a harness background task whose cwd was `os.tmpdir()` (`/var/folders/…`, itself a symlink to `/private/var/folders/…`).

So a session in `/tmp/foo` has its transcript under the encoding of `/private/tmp/foo`, `canResume` stats the wrong path, and this PR **refuses a resume that succeeds on `main`** while labelling the row `archived`. `listPastSessions` shares the blind spot, but there it only omits a row (pre-existing behaviour); the pre-flight turns it into a blocked action plus a wrong label — a fail-closed regression in the exact surface being fixed.

Suggest `realpath`-ing the cwd (falling back to the raw string on failure) inside one helper used by both `canResume` and `listPastSessions` — which also fixes history discovery for symlinked project dirs — plus a fixture test with a real symlink.

### 2. The new dead-pane history fetch clobbers the whole history store

`loadHistory` **replaces** the store — `setHistory(Array.from(byAgentId.values())…)` with only the requested cwds' rows (`use-harness-state.ts:615-635`). The new effect in `App.tsx` calls `loadHistory([deadCwdNeedingHistory])` with a single directory, while both existing callers load broadly (`App.tsx:186`, up to 12 dirs; `:614`, the popover's cwds).

Opening an exited session whose row isn't already loaded therefore drops every other directory's rows, the rail's `resumeModeByAgentId` map goes empty for them, and their tags render **`checking…` indefinitely** — the boot effect doesn't re-run on a history change, so it only self-heals when the popover is reopened.

The comment says the present-row guard "keeps this from clobbering a broader load … with a single dir", but that guard only skips *redundant* fetches; in the case it exists for — row absent — it clobbers. Cheapest fixes:

- Make `loadHistory` merge into the store rather than replace (it already keys by `agentSessionId`); or
- drop the lookup entirely by putting `resumeMode` on `HarnessSession`, which also closes the probe window where a phantom's Resume is briefly clickable.

### 3. Codex pays a full rollout-tree walk per registry row

`CodexAdapter.canResume` walks the whole `~/.codex/sessions/YYYY/MM/DD` tree and `readSessionMeta`s each file until a match, and the history endpoint calls it once per registry row **sequentially** (`for … await`, `rest.ts:436-450`). For M codex rows that's M full walks on top of the one `listPastSessions` already did — every file opened, on a user-blocking dropdown open.

The PR already identified this cost and applied the mitigation only to the transcript branch: *"Re-probing every row would also turn codex's single rollout-tree walk into one walk per row."* Same argument applies to registry rows — resolve them against the single `listPastSessions(cwd)` listing first (by the PR's own reasoning, "finding it IS the verification" is equivalent proof), and probe only the misses, in `Promise.all`. That's one walk.

### 4. The route-ordering rationale is incorrect, and two tests guard a non-risk

> *"Registered ahead of `/sessions/:id/resume` because both are three-segment paths — reversed, `:id` would swallow 'adopt'."*

`/sessions/adopt` is **two** segments; `/sessions/:id/resume` is three and cannot match it. There's no `POST /sessions/:id` either (only `PATCH …/workflow`, `DELETE /sessions/:id`, `POST …/input`, `POST …/image`). Ordering is irrelevant here, and `"':id' never swallows 'adopt'"` asserts something that can't happen. Not a bug — but it's a wrong model of Express matching that the next reader inherits.

### Smaller

- **Two definitions of "has a conversation":** `canResume` accepts any file with `size > 0`; `listPastSessions` requires ≥1 parseable entry. A non-empty-but-unparseable transcript is `agent-resume` via the registry path and invisible via the transcript path.
- **The claude-code test re-implements the encoder** (`replace(/[/.]/g, "-")`, dropping the `:` strip) instead of importing it, so encoder drift wouldn't fail the tests that most need to catch it.
- **Inconsistent measurement:** tests and e2e say "15 of 46"; the body and changeset say "16 of 49".
- **`data-resumable`** now carries `"unknown" | true | false` — works, but a mixed type for a documented test hook.

### Tests

Strong where it counts: the phantom shape (id held, project dir populated by *other* sessions), cwd-scoping both ways, a directory named `<id>.jsonl`, codex shard-crossing, traversal rejection, `canResume` throwing anyway, adopt idempotency, and the `lastActiveAt` rollback — that last one is a subtle catch, and driving it with a throwing `spawnPty` is the right seam. Gaps: nothing covers the new `App.tsx` effect (§2 would have surfaced there), and there's no symlinked-cwd fixture (§1).

Right diagnosis, right architecture. §1 and §2 before merge — one reintroduces a wrong "can't resume" verdict from the other direction, the other degrades the same list this fixes. §3 is a real cost on a hot path with a cheap fix; §4 and the smaller items are cleanup.

gwitwer:
Addressed all four in `b18341e`. The symlink one was the real find — thank you.

### 1. Symlinked cwds — fixed, and it was worse than described

Reproduced before touching anything:

```
canResume(via symlinked cwd)  : false   <- want true
canResume(via real cwd)       : false
listPastSessions(symlinked)   : 0 rows  <- want 1
```

Note line 2: the *unsymlinked* project path also failed, because `mkdtemp` sits under `/var/folders/…`, itself a symlink to `/private/var/folders/…`. So it isn't only cwds the user typed through a link — anything under a symlinked ancestor is affected, which is why `-private-var-folders-…` shows up in a real `~/.claude` at all.

Both readers now share one `projectDirsFor(homeDir, cwd)`: resolved form first, raw form retained as a second candidate. Two reasons for the fallback rather than a straight swap — nothing that resolved before can stop resolving, and a cwd that no longer exists (realpath rejects) still answers. `listPastSessions` dedupes across the candidate dirs so a session can't surface twice, and its rows keep reporting the **caller's** cwd, not the resolved one, so a session started in a directory resumes in it.

Same reproduction after:

```
canResume(via symlinked cwd)  : true
canResume(via real cwd)       : true
listPastSessions(symlinked)   : 1 rows
```

Codex got the same tolerance on `session_meta.cwd`, with the honest caveat in the comment: **claude-code's realpath behaviour is verified, codex's is not.** Accepting either form can only find rollouts a raw comparison would have missed, and for a resumability probe, "no" for a conversation that exists is the failure mode worth engineering against. Fixture tests with real `symlink()` calls for both adapters, plus one for the realpath-fails path.

Re-ran the real-data check: still 33 `agent-resume` / 16 `rehydrate` — unchanged, because none of my 49 registry cwds happen to be symlinked. The fix is purely additive on this machine, which is worth saying rather than implying it moved a number.

### 2. `loadHistory` clobbering — fixed, and you're right that my comment described a guard that doesn't guard

The guard skips *redundant* fetches; in the case it exists for — row absent — it clobbered. `loadHistory` now replaces only the rows of directories that actually answered and retains the rest.

I kept it cwd-scoped rather than a blind merge, for the reason a merge can't handle: re-loading a directory must still be able to **drop** a row whose transcript is gone. Only `fulfilled` cwds count as refreshed, so a failed fetch retains what we had instead of reading as "refreshed to empty".

On your alternative — `resumeMode` on `HarnessSession` — I went the other way, and it's a genuine trade rather than a clear win. It would close the probe window, but `HarnessSession` is the *persisted* registry record, so a derived field either gets written to `sessions.json` and goes stale, or needs computing on every `GET /api/state` poll for every session. The window it closes is also narrower than it looks: the pre-flight is the real gate, so a click inside it 409s with the reason rather than spawning anything. Happy to revisit if you think the persisted-field cost is acceptable.

Coverage: extracted as a pure `mergeHistory` and unit-tested (5 cases, including the single-dir retention and the failed-fetch retention). Being straight about the limit — the repo's web unit tier is Node-env, framework-free by design (`vitest.config.ts` says React components belong to the Playwright tier), so I tested the extracted logic rather than the `App.tsx` effect itself. The effect's own symptom is hard to pin deterministically in Playwright because the popover load heals it.

### 3. Codex walk per row — fixed

Restructured as you suggested: the adapter scan runs **first**, once, and any row it found is settled without a probe — the same "finding it IS the verification" argument the transcript branch already used, which I'd applied asymmetrically. Only the misses are probed, in `Promise.all`. Two tests pin it: `canResume` not called at all when the scan covers the row, and `canResume.mock.calls` equal to exactly the missed row.

### 4. Route-ordering rationale — you're right, it was wrong

`/sessions/adopt` is two segments, `/sessions/:id/resume` is three, and there's no `POST /sessions/:id`. Removed the false rationale. I kept the test but reframed it as what it actually earns — a guard against a future `POST /sessions/:id` silently rerouting adopt through `resume()` — rather than asserting something that can't happen today.

### Smaller

- **Two definitions of "has a conversation"** — kept, now documented as deliberate. `canResume` accepting a non-empty-but-unparseable transcript is the *right* asymmetry: refusing a resume because **our** parser can't read the file would be the same fail-closed mistake as §1, and `--resume` is Claude's call to make. The §3 restructure makes the consequence explicit — such a transcript is settled by the probe, not the scan.
- **Test re-implemented the encoder** — fixed; `encodeProjectPath` is exported and imported by both helpers in the test file. Good catch: it was the one place a copy could hide drift.
- **15/46 vs 16/49** — reconciled. Mine is 16 of 49; the ticket's original was 15 of 46. Comments now cite the measured figure and note the ticket's separately.
- **`data-resumable`** — now always `"true" | "false" | "unknown"`, with a note that it's a documented test hook and a mixed type invites `=== "true"` checks that miss the unknown state.

### Verification

`pnpm build && pnpm typecheck && pnpm lint` clean (0 errors; the single harness lint warning is pre-existing on `main`). **1296 unit tests pass** (up from 1286).

Playwright: 251/252, failing on `step-macros.spec.ts:308`. I checked rather than assumed — a full run on pristine `main` fails **2** tests, and *different* ones (`rich-step-detail`, `welcome`). The suite is flaky under parallel load on this machine independent of this branch; `step-macros` passes in isolation here. Not claiming it green — if CI reproduces a specific failure I'll chase that one.

positiveblue (approved):

PR #423, opened 2026-07-27
feat(harness): rebuild and render session history from our own events (SAP-2058)

gwitwer:
Closes [SAP-2058](https://linear.app/sapiom/issue/SAP-2058/h2-rebuild-and-render-session-history-from-our-own-events) (Epic H, parent SAP-2056).

## Why

Resume-as-a-viewer could never work. Relaunching a CLI hands us a pty that redraws its own scrollback, so the Studio saw pixels, not content — and only for Claude Code, only on the machine that ran the session. Meanwhile our own recording already held the conversation: `prompt.submitted` (full text), `tool.call` (name + input + response summary), `turn.completed` (assistant text + model + tokens).

So our recording becomes the history browser. Because it reads *our* normalized events rather than any vendor's files, it works identically for claude-code and codex — and it survives the vendor deleting, rotating, or never writing its own transcript.

## What changed

**Server**

- `EventStore` (`src/core/collector/store.ts`) gains a streaming `read(filter)` — `node:readline`, tolerant of a torn final line, deliberately **outside** the append queue.
- A byte-offset index keyed by `harnessSessionId`, built lazily on first request and extended afterwards by scanning only the bytes appended since — so opening a record never rescans the (up to 50 MB) ndjson. A shrink or inode change (a retention sweep) rebuilds it.
- New `src/core/session-record.ts` folds events into ordered turns: `prompt.submitted` opens, `tool.call`s attach, `turn.completed` closes with `assistantText` / `model` / `usage`. A trailing open turn is kept and marked incomplete.
- **Ordering is `(ts, seq)`, never `seq`.** That counter lives in the server's memory (`collector/seq.ts`), so it restarts at 1 on every boot and every fresh pty — a real resumed session reads seq 1, 2 then 1, 1, 1. A reset is a new process epoch, not loss. `AnalyticsEvent.seq`'s doc comment claimed per-session monotonicity and is corrected here.
- `GET /api/sessions/:id/record` → `SessionRecord`, resolving either a `harnessSessionId` or the agent's own session id (transcript-sourced history rows only have the latter). Harness sessions sharing an agent session — a resumed conversation — fold into one record.
- History rows gain an exact `turnCount` from the index, which outranks the claude adapter's `messageCount` (`undefined` above its 5 MiB full-scan cap). A failing reader degrades the rows, never the request.

**UI**

- `PastSessionPane` is now a read-only transcript: prompts, assistant text, collapsible tool calls, model + tokens per turn.
- `DeadSessionPane` shows it too. That was not in the ticket text but is required by its acceptance criterion — an exited registry session (which is what a Codex past session *is* in the fixtures) routes there, not to the review pane. The metadata card is unchanged; the transcript is added below it, and a session with nothing recorded keeps exactly the pane it had.

## Honesty, in the UI and not just the ticket

The view is labeled **Reconstructed**, and the record carries machine-readable `limitations` that the UI spells out in prose:

- `truncated-tool-output` — `tool.call` stores a size-capped `toolResponseSummary`; the collapsed tool row is marked `truncated` and the expanded body says the rest isn't recoverable.
- `assistant-narration-gap` — `turn.completed` carries only the Stop hook's **last** assistant message (see the header of `collector/transcript.ts`), so narration *between* tool calls is absent. The tool-call stream fills the chronology, not the prose.
- `missing-assistant-text` — Codex's rollout has no equivalent field, so its turns carry prompts and tool calls and say outright that replies weren't recorded.
- `incomplete-final-turn` — a session killed mid-turn.

Nothing is inferred to fill a hole: a turn with no recorded prompt says so rather than borrowing one. Vendor-transcript enrichment (claude-code only, reusing `readLastAssistantTurn`'s tail-read) fills the last **completed** turn's model/usage/text when that file happens to exist — never an incomplete turn, where the transcript's last message would belong to some earlier turn, and never overwriting what our own events recorded.

## Two decisions worth a reviewer's attention

**Index mutation is confined to one serialized reconcile.** Writing the index entry from `append` as well is the obvious optimization and it is wrong: a reader's reconcile can stat a size that already includes a line whose write is still in flight, so both paths record the same event and every offset after it shifts by a line. I wrote it that way first; the invariant test (30 concurrent appends interleaved with 30 reads) caught it at 53 events instead of 31. The cost of one writer is a tail read on the next `index()` call, bounded by what was appended since the last one.

**Span reads use one file handle with positioned reads, not a stream per span.** A session interleaved with others has one span per contiguous run of its lines — ~1900 of them in the 50 MB fixture — and standing up a `createReadStream` + readline for each costs far more than the bytes do. Spans are still read in 1 MB chunks (a session owning the whole file can't be pulled into memory at once), with a `StringDecoder` carrying multi-byte characters across chunk boundaries.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm lint` — all green across the workspace (0 errors; the one harness lint warning is a pre-existing unused import in `rest.test.ts:7`).
- Harness unit suite: **1307 passing**. New coverage: the fold (interleaved sessions, the `seq` restart, a trailing incomplete turn, a torn final line, a `tool.call` with no enclosing turn, the Codex shape, merged sessions, enrichment present/absent/never-overwriting), the store (`read` filters, torn/blank/non-event lines, multi-byte offsets, reads not blocking behind `runExclusive`, index maintenance, a foreign process's appends, rebuild after a sweep, the interleave invariant), the record route + history `turnCount`, and the view helpers.
- Playwright: **258 passing**, including 7 new specs covering claude-code and Codex transcripts, the reconstruction label and its named gaps, truncation and incomplete turns, collapsible tool calls, the no-events empty state, and the pane scrolling without the shell scrolling.
- Perf (acceptance: p95 record render < 300 ms against a 50 MB ndjson): **cold 126 ms (index build), hot p95 23 ms, median 22 ms** over 1878 events. Plus a timing-free companion assertion that a record open reads under a tenth of the file — the wall clock alone can't tell "the index worked" from "a full scan happened to be fast on this disk".

Perf tests run in their own sequential pass (`vitest.perf.config.ts`, chained from `pnpm test`) because a timing assertion measured against ~90 parallel test files measures the runner: the same read timed 27 ms alone and 340 ms alongside the suite.

**Pre-existing flake, not from this branch:** running every package's suite back-to-back, one fs-watch/timing test fails per run and it's a different one each time (`codex-tailer-wiring`, `canvas-watcher`, `auto-bind-rescan`). I confirmed this on a stashed clean `main` checkout, which flaked the same way. Each passes in isolation.

https://claude.ai/code/session_01SRnGrJbFsERreebvvLG7JF


gwitwer:
## Review

The engineering in the store is the strongest part of this. Three things I checked rather than trusted:

- **The single-writer index invariant.** Writing the index from `append` as well is exactly the bug I'd expect this design to ship with; it's already found, fenced, and covered by an interleave test that caught it (53 events instead of 31). The `runExclusive` → `state = null` drop plus inode-change detection closes the sweep path too.
- **The torn-line rewind in `scanInto`.** I traced all four cases — torn JSON, complete event with no newline yet, blank final line, complete line — and it's correct in each, including not double-counting the event whose newline arrives on the next reconcile.
- **`readSpanLines`.** One handle with positioned reads, `StringDecoder` carried across chunk boundaries, and offsets advanced by `Buffer.byteLength` rather than character count. Those last two are precisely the mistakes that would silently corrupt every span after a multi-byte prompt.

Also verified rather than assumed: `Markdown` is `dangerouslySetInnerHTML`-free (safe for model-authored text), tool input/results render inside `<pre>`, and `.transcript-prompt` sets `white-space: pre-wrap` — so the transcript is injection-safe and preserves formatting.

Four things worth addressing.

### 1. This PR and #422 break each other — land one, rebase the other

Both branch from the same base (`a0011ec6`) and overlap on **12 files**. This isn't only conflict risk; there's a concrete typecheck break:

- **#422 makes `SessionSummary.resumeMode` a required field.** This PR adds a new `MOCK_HISTORY` fixture literal (`source: "transcript"`, `mock-data.ts`) without it. Whichever merges second fails `typecheck` until someone patches the fixtures.
- `App.tsx`: #422 *removes* `resumable` from `PastSessionPane` while this PR *adds* `loadRecord` to the same JSX block, and both change the `DeadSessionPane` signature.
- `historyRowMeta`: #422 adds a `now` parameter; this PR adds `turnCount` preference — same function body.
- Both rewrite the same `GET /sessions/history` handler and the same region of `claude-code.ts`.

Neither PR's suite (1286 / 1307 passing) reflects the merged state, and GitHub already reports `mergeable_state: blocked`. Suggest merging #422 first and rebasing this on it — the H1→H2 order the epic already implies.

### 2. `assistant-narration-gap` is not recomputed after enrichment

`foldSessionRecord` computes `limitations`, then `createSessionRecordReader` fills the final turn's `assistantText` from the vendor transcript and only *removes* `missing-assistant-text`:

```ts
limitations: turns.some((t) => t.completedAt !== null && t.assistantText === null)
  ? record.limitations
  : record.limitations.filter((l) => l !== "missing-assistant-text"),
```

If that enriched turn has tool calls, the record now satisfies the `assistant-narration-gap` condition (`toolCalls.length > 0 && assistantText !== null`) and never says so — the UI under-reports a real gap, in the PR whose thesis is stating gaps out loud. Recompute limitations from the final turn list rather than patching one code.

### 3. The realpath blindness from #422 is duplicated here

`createClaudeTranscriptEnricher` builds `~/.claude/projects/<encodeProjectPath(cwd)>/<id>.jsonl`. Claude Code records the **realpath** — verified against a real `~/.claude`: `-private-tmp-claude-print-mode-test` exists, `-tmp-claude-print-mode-test` does not, and `realpathSync('/tmp') === '/private/tmp'`. So enrichment silently no-ops for any symlinked cwd (`/tmp/*`, anything under `os.tmpdir()`).

Harmless *here* — enrichment is best-effort by design — but it's the same latent bug that is load-bearing in #422's `canResume`, now in a second place. This PR already exports `encodeProjectPath` to stop the *encoding* from drifting; extend that to the whole path construction, realpath included, so both PRs inherit it from one helper. There will be three callers post-merge (`canResume`, the enricher, `listPastSessions`).

### 4. Full-record buffering is the real memory ceiling

`reader.read` pushes every event into an array, then `sortEventsForFold` copies it again. The chunked span reader carefully avoids holding the *file* in memory — and then the parsed events for one session are all resident, twice over during the sort.

At the observed real-world average (16 MB / 3737 events ≈ 4.3 KB/event), a session owning most of a 50 MB log is ~11k events — tens to low hundreds of MB of heap in a long-lived local server. The perf fixture doesn't exercise this: 1878 events for 50 MB is ~26 KB/event, roughly 6× fewer events than a realistic log of that size.

A global `(ts, seq)` sort makes streaming hard, so this may be an acceptable ceiling — but it should be a stated one, or bounded with a `?limit=` / most-recent-N on the wire. Worth an explicit answer before H3's rehydration reads from this.

### Smaller

- **`pnpm test` now writes a 50 MB fixture and runs a second sequential vitest pass** on every invocation, local included. Flake risk is genuinely low — 300 ms budget against 23 ms measured, 15 s cold against 126 ms, plus the timing-free `spanBytes < fileBytes / 10` assertion doing the real work — but consider a smaller default with an env knob for the full-size run.
- **`turnCounts()` sets harness-session keys to the *segment* count** while `read()` folds the merged conversation. The endpoint prefers the agent key, so the disagreement is nearly unreachable — but the fallback can contradict the record it links to. Sum for both key kinds, or drop the fallback.
- **`index()` returns the live internal maps** typed as `ReadonlyMap`; a later `reconcile()` mutates entries in place, so a held reference's `turnCount` / `lastTs` can change under the caller. `read()` correctly deep-copies spans — the counts are the leaky part.
- **`formatClockTime` renders only `HH:MM`**, so a transcript spanning days shows repeating times with no date boundary.
- This PR's route-ordering comment (*"express matches on the full path, so ordering here is cosmetic"*) is **correct**, and directly contradicts #422's claim that `:id` would swallow `adopt`. Fix #422's, keep this one.

The store and fold are careful, well-tested work, and the honesty model (`reconstructed: true` on the wire, limitation codes rather than prose, an unknown-code fallback, refusing to enrich an incomplete final turn) is the right design for a reconstruction. §1 is the gating issue and it's a sequencing decision rather than a code fix; §2 and §3 are small.

gwitwer:
Thanks — §2 and §3 were real, and §3 I verified rather than took on faith. Pushed as 493b604.

### 2. `assistant-narration-gap` after enrichment — fixed

Correct, and it's the worst kind of bug for this PR to ship. Extracted `computeLimitations(turns)` and call it again after enrichment instead of patching one code:

```ts
return { ...record, turns, limitations: computeLimitations(turns) };
```

Filling that turn closes `missing-assistant-text` and *opens* `assistant-narration-gap` in the same move. Regression test asserts exactly that transition — the enriched turn has a tool call, and the record must end up claiming the narration gap it didn't claim before.

### 3. Realpath — fixed, and confirmed against a real `~/.claude`

```
$ ls ~/.claude/projects/ | grep -E '^-(private-)?tmp-claude-print-mode'
-private-tmp-claude-print-mode-test
-private-tmp-claude-print-mode-test2
```

No `-tmp-` variant. So enrichment was silently no-opping for every symlinked cwd, as described.

Path construction now lives in one helper rather than just the encoding:

```ts
export async function claudeProjectDir(cwd: string, homeDir?: string): Promise<string>
export async function claudeTranscriptPath(cwd, agentSessionId, homeDir?): Promise<string>
```

`encodeProjectPath` stays exported but its doc now says it takes an already-resolved path. New test covers it via `os.tmpdir()` (itself symlinked on macOS): the transcript is placed *only* at the resolved encoding, so a hand-built path returns null and the test fails.

I deliberately did **not** convert `listPastSessions` (or #422's `canResume`). Both live in the same lines #422 is rewriting, and changing them here maximises the conflict you're already flagging in §1. The helper is there for the rebase to adopt — three callers, one implementation.

### 1. Sequencing — agreed, #422 first

No argument. Both are `BLOCKED` on review, neither is merged, so there's nothing to rebase onto yet. Plan: **#422 lands, I rebase this on `main`, add `resumeMode` to the new `MOCK_HISTORY` fixture, reconcile the four overlapping call sites (`PastSessionPane` props, `DeadSessionPane` signature, `historyRowMeta`, the history handler), and adopt `claudeTranscriptPath` in `canResume` + `listPastSessions` in the same pass.** Ping me when it merges. H1→H2 is the epic's order anyway.

### 4. Buffering ceiling — measured, and one premise is off

Measured rather than argued, with a second fixture at the density you cite (~4.3 KB/event, one session owning the whole file):

```
[perf] dense 50MB single session · 12062 events · record content 50MB vs 52MB on disk · heap delta 2MB · 99ms
```

**One copy, not two.** `sortEventsForFold` does `[...events].sort(...)`, which copies the array of *references* — 8 bytes per event, ~96 KB here — not the events. The turns then reference the same payload strings the events do. So the peak is one copy of the session's own bytes, not the doubling the review assumes.

That makes the ceiling statable, which is what you asked for, and it's now on `createSessionRecordReader`: *"one session's share of a file capped at 50 MB"* — worst case ~50 MB transient in a local server. Asserted as a ratio (`contentBytes < spanBytes * 1.5`) so a regression that duplicates the data or buffers the raw file trips it, rather than a wall-clock number that says nothing. If H3 needs tighter, `limit`/most-recent-N on the wire is the fix; I've said so in the doc rather than building an API this ticket doesn't need.

(The 2 MB heap delta is noise — no `--expose-gc`, so a collection during the read moves it either way. It's logged, not relied on.)

### Smaller

- **`turnCounts()` segment vs merged** — fixed. Both key kinds now report the whole conversation `read(id)` would fold, via the same `resolveSessionIds`. A row can't contradict the record it links to regardless of which id keyed it.
- **`index()` returning live maps** — fixed. Returns a snapshot (entries, spans, and the alias arrays copied). Test holds a result across an append and asserts it doesn't shift under the caller.
- **`formatClockTime`** — fixed. `"14:32"` today, `"1 Jul, 14:32"` otherwise. Test uses the same clock time three days apart, which was indistinguishable before.
- **50 MB fixture on every `pnpm test`** — I'd rather keep it, and the numbers are why: the whole perf pass is **~1.2 s wall-clock** including writing both fixtures. A smaller default would mean `pnpm test` no longer checks the acceptance criterion it exists to check, and the env knob would be the thing nobody sets. Happy to shrink it if the number changes on slower CI, but at 1.2 s it isn't paying rent yet.
- **Route-ordering comment** — noted, leaving mine.

Full suite after the fixes: **1312 unit** (5 new), **258 Playwright**, typecheck + lint clean.

One flake note, since it'll show up if you re-run: `rich-step-detail.spec.ts` fails intermittently, a different test in the `Capability calls block` / `Links block` describes each time. I checked it against a stashed clean `main` — 2 of 3 runs failed there too, so it's pre-existing and not from this branch.

gwitwer:
Rebased onto `main` now that #422 has landed — §1 resolved. Head is `093a9b9`, `mergeable: MERGEABLE`.

Ten of the twelve overlapping files conflicted. How each was reconciled:

**Adopted #422's design over mine where it was better.** I'd added `claudeProjectDir` / `claudeTranscriptPath` (single resolved path) for the realpath fix. #422's `projectDirsFor` is strictly stronger — it returns the resolved candidate *and* keeps the raw one as a fallback, so a cwd that no longer exists on disk still resolves. I deleted my helpers, exported `projectDirsFor`, and the enricher now walks its candidates. One implementation, three callers (`canResume`, `listPastSessions`, the enricher), as §3 asked for.

**`PastSessionPane`**: dropped my `resumable` prop entirely; it now derives from `summary.resumeMode === "agent-resume"`, and keeps #422's honest blocked-reason copy. `loadRecord` is the only prop I add.

**`DeadSessionPane`**: takes both `resumeMode` and `loadRecord`. Git's merge had duplicated half the `<dl>`, so I rewrote the file rather than untangle it. Worth stating explicitly, because the combination is new and the two PRs pull in opposite directions:

> A session the agent **can't** resume can still have a **full transcript** here.

Resumability is a question about the agent's store; readability is a question about ours. `rehydrate` + a complete transcript is now a normal, coherent state, and the pane says both things at once instead of treating "can't continue" as "nothing to see". There's a mock fixture for exactly that pairing (`agent-resume` with no recorded events — the other diagonal).

**`historyRowMeta`**: one signature carrying both changes — #422's injectable `now`, my `turnCount` preference over `messageCount`.

**`history-meta.test.ts`** (add/add — we both created it): kept #422's file wholesale and added only the two assertions of mine that weren't already covered by it (turnCount outranking messageCount; turnCount rendering alone). Dropped my four duplicates rather than leave two overlapping suites.

**`rest.ts`**: `/sessions/adopt` and `/sessions/:id/record` both present; the history handler does #422's `resumeMode` verification and then stamps `turnCount`. Also fixed my route-ordering comment to name `adopt` explicitly, since that's the sibling the ordering question was actually about.

**`MOCK_HISTORY`**: the predicted typecheck break — my new fixture now carries the required `resumeMode`.

### Verification

Per a new project rule (no local full-suite runs — they exhaust memory across parallel agent sessions), the gate I ran is `pnpm build && pnpm typecheck && pnpm lint`: **all green, 0 errors.**

I made one bounded exception, disclosed: I hand-merged `history-meta.test.ts`, and hand-merged test files are exactly what shouldn't go to CI unverified. So I ran those two files individually, single-worker, sequentially — `history-meta.test.ts` (13 passed) and `rest.test.ts` (69 passed). Nothing wider. The rest is CI's.

evtran0209 (approved):

PR #424, opened 2026-07-27
feat(examples): declare the template setup surface, and gate the copy the card renders (SAP-2076)

gwitwer:
Closes SAP-2076 — https://linear.app/sapiom/issue/SAP-2076

Foundation for the A1 epic ([SAP-2061](https://linear.app/sapiom/issue/SAP-2061)). Blocks SAP-2077 / 2078 / 2079 / 2080.

Two commits. The first is SAP-2076 as ticketed. The second came out of a design review of the template overview card and is unticketed — happy to file one and retro-link.

---

# 1. `d2b9c37` — the declaration surface (SAP-2076)

## The drift this closes

`examples/template.schema.json` is `additionalProperties: false` and its properties were exactly `$schema`, `manifestVersion`, `author`, `longDescription`, `useCases`, `notes`, `examples`. **`requiredSecrets` was absent** — even though `template-registry.service.ts:135` has parsed it for months and `TemplateDetailDto` already ships it to the frontend. An author adding the field today writes something their own CI rejects. That is why 0 of 26 templates declare a credential, and why the credential dialog has never rendered in production.

Separately, `examples-check.mjs` **never opened a `template.json`** — it only checked the file existed. That is exactly how `repoSlug: "my-app"` shipped.

## What changed

**`examples/template.schema.json`** — four new optional top-level properties:

| Field | What it declares |
|---|---|
| `requiredSecrets[]` | Provider-aware third-party credentials (`key`, `label`, `provider` required). |
| `settings[]` | Non-secret config with a **required** `default` — where a `RECIPIENT` belongs instead of the write-only vault. |
| `defaultInput` | The input a zero-interaction run starts with. Distinct from `examples[0].input`, which is documentation. |
| `zeroSetup` | The machine-checkable expectation: `terminalState`, a closed `expect[]` assertion vocabulary, and the verbatim `narrative`. |

**`examples/registry.schema.json`** — the `setup` block on `definitions.template`, denormalised on purpose because the shelf renders from `listTemplates`, which never fetches manifests. No `pauses` flag: approval state is already derivable from `steps[].checkpoint`, which the check already enforces.

**`scripts/examples-check.mjs`** — a `manifest-schema` check validating every `<sourcePath>/template.json`, reusing the ajv instance already built for the registry. Plus `manifest-secret-key` for the process-env namespace rules the pattern can't express, and `manifest-parse` so malformed JSON reports instead of throwing.

## A declaration never names a storage location

No `vaultRef`, no `connectorId`, no `store` — in any of the four. `key` is the logical identifier the resolver binds (`process.env[key]` today), which is what lets resolution move vault → connectors later without touching a template or the deploy panel. There's a test asserting the schema stays that way.

## Verification

`pnpm examples:check` — all 26 existing manifests validate **unchanged** (every new field is optional, so this holds before any migration). End-to-end against a deliberately broken manifest:

```
examples/ failed validation (5 problem(s)):

  - manifest-schema: "nl-db-query-endpoint" template.json /requiredSecrets/0 must have required property 'label'.
  - manifest-schema: "nl-db-query-endpoint" template.json /requiredSecrets/1 must NOT have additional properties.
  - manifest-schema: "nl-db-query-endpoint" template.json /settings/0 must have required property 'default'.
  - manifest-schema: "nl-db-query-endpoint" template.json /zeroSetup/expect/0/assert must be equal to one of the allowed values.
  - manifest-secret-key: "nl-db-query-endpoint" template.json /requiredSecrets/0/key "SAPIOM_TOKEN" is reserved — a secret key cannot be PATH and cannot start with SAPIOM_ or WORKFLOWS_, because the engine never lets a template override its own namespaces.
```

---

# 2. `0ff6184` — house-style rules for the copy a buyer reads

## The problem, measured

| | Today (26 templates) | The card |
|---|---|---|
| Names with a mechanism word | **11 / 26** | — |
| Names containing `→` or `/` | **5 / 26** | — |
| `whatItDoes` starting with the word *"For"* | **19 / 26** | starts with a verb |
| `whatItDoes` length | median **651**, max 1071 | ~200 |
| `useCases` item length | median 96, **min 62** | **14–27** |

The shortest of our 79 use cases is 62 characters — more than double the longest one the card has room for. Not one of them fits.

Root cause: **every prose field in both schemas is `{"type": "string"}` with no constraint of any kind.** The house style has only ever lived in `AUTHORING.md` prose no tool reads. Same failure as `requiredSecrets` above — a rule that existed, nothing checked.

## The naming rule is not new policy

`registry.schema.json`'s own `category` description already states it:

> The axis is OUTCOME — "what am I trying to produce?" — not mechanism… Mechanism words (durable, pause-resume, hitl, evals, media, orchestration) stay in freeform `tags`.

We enforced that on `category`, where an enum made it free, and never on the field a user reads and searches. And **five of the eleven offending names repeat a word the same template already carries as a tag**:

| Name | Word | Tags today |
|---|---|---|
| Multi-Party Approval Chain **(Saga)** | `saga` | `approval`, **`saga`**, `pause-resume`, `durable` |
| **Durable** Backfill / Long-Job Runner | `durable` | **`durable`**, `backfill`, `batch`, `cron` |
| Eval **Gate** | `gate` | `evals`, **`quality-gate`**, `llm-judge`, `composition` |
| **Human-in-the-Loop** Approval | `human-in-the-loop` | `approval`, **`human-in-the-loop`**, **`hitl`**, `fallback` |
| Natural-Language DB Query **Endpoint** | `endpoint` | `database`, `sql`, **`endpoint`**, `llm` |

Cutting the word from the name loses nothing in search and buys the title back for the job.

## What this adds

`scripts/examples-copy-check.mjs` — four rules, all **warnings**:

- `copy-name` — ≤32 chars; no `→` `/` `( )`; no mechanism word (narrow denylist: `saga`, `engine`, `pipeline`, `runner`, `endpoint`, `gate`, `durable`, `keep-alive`, `multi-party`, `human-in-the-loop`, `orchestrator`, `daemon`, `worker`, `handler`)
- `copy-description` — ≤160
- `copy-what-it-does` — ≤320, and reject a leading `"For "`
- `copy-use-case` — ≤40 per item

## Why warnings, and why not `maxLength` in the schema

All four limits fail all 26 templates today, so a schema `maxLength` would be a hard CI failure on day one and a gate would block every PR. This follows the play already in `examples-check.mjs` for `category`/`cadence` — *"a nudge, not a gate — flip to `errors.push` once every template carries them."*

```
copy style: 145 warning(s) across 26 of 26 template(s) — not a gate yet:
```

145 is the starting burn-down: 79 `copy-use-case`, 45 `copy-what-it-does`, 20 `copy-name`, 1 `copy-description`. When it reaches zero, the caps move into the schemas as `maxLength` and the style rules flip to errors.

## The copy itself

A redline for all 26 — proposed names, verb-first `whatItDoes`, three short use cases each — is drafted in the Sapiom repo at `plans/onboarding-templates/copy-clarity-redline.md`. Run back through this checker it scores **145 → 0**. It is a redline, not a decision; names especially are a judgment call, and no `id` changes.

---

# Verification (both commits)

- `pnpm build && pnpm typecheck && pnpm lint` — exit 0. Per a new project rule this branch does **not** run the jest suite locally; CI owns it.
- `pnpm examples:check` — 26/26 manifests schema-valid, exits 0; copy warnings do not gate.
- `pnpm examples:check:test` — **44 tests pass** (`node --test`, single process, no jest workers), wired into the `examples` CI job. Covers every SAP-2076 acceptance criterion, plus: every denylisted word caught case-insensitively, `Engine` does not fire on "Engineering", `gate` does not fire on "Aggregate", `"Format"` is not `"For"`, plain-English product words (PR Review Bot, Newsletter Autopilot) left alone, and `checkCopy` never throws so nobody can promote these to errors by accident.

# Also

`examples/AUTHORING.md` documents both surfaces — the four declaration fields, and the house-style limits. It previously told authors `examples:check` only verifies a `template.json` *exists*, which is no longer true.

# Open, not decided

**Should `whatItDoes` move from `registry.json` into `template.json`?** Not done here. It is detail-only data (`toSummary` omits it; only `toDetail` reads it at `template-registry.service.ts:356`) living in the thin registry index, where it is **25% of the file — 17KB of prose the shelf never renders**. Moving it would leave `TemplateDetailDto` unchanged, so the frontend is untouched, but it needs an expand/contract across the pinned-ref boundary. Worth its own ticket.

https://claude.ai/code/session_01SCBPYYakJZZaPysmTECLix


gwitwer:
### Update — `whatItDoes` moved, and the copy redline is applied (`dbe347e`)

Third commit lands the two things this PR was leaving open.

**`whatItDoes` moved from `registry.json` into `template.json`.** It was detail-only data in the thin gallery index: `toSummary` omits it, only `toDetail` reads it, and `toDetail` already loads the manifest — so this costs zero extra fetches. It was **25% of `registry.json`, 17KB of prose** every `listTemplates` caller downloaded and none rendered. Its three neighbours on the same card already lived in the manifest.

**The copy redline is applied, not just drafted** — 18 renames, 26 verb-first `whatItDoes` rewrites, 26 `useCases` sets rewritten as chips, one over-long description. Names say the outcome now:

| Was | Is |
|---|---|
| Natural-Language DB Query Endpoint | Ask Your Database |
| Multi-Party Approval Chain (Saga) | Sequential Sign-Off |
| Human-in-the-Loop Approval | Approval Before Action |
| Durable Backfill / Long-Job Runner | Large Dataset Backfill |
| Scheduled Compliance Audit + Attestation | Signed Compliance Audit |
| The Brain | Workflow Fleet Monitor |

**No `id` changed** — those are directory names, `sourcePath`s, and the gallery's stable keys.

**Nothing is lost from search.** Five of the retired mechanism words were already tags on the same template. The five that weren't (`pipeline`, `keep-alive`, `natural-language`, `quote`, `logs`) moved into `tags`, which is what actually drives search.

**Burn-down is zero, so the rules are gates.** The length caps moved into the schemas as `maxLength` (name 32, description 160, whatItDoes 320, useCases[] 40) so ajv reports them with a JSON pointer; `examples-copy-check.mjs` keeps only the two rules a `pattern` could reject but not explain. Proof they bite:

```
examples/ failed validation (6 problem(s)):

  - schema: /templates/23/name must NOT have more than 32 characters
  - manifest-schema: "the-brain" template.json /useCases/0 must NOT have more than 40 characters.
  - copy-name: "the-brain" name "Workflow Fleet Orchestrator (Saga) / Runner" contains an arrow or a slash — …
  - copy-name: "the-brain" name "…" carries a parenthetical — nobody searches for the part in brackets. Move it to `tags`.
  - copy-name: "the-brain" name "…" is built on the mechanism word "Orchestrator" — …
  - copy-what-it-does: "the-brain" whatItDoes opens with "For" — lead with the verb …
```

### One thing a reviewer must not miss

**The backend still reads `whatItDoes` off the registry entry.** The "What it does" section on `/templates/[id]` goes empty when the pinned ref advances past this commit. The pin is the safety valve — nothing breaks until someone advances it — and [SAP-2089](https://linear.app/sapiom/issue/SAP-2089/read-whatitdoes-from-the-manifest-not-the-registry-entry) covers the backend read with an expand/contract fallback. It has to land before the pin moves. `TemplateDetailDto` doesn't change, so there is no frontend work.

### Verification

`pnpm build && pnpm typecheck && pnpm lint` exit 0 · `pnpm examples:check` clean, now gating · `pnpm examples:check:test` 40 pass.

gwitwer:
## Add `resources[]` (with `seed`) before this merges

This PR is the declaration surface, and it declares four of the five requirement kinds: `requiredSecrets`, `settings`, `defaultInput`, `zeroSetup`. The missing one is the resources a template provisions — a Postgres, a sandbox, a repo, an inbox — and its absence blocks two things downstream that nothing else can unblock.

Raising it here rather than as a follow-up because a schema change wants to land as one thing, and both consumers below are being built against this file.

### Why it blocks the setup UI

The setup panel's first group is "here's what Sapiom provisioned for you." With no declaration, the only available source is a tenant-wide resource list — and resources key on the **payment account**, not the definition, so a resource can't be attributed to a workflow. The best that surface can honestly say is "resources in your account this workflow *can* use," which is a dodge around the missing field rather than a design.

Declared, it gives three things nothing else does:

- **Before the run:** "this creates a Postgres database, 7-day lifetime" — real cost and lifetime disclosure on the card.
- **After the run:** the actual handle and `expiresAt`, so a returning user sees their database is about to expire instead of silently getting a fresh empty one. Managed Postgres caps at 7 days and **has no renew verb**, so this is the single most likely way a returning user's workflow breaks.
- **Reuse has somewhere to put the answer.** "Use my existing database" means writing a handle into a declared slot. With no slot there is nothing to bind the choice to.

### Why it blocks a non-empty first run

Provisioning alone doesn't produce an onboarding experience. A template that reads from a database, handed a freshly-provisioned empty one, runs green and produces nothing — that's 7 of the current 26: terminal, honest-looking, empty. `zeroSetup.expect[]` in this PR will assert against those runs, and they'll fail on content that can't exist yet.

What fixes it is provisioning **plus seed data**, which means the declaration carries a `seed`.

To be explicit about what this is *not*: not canned capability responses. That was sample mode, and it's cut — a run labelled "sample" that really posts to Slack is a hazard. Real resource + real seed = a real run, nothing pretends.

### Shape

```jsonc
"resources": {
  "type": "array",
  "description": "Sapiom-managed resources the template needs. Declarative so deploy can pre-provision and seed, and the UI can state cost and lifetime BEFORE a run.",
  "items": {
    "type": "object",
    "required": ["kind", "handle"],
    "additionalProperties": false,
    "properties": {
      "kind": { "enum": ["postgres", "sandbox", "repository", "inbox"] },
      "handle": {
        "type": "string",
        "pattern": "^[a-z0-9][a-z0-9-]{0,62}$",
        "description": "Stable lookup handle the step code addresses it by (ctx.sapiom.database.get(handle))."
      },
      "duration": {
        "enum": ["15m", "1h", "4h", "24h", "7d"],
        "description": "postgres only. 7d is the maximum and there is NO renew verb — a managed Postgres is scratch space, not a datastore."
      },
      "seed": {
        "type": "string",
        "description": "Path (relative to the example dir) of a .sql file run once after provisioning, so the first run has data to work with. READ-SIDE ONLY — see below."
      },
      "ephemeral": {
        "type": "boolean",
        "default": true,
        "description": "False signals the 7-day TTL problem applies and the template needs a durability story in `notes`."
      }
    }
  }
}
```

### The one rule a fast pass gets backwards

**Only seed what a template reads.** Five templates provision tables that are their own *output* — `error_clusters`, `crm_contacts`, `outreach_touches`, the fleet event log, approval transitions. Seeding those fabricates data a user might act on. Four read-side templates need a seed. A contract phrased "give every resource a seed" instructs the opposite, so `AUTHORING.md` should say which side of the line a table is on rather than "add a seed."

### Bonus: it removes a hand-maintained field

`registry.setup.provisions[]` in this PR is author-declared with no source of truth to check it against — the same drift that let `capabilities[]` fill up with SDK method paths instead of catalog ids. With `resources[]` in the manifest, `provisions[]` becomes derivable, so the sync script generates it and CI can verify rather than trust.

### Also, separately

<https://linear.app/sapiom/issue/SAP-2089> — `whatItDoes` moves from `registry.json` into the manifests here, but the backend still reads it off the registry entry, so **"What it does" goes empty on `/templates/[id]` the moment this merges.** That fix needs to land with this PR or immediately behind it.

gwitwer:
### Both points addressed — `resources[]` in `12b64d8`, SAP-2089 in [Sapiom#3728](https://github.com/sapiom/Sapiom/pull/3728)

Agreed on the reasoning for `resources[]`, particularly the attribution argument — resources keying on the payment account rather than the definition is what makes "resources in your account this workflow *can* use" the only honest thing an undeclared panel could say, and that is a dodge, not a design. Added as specified.

**Three checks the declaration makes possible**, all verified biting:

```
examples/ failed validation (3 problem(s)):

  - manifest-resource-handle: "hello-agent" template.json /resources/1/handle "demo-db" duplicates /resources/0 — step code looks a resource up by handle, so two resources sharing one collide.
  - manifest-resource-seed: "hello-agent" /resources/0/seed points at "seed.sql", which does not exist in the example directory.
  - setup-provisions: "hello-agent" registry setup.provisions is [inbox, postgres] but the manifest's resources[] derive [postgres, sandbox] — provisions is generated from the manifest, never hand-maintained.
```

The seed-file check is the one I'd flag as load-bearing: a declared seed that isn't on disk provisions an empty database and reports success, which is precisely the failure the field exists to prevent.

`setup.provisions[]` is now **derived and verified** rather than author-declared, so your `capabilities[]` comparison doesn't get to repeat itself. Its schema description says so explicitly.

**On "only seed what a template reads" — your five are exactly right, and I checked rather than took it on faith.** `approval-chain`, `cold-outreach-engine`, `error-triage-digest`, `meeting-notes-crm` and `the-brain` are the five with `CREATE TABLE` + `INSERT` and no read of data they didn't write. That leaves four read-side: `nl-db-query-endpoint`, `scheduled-db-insight-report`, `personalized-media-at-scale`, `durable-backfill`. `AUTHORING.md` now has a two-row table keyed on *what the table is for* — input you read vs. output you write — and names both sets, rather than saying "add a seed". It also states the sample-mode boundary in the same breath, since that is the misread the field invites.

**Populating actual seeds is not in this PR** — that is per-template SQL against four different shapes, and it belongs with the migration in SAP-2078. This PR is the declaration and the gates. Flag it if you want the four `.sql` files here instead.

### SAP-2089

Done rather than deferred: [Sapiom#3728](https://github.com/sapiom/Sapiom/pull/3728). `toDetail` loads the manifest once and projects both fields from it, which meant caching the parsed manifest instead of the projected DTO so `whatItDoes` doesn't also have to ride on `TemplateManifestDto`. `TemplateDetailDto` is unchanged, so there is no frontend diff. 39 tests pass, 3 new — including a manifest that is absent entirely, which must degrade to `null` rather than resurrect a registry value.

**Merge order:** #3728 first (or together). This PR is safe to sit — the pinned ref means nothing goes empty until the pin advances.

### Verification here

`pnpm build && pnpm typecheck && pnpm lint` exit 0 · `pnpm examples:check` clean · `pnpm examples:check:test` **63 pass** (54 mine + 9 from the complexity work merged in as #427).

ratataque (approved):

PR #431, opened 2026-07-28
feat(harness): keep session records past event retention (SAP-2060)

gwitwer:
Closes [SAP-2060](https://linear.app/sapiom/issue/SAP-2060/h4-keep-session-records-past-event-retention) — H4 of Epic H (SAP-2056).

## Why

H2 made past sessions readable by folding them out of `events.ndjson`. That file is an *analytics sink* with an analytics sink's retention — 50 MB / 30 days, truncated oldest-first (`collector/store-retention.ts`) — so session history was built on bytes that get deleted on a schedule. A session a user can read today quietly stops existing next month, with no signal, and no way to tell "never recorded" from "swept".

## What changed

**The archive** — `src/core/record-archive.ts`, `~/.sapiom/harness/records/<harnessSessionId>.json`, one file per *conversation* (a resumed session folds into the primary id's record, same as `read` does).

- **Compaction is lossy on purpose, and says so.** Tool inputs and results are clipped to 512 chars — they are most of the bytes (`tool.call` was ~80% of observed event volume) and the least useful part of a months-old record. Prompts and assistant text are clipped only at the collector's own 4k field cap, so in practice the conversation is kept whole. Over ~64 KB it drops whole turns oldest-first, **never below one**. Both losses are new `SessionRecord.limitations` codes (`compacted-archive`, `dropped-early-turns`) and the transcript view renders them as prose, next to a note naming the record as an archived copy (`SessionRecord.archivedAt`).
- **Bounded, not just durable:** 16 MB total and 365 days, evicted oldest-first, swept after every write (the only thing that grows the store) and once at boot. Writes are a temp + rename and serialized through a queue — both triggers below can fire for one session in the same millisecond.
- **`turnCount` keeps the conversation's count** even when turns were dropped, so a history row's count doesn't change when its events are swept, and the row can't contradict the record it opens. `turnCounts()` falls back to the archive for conversations the log no longer holds.

**Two write triggers, because one isn't enough.** `session.end` reaching the store covers the normal exit (via a new `onEventPersisted` ingest seam — fired from the existing `onNormalizedEvent` it would race the append and store a record with `endedAt: null`); the session's transition to `exited` covers a killed pty that never produced that hook. A **boot pass** archives conversations the log still holds but the archive doesn't, which is what keeps history predating this feature — and anything lost to a force-kill — from vanishing at its 30-day mark. It's capped per boot and logs what it deferred.

## Two decisions worth a reviewer's attention

**1. Reads prefer whichever source still holds the WHOLE conversation — not the archive unconditionally.** The ticket says "the reader prefers it when present and falls back to scanning events". Taken literally that is wrong in two directions: it serves a compacted excerpt for a session that ended a minute ago (when the full events are right there, and H2's byte-offset index makes reading them a ~23 ms hot p95, not the expensive thing it once was), and it freezes a **resumed** conversation's record at its first exit, hiding every later turn. Because retention truncates oldest-first, coverage is always a suffix — so an intact *first* event proves the log lost nothing:

| log state | source | why |
|---|---|---|
| holds the beginning | events | strictly richer, uncompacted |
| has events newer than the archive | events | the conversation continued; the archive is stale |
| beginning swept | archive | only source with the whole conversation |
| nothing | archive | ditto |
| neither | `null` → 404 | "swept" and "never recorded" stay distinguishable |

`readFromEvents` is exposed separately and is what the archiver uses — re-archiving whatever `read` returned would compact an excerpt of an excerpt and re-stamp it as current.

**2. Records are NOT under `<generated>/<sessionId>/`, which the ticket recommended as "already covered by the generated-config retention sweep".** That directory is deleted the moment a session's pty exits (`removeGeneratedSessionDir`, wired to the `exited` transition) and whatever survives that is swept 7 days after going stale. Both lifetimes are *shorter* than the 30 days of events the archive exists to outlive, so a record written there would die sooner than the log it was meant to survive. It gets its own root and its own policy instead.

Also: `truncateForPayload`'s marker is now exported as `PAYLOAD_TRUNCATION_MARKER` (the fold had a private copy of the regex), and the archive's clip *unwraps* an existing marker and folds its count into the new one — so the collector's already-truncated 16 KB results get clipped rather than waved through, and clipping is idempotent instead of shaving another ~23 characters off the text on every re-archive.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm lint` — green across the workspace (0 errors; the one harness warning is the pre-existing unused import at `rest.test.ts:7`).
- Harness unit suite: **1397 passing**, 94 files, plus the 4 perf tests (`cold 131ms · hot p95 23.8ms` unchanged — the archive lookup is one `stat`/read, and the perf reader has none wired).
- New coverage: compaction (clipping, marker folding, idempotency, oldest-first turn drops keeping their original ordinals, the never-drop-the-last-turn floor, the cap measured against what comes *out*), the store (round-trip by primary / merged / agent id, replace-not-accumulate, refusing an unsafe session id, corrupt and foreign files read as absent, another process's write picked up via the dir mtime, both sweep caps, and concurrent write+write+sweep leaving one file and no temp litter), the source-selection rule in all five states above, `turnCounts` after a sweep, `conversationIds`, and the backfill (skips live and already-archived, honors its cap and reports the remainder).
- **Wiring test** (`src/server/record-archive-wiring.test.ts`) boots a real server and proves the acceptance criterion end to end: hook events → archived on `SessionEnd` with a clipped tool input → `events.ndjson` emptied → `GET /api/sessions/:id/record` still 200 by both harness and agent session id, while an unknown id is still 404. Plus the killed-pty path (archived with `incomplete-final-turn`, no `SessionEnd` involved) and the boot backfill of a pre-existing log.
- Playwright: **260 passing**, including a new spec for the archived row — renders 2 turns of a 9-turn session and states all three of "archived copy", "shortened", "earlier ones are gone". Run with `E2E_PORT=<free port>` when several worktrees are live.

## Note on repo

Like H1–H3, the orchestrator metadata said `sapiom/Sapiom`, but every path this ticket names (`src/core/collector/store-retention.ts`, `src/core/inject/retention.ts`) is in `sapiom-js`'s `packages/harness`, so the work landed here. Based on `main`, not on H3's stack branch (`feat/harness-add-workspace-doors`, PR #425) — this depends only on H2, which is on `main`.

https://claude.ai/code/session_01BeLBRx8dsFg5tDcNgt5CUi


gwitwer:
**Rebased onto `feat/harness-add-workspace-doors` and retargeted.**

H3 (SAP-2059, #430) merged into `feat/harness-add-workspace-doors`, not `main` — so `main` does not contain it, and this branch was cut from `main`. The green CI on the previous head was green against a base without H3, which hid a real collision: SAP-2060 `blockedBy` SAP-2059 was set on the tickets precisely because both touch `session-record.ts`, and the branch was cut from `main` anyway.

Rebasing surfaced conflicts in `src/server/index.ts` and `src/shared/types.ts`, and then **four type errors that neither branch sees alone**:

| Break | Cause |
|---|---|
| `resume-brief.ts:268` — `Record<SessionRecordLimitation, string>` missing 2 keys | H4 adds `compacted-archive` + `dropped-early-turns`; H3's exhaustive prose map predates them |
| `rehydration.test.ts`, `resume-brief.test.ts`, `rolling-summary.test.ts` fixtures | H4 makes `SessionRecord.archivedAt` required; H3's fixtures don't set it |

Resolutions:

- **Conflicts** — both sides kept. `statePaths` gains H4's `records` entry while keeping the doors branch's corrected `sampleProject` comment (H4's copy still described the removed `POST /api/sample-project` route). `server/index.ts` keeps H3's rehydration/rolling-summary imports alongside H4's record-archive ones, and H3's note on why `eventStore` is declared above the launch-opts builder.
- **`resume-brief.ts`** — added agent-facing prose for both new limitations. This is the substantive one: a brief built from a compacted archive is *more* lossy than one built from live events, and the brief has to say so. Wording is aimed at the agent reading it ("re-read the file or re-run the command rather than trusting it"), matching the file's existing voice rather than `session-record-view.ts`'s human-facing copy.
- **Fixtures** — `archivedAt: null` (folded from the live event log, not the archive). H4's own archived fixture at `mock-data.ts:805` already demos the non-null state, so the H3 mock stays a live fold like its siblings.

Verified on the combined tree — the first time H3 and H4 have been compiled or run together:

```
pnpm typecheck   exit 0
vitest run       101 files, 1550 tests passed
pnpm lint        0 errors (1 pre-existing warning in rest.test.ts)
pnpm build       exit 0
```

PR #436, opened 2026-07-29
fix(tools): support synchronous video responses

oramada:
## Summary

Make `@sapiom/tools` content-generation video calls handle both synchronous provider results and valid asynchronous handles that expose only a status URL.

## Changes

- Return synchronous `{ video }` results without entering the polling loop.
- Derive the canonical result endpoint from a `status_url` when `response_url` is absent.
- Preserve existing queued-video launch, wait, and resume behavior.
- Add regression coverage and a patch changeset for `@sapiom/tools`.

## Root cause

The SDK assumed every video operation returned an async queue handle with `response_url`. FAL ffmpeg merge may return the completed video synchronously, and some queue handles expose only `status_url`.

## Testing

- [x] Content-generation SDK suite: 55/55 tests
- [x] `@sapiom/tools` typecheck
- [x] `@sapiom/tools` lint
- [x] Changeset validation
- [x] `git diff --check`

## Related

- Refs: SAP-2143

## Checklist

- [x] Patch release metadata included
- [x] No unrelated template or deployment changes
- [x] Self-reviewed

### DIFF
--- a/.changeset/calm-videos-return.md
+++ b/.changeset/calm-videos-return.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": patch
+---
+
+Accept synchronous video responses and status-URL-only asynchronous handles from content-generation providers.
--- a/packages/tools/src/content-generation/index.spec.ts
+++ b/packages/tools/src/content-generation/index.spec.ts
@@ -273,6 +273,57 @@ describe("createClient().contentGeneration.images.create", () => {
 // ---------------------------------------------------------------------------
 
 describe("contentGeneration.video.create()", () => {
+  it("returns a synchronous video result without trying to poll", async () => {
+    const { transport, calls } = makeTransport([
+      (c) =>
+        c.init.method === "POST"
+          ? jsonResponse({
+              video: {
+                url: "https://media/merged.mp4",
+                content_type: "video/mp4",
+              },
+            })
+          : null,
+    ]);
+
+    const out = await createVideo({ prompt: "merge" }, transport, BASE);
+
+    expect(out).toEqual({
+      video: {
+        url: "https://media/merged.mp4",
+        contentType: "video/mp4",
+      },
+    });
+    expect(calls).toHaveLength(1);
+  });
+
+  it("derives the result URL when Fal returns only status_url", async () => {
+    const { transport, calls } = makeTransport([
+      (c) =>
+        c.init.method === "POST"
+          ? jsonResponse({
+              request_id: "req-status-only",
+              status_url: `${BASE}/queue/fal-ai/ffmpeg-api/requests/req-status-only/status`,
+            })
+          : null,
+      (c) =>
+        c.init.method === "GET"
+          ? jsonResponse({ video: { url: "https://media/merged.mp4" } })
+          : null,
+    ]);
+
+    const out = await createVideo(
+      { prompt: "merge", pollIntervalMs: 1 },
+      transport,
+      BASE,
+    );
+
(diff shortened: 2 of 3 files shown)


gwitwer:
Reviewed against the branch source, plus the `@sapiom/fetch` and `agent-core` internals it interacts with. CI is green. Tight change, good tests — one structural gap that I think is worth closing here rather than working around downstream.

## 1. `launchVideo` still throws on a synchronous response — the title over-promises

`createVideo` got the sync branch; `launchVideo` did not. It reads the handle and immediately demands a poll URL (`packages/tools/src/content-generation/index.ts:549-552`), so any operation that completes synchronously via `launch()` throws **"Video submit did not return a result URL to poll" despite a successful generation** — the exact misleading error this PR set out to remove.

This is why #437 had to rewrite `scene-to-video`'s `stitch` step to use `create()`, and that rewrite trades a durable pause for a blocking 5-minute in-step poll (detail in my review over there). Closing the gap here would let that template keep `pauseUntilSignal`:

```ts
const submitted = (await submitRes.json()) as QueueHandle & RawVideoResult;
if (submitted.video?.url) {
  const result = mapVideoResult(submitted);
  return { requestId: submitted.request_id ?? "sync", wait: async () => result, /* dispatch? */ };
}
```

The open question is `dispatch` — a sync result has no correlation id to pause on, so callers doing `pauseUntilSignal(handle)` would need to handle an already-resolved handle. If that's more surgery than you want in a patch, fine — but then please say in the PR body that `launch()` is async-only, because the title reads as though both paths are covered.

## 2. Sync detection keys on `video.url` alone

`if (submitted.video?.url)`. `RawMedia.url` looks required today so this works now. But once the gateway persistence fix (sapiom/Sapiom#3818) starts annotating sync results, a `video` carrying only `download_url`/`file_id` falls through to the queue branch and throws that same misleading error. Suggest gating on `submitted.video` presence, or `url || download_url || file_id`.

## 3. Queue plumbing leaks into the public result

`mapVideoResult(submitted)` spreads `...rest`, so `request_id` / `status_url` / `response_url` land in the returned `VideoGenerationResult` for any response carrying both a video and handle fields. It typechecks via the index signature (`:311`), and today's sync responses probably don't carry both — but stripping the three handle keys before mapping is a one-liner and keeps queue internals out of a user-facing object.

## 4. `queueResultUrl` edge cases

Both cheap to harden:

- A `status_url` ending in `/status/` (trailing slash) fails `endsWith("/status")` → returns `undefined` → throws. `replace(/\/status\/?$/, "")` covers it.
- Query params on `status_url` are carried onto the derived result endpoint. If FAL ever appends anything there, it rides along to a URL that didn't ask for it. `url.search = ""` if that's not intended.

## 5. Tests

The additions are the right shape — asserting `calls[1]!.url` is exactly the check that makes finding 4 verifiable. Two gaps:

- No coverage of the `/status`-suffix-absent path (`queueResultUrl` → `undefined` → throw). That's the branch most likely to regress if someone "simplifies" the string handling.
- The sync-path assertion only checks `url` + `contentType`. Since the whole point of the storage annotation work is `fileId`/`downloadUrl`/`storageError`, worth asserting the sync path preserves them — that's the field set finding 2 is about.

---

Nothing here is a blocker on correctness for the case you tested. My only real ask is that (1) either gets fixed or gets named in the PR body, since #437 is currently absorbing the cost of it.

gwitwer (changes requested):
see review

gwitwer (approved):

PR #437, opened 2026-07-29
fix(examples): repair workflow template reliability

oramada:
## Summary

Repair the seven workflow templates that exposed provider, sandbox, deployment, metering, or tenant-isolation defects during live cost-distribution testing.

### Media

- Retry transient Content Repurposing image failures with bounded exponential backoff.
- Validate every Scene-to-Video clip output before advancing.
- Normalize Kling durations to 5 or 10 seconds.
- Bypass FAL merge for a valid one-clip scene and use the corrected SDK create path for multi-clip merges.

### Sandboxes

- Constrain Durable Backfill sandbox names to Blaxel's 49-character limit and make retry names unique.
- Give Proposal Generator retries unique sandboxes, use an explicit render directory, and retain useful stdout/stderr diagnostics.

### Orchestration

- Replace Approval Chain's missing runtime `pg` driver with bundled `postgres`.
- Route Slack Notifier calls through metered, fail-closed `@sapiom/fetch`.
- Tenant-scope The Brain's definition cache and support explicit child definition IDs.

## Dependencies

Merge and release the SDK contract before deploying Scene-to-Video:

- https://github.com/sapiom/sapiom-js/pull/436

Deploy the gateway persistence fix before expecting synchronous merged videos to receive durable file metadata:

- https://github.com/sapiom/Sapiom/pull/3818

## Testing

- [x] All 7 examples installed independently against their declared versions
- [x] All 7 example typechecks
- [x] All 7 formatting checks
- [x] `git diff --check`

## Related

- Refs: SAP-2144

## Checklist

- [x] No environment-specific `sapiom.json` files
- [x] No experiment logs or pricing exports
- [x] No agent-route change
- [x] Self-reviewed

### DIFF
--- a/examples/approval-chain/AGENTS.md
+++ b/examples/approval-chain/AGENTS.md
@@ -37,8 +37,9 @@ went silent). Inside a step's `run`, Sapiom capabilities are pre-auth'd on
 The canonical state is `ctx.shared`. The `database` capability is an **optional,
 best-effort** durable copy: `recordTransition` appends every transition to
 `ctx.shared.trail` and, when a `ledgerHandle` is configured and it's not a
-`dryRun`, also to a Postgres `approval_chain_ledger` table (via a `pg` client on
-the provisioned connection string). A missing handle, a `dryRun`, or any DB error
+`dryRun`, also to a Postgres `approval_chain_ledger` table (via the ESM-safe
+`postgres` client on the provisioned connection string). TLS behavior follows
+the connection string's `sslmode`; a missing handle, a `dryRun`, or any DB error
 degrades to shared + logs and never fails the chain. Rows are keyed on
 `(execution_id, seq)` with `ON CONFLICT DO NOTHING`, so a step retry is idempotent.
 
--- a/examples/approval-chain/index.ts
+++ b/examples/approval-chain/index.ts
@@ -290,10 +290,12 @@ async function persistTransition(
     });
     return;
   }
-  // `postgres` is ESM-friendly and is included in the self-contained workflow
-  // artifact. A hidden runtime import of `pg` built successfully but left the
-  // driver out of the deployed bundle, so live ledger writes always failed.
-  const sql = postgres(connectionString, { ssl: "require", max: 1 });
+  // Both drivers are externalized into the deploy manifest. `postgres` works
+  // here because its ESM default export is stable; `pg` builds `module.exports`
+  // dynamically, so the previous named ESM import failed at runtime.
+  // Leave TLS policy to LEDGER_DATABASE_URL (`sslmode=...`) so local and
+  // in-network non-TLS databases keep working.
+  const sql = postgres(connectionString, { max: 1 });
   try {
     await sql.unsafe(
       `CREATE TABLE IF NOT EXISTS ${LEDGER_TABLE} (
--- a/examples/approval-chain/index.ts
+++ b/examples/approval-chain/index.ts
@@ -6,7 +6,7 @@ import {
   terminate,
   type AgentExecutionContext,
 } from "@sapiom/agent";
-import { Client } from "pg";
+import postgres from "postgres";
 
 /**
  * Multi-Party Approval Chain (Saga) — a durable, sequential sign-off flow.
@@ -290,10 +290,12 @@ async function persistTransition(
     });
     return;
   }
-  const client = new Client({ connectionString });
-  await client.connect();
+  // `postgres` is ESM-friendly and is included in the self-contained workflow
+  // artifact. A hidden runtime import of `pg` built successfully but left the
+  // driver out of the deployed bundle, so live ledger writes always failed.
+  const sql = postgres(connectionString, { ssl: "require", max: 1 });
   try {
-    await client.query(
+    await sql.unsafe(
       `CREATE TABLE IF NOT EXISTS ${LEDGER_TABLE} (
          execution_id text NOT NULL,
          seq          integer NOT NULL,
@@ -306,7 +308,7 @@ async function persistTransition(
          PRIMARY KEY (execution_id, seq)
        )`,
     );
-    await client.query(
+    await sql.unsafe(
       `INSERT INTO ${LEDGER_TABLE}
          (execution_id, seq, subject, gate_index, approver_id, phase, note)
        VALUES ($1, $2, $3, $4, $5, $6, $7)
@@ -322,7 +324,7 @@ async function persistTransition(
       ],
     );
   } finally {
-    await client.end();
+    await sql.end({ timeout: 5 });
   }
 }
 
--- a/examples/approval-chain/package.json
+++ b/examples/approval-chain/package.json
@@ -12,11 +12,10 @@
   "dependencies": {
     "@sapiom/agent": "^0.6.0",
     "@sapiom/tools": "^0.20.0",
-    "pg": "^8.11.5"
+    "postgres": "^3.4.5"
   },
   "devDependencies": {
     "@types/node": "^20.11.30",
-    "@types/pg": "^8.11.5",
     "typescript": "^5.4.2"
   }
 }
--- a/examples/content-repurposing-pipeline/index.test.mjs
+++ b/examples/content-repurposing-pipeline/index.test.mjs
@@ -16,8 +16,15 @@ test("does not treat unrelated numbers as transient statuses", () => {
   assert.equal(isTransientImageError(new Error("failed after 500 ms")), false);
   assert.equal(isTransientImageError(new Error("request id 429")), false);
   assert.equal(isTransientImageError(new Error("HTTP 503 upstream")), true);
+  assert.equal(isTransientImageError(new Error("status code = 429")), true);
+  assert.equal(isTransientImageError(new Error("nothttp 503")), false);
   assert.equal(
     isTransientImageError(new Error("temporary network error")),
     true,
   );
+});
+
+test("handles long repetitive error messages without regex backtracking", () => {
+  const message = `status${" ".repeat(100_000)}not-a-status`;
+  assert.equal(isTransientImageError(new Error(message)), false);
 });
--- a/examples/content-repurposing-pipeline/index.test.mjs
+++ b/examples/content-repurposing-pipeline/index.test.mjs
@@ -1,0 +1,23 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { isTransientImageError } from "./index.ts";
+
+test("uses structured image error status when available", () => {
+  assert.equal(isTransientImageError({ status: 503 }), true);
+  assert.equal(isTransientImageError({ response: { status: "429" } }), true);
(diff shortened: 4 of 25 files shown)


gwitwer:
Reviewed against the branch source, `packages/agent-core/src/bundle.ts`, `packages/fetch/src/interceptors.ts`, and the Sapiom backend's scope-header handling. CI is green.

These are real bugs and the diagnostic work behind them is good — several of these could only have been found by actually running the templates. My concerns are that a few fixes reach further than the PR describes, one has a correctness bug, and `registry.json` wasn't brought along.

Ranked.

## 1. `stitch` trades a durable pause for a blocking 5-minute poll

`pause: { signal: VIDEO_RESULT_SIGNAL, resumeStep: "finalize" }` → a synchronous `create()`. Two consequences:

- `createVideo`'s inline poll loop caps at `DEFAULT_VIDEO_TIMEOUT_MS` = **5 minutes** (`packages/tools/src/content-generation/index.ts:258`), and `stitch` declares no `timeoutMs`. A 6-shot × 10s merge that exceeds 5 minutes now throws, and the step retries the *whole merge* — re-billed.
- A runner restart mid-poll loses the in-flight merge. That's precisely what the durable pause protected against.

If FAL's merge is reliably sub-5-minute for a 6×10s scene, this is fine — but the PR should say so, because the current comment (`// The SDK handles both synchronous Fal merge responses and queued jobs.`) is true yet glosses over the fact that the *queued* case is now a blocking in-step poll rather than a durable pause. Otherwise: keep the pause and close the `launch()` gap in #436, or at minimum pass an explicit generous `timeoutMs` with a note on why.

## 2. `registry.json` wasn't updated and now contradicts the code

- `scene-to-video.steps[stitch]` is still `kind: "pause"`, described as *"Join the clips into one video with a merge job, then pause until it's ready."* It no longer pauses. That declaration drives the gallery graph glyph and the derived complexity band — it is exactly the drift the divergence check in #427 was built to catch.
- `slack-notifier` declares `capabilities: []` with `post: kind "compute"`, while the code now routes through metered `@sapiom/fetch`. The template's entire teaching pitch is "Slack has no Sapiom capability namespace."

`pnpm examples:check` passes because it can't see step-kind-vs-code — that's the gap, not a green light. Worth updating both entries in this PR.

## 3. Single-clip bypass pairs the wrong clip's `fileId`, and silently truncates the video

`videoUrls` is a **filtered** list, but the `length === 1` branch then reads `clips[0]`:

```ts
if (videoUrls.length === 1) {
  const only = clips[0];
  return goto("finalize", { outputs: [{ ...(only.fileId !== undefined && { fileId: only.fileId }), downloadUrl: videoUrls[0] }] });
}
```

Three shots where only shot 3 resolved a URL → `videoUrls.length === 1` → this returns `clips[0].fileId` (a **different** clip) paired with shot 3's `downloadUrl`. And more importantly, a 3-shot run silently terminates with a 1-clip video and reports success.

Suggest keeping clip identity through the filter (map to `{ clip, url }` pairs, then filter), bypassing only when `clips.length === 1`, and failing loudly when `clips.length > 1 && usable === 1`. You added exactly that instinct in `collect` — this branch should match it.

## 4. Webhook mode now sends the Slack webhook secret to Core

`@sapiom/fetch` puts the full request URL into the authorization payload — `packages/fetch/src/interceptors.ts:214` sends both `url` and `urlParsed`, and only **headers** are sanitized against `sensitiveHeaders`. A Slack incoming-webhook URL *is* the credential; there is no separate token.

Bot mode is unaffected (token in an `Authorization` header, public `chat.postMessage` URL). So the narrow fix is to keep `createFetch` for bot mode and bypass it for webhook mode — the package already supports this:

```ts
const req = new Request(url, init);
(req as any).__sapiom = { enabled: false };
```

Alternatively, confirm with whoever owns Core's authorization records that `hooks.slack.com` paths are redacted before storage. Either is fine; silently persisting webhook URLs is not.

## 5. `slack-notifier` now hard-requires `SAPIOM_API_KEY`

`failureMode: "closed"` + `apiKey: process.env.SAPIOM_API_KEY`. If that env var isn't injected into the step, every send now fails where plain `fetch` used to work. Fail-closed is the right default for a metered call — the issue is that this is the "bring your own API" onboarding template, and its description still tells users the Slack credential is the only one they need. Worth confirming the engine always injects it, and saying so in the comment next to `createFetch`.

## 6. `approval-chain`: the comment's stated mechanism is wrong

> `postgres` is ESM-friendly and is included in the self-contained workflow artifact. A hidden runtime import of `pg` built successfully but left the driver out of the deployed bundle

`packages/agent-core/src/bundle.ts:62` sets `packages: 'external'`, and `:76` pins each external to the author's installed version for the deploy manifest. `postgres` is externalized and server-installed *exactly* like `pg` was — the artifact contains neither.

The fix is almost certainly still correct, but for the other reason in that sentence: `pg` assembles `module.exports` dynamically, so `import { Client } from "pg"` in an ESM artifact throws `SyntaxError: Named export 'Client' not found` at runtime. That matches "built successfully but live ledger writes always failed" far better than a missing dependency would. Worth correcting, or the next person to hit a driver problem will go looking for a bundler allowlist that doesn't exist.

## 7. `approval-chain`: `ssl: "require"` is a new undocumented connection requirement

`pg` with a bare connection string did not force TLS; `postgres(connectionString, { ssl: "require" })` does. Any `LEDGER_DATABASE_URL` pointing at a non-TLS Postgres — local, or an in-VPC instance — now fails where it previously worked. Either respect `sslmode` from the URL, or document the requirement on the env var.

## 8. `content-repurposing`: the transient-error test matches any number in the message

```ts
/\b(408|425|429|500|502|503|504)\b/.test(String(error))
```

This fires on `"failed after 500 ms"`, a request id containing `429`, or an echoed prompt with `500` in it — and `String(error)` throws away the structured status that's probably already on the error. Suggest reading `error.status` / `error.response?.status` first and keeping the string regex as a fallback. The backoff shape itself (bounded, 3 attempts, exponential, logged) looks right.

## 9. `durable-backfill`: the slice can re-introduce a double hyphen, plus an undeclared `tier`

`(base || "backfill").slice(0, 49 - suffix.length)` can land on a `-`, producing `...--a0` — undoing the `.replace(/-+/g, "-")` normalization two lines above. `.replace(/-+$/, "")` after the slice.

Separately, `tier: "s"` is a cost/behavior change that isn't in the PR summary or the ticket scope. If `s` isn't today's platform default, this moves the cost curve in the opposite direction from what the comment intends ("cannot silently change the workflow's cost curve") — worth stating what the current default is. Also the function's doc comment still says "2–63 chars," now contradicted by the new 49-char comment directly below it.

## 10. `proposal-generator`: the render payload now travels on a shell command line

Swapping `writeFile` for `printf '%s' '<base64>' | base64 -d` puts `proposalJson` — which embeds the LLM-generated `draft` — plus the ~3.7 KB script into a single `exec` string. Base64 can't break the single-quoting, so this is safe as written. But the command length now scales with draft size, and the exec channel likely has a lower limit than ARG_MAX. A length guard, or `writeFile` with an absolute path into `renderDir` (your comment implies the file API may accept one), would remove the ceiling.

The wider diagnostics are a clear improvement. Just note the thrown error now embeds 1 KB of stderr, which lands in execution records — fine for npm/node output, worth knowing.

## 11. `the-brain`: the `no-tenant-id` guard runs before the explicit-`definitionId` check

An operator who supplies `member.definitionId` needs no tenant resolution at all, yet a null `ctx.tenantId` still skips the launch with `no-tenant-id`. Moving the guard below the `explicitDefinitionId ?? (await resolveDefId(...))` line makes the explicit path work regardless — which is presumably the point of adding it.

`Sapiom-Scope: tenant:<uuid>` is the right header (backend resolves it at priority 1). Two notes: it's UUID-validated server-side and 400s otherwise, and `resolveDefId`'s `try/catch` swallows that — so a malformed tenant id degrades silently to `child-not-deployed` rather than surfacing. Logging the non-ok status would make that debuggable. Also, header casing is inconsistent between the two call sites' additions (`Sapiom-Scope` vs `x-sapiom-trace-external-id`) — harmless, but `packages/tools/src/_client/index.ts:125` uses the lowercase form for the trace header, so that one at least matches.

While here: this template's `capabilities` declaration (`models.run`, `database.get`, `database.create`) doesn't cover the raw `fetch` calls to the workflows REST API that `launchChild`/`resolveDefId` make. Pre-existing, not yours — but this PR adds headers to those calls, so it may be a good moment to file it.

## 12. Smaller things

- `examples/scene-to-video/index.ts:325` — the comment is now mangled: `// ...pause the step on its result signal. A` / `// The shared aspect ratio keeps...`. The dangling `A` is left from the previous sentence.
- `seed: 42` was removed with no explanation, and the comment that credited it for cross-clip visual consistency was rewritten to drop the mention. If Kling 2.1 Pro rejects `seed`, please say so in the comment — otherwise this reads as a quiet consistency regression alongside an unrelated fix.
- `normalizeClipDuration` rounds **up** at the midpoint: `duration <= 5 ? 5 : 10` sends 6s and 7s to 10s, roughly doubling those clips' cost. Nearest (`< 7.5 ? 5 : 10`) is closer to author intent. The 5-or-10 enum itself is clearly right.
- `stitch`'s merged output sets `downloadUrl: merged.video.downloadUrl ?? merged.video.url` unconditionally while every neighbouring field uses the conditional-spread idiom; on a `fileId`-only result that writes an explicit `undefined`.
- `finalize` logs `hasVideo: Boolean(out?.fileId)` and returns `videoFileId: null` when a sync merge carries no `fileId` — expected until sapiom/Sapiom#3818 lands, but the header comment still promises "the stitched video's durable `videoFileId`." A one-line note there would save someone a confused bug report.

## Version pinning encodes the wrong constraint

`examples/scene-to-video/package.json` bumps to `@sapiom/tools@^0.22.0`. But `packages/tools` is already at `0.22.0` on `main`, and #436 is a **patch** — so the fix this template depends on ships as `0.22.1`. `^0.22.0` resolves to `0.22.0` for anyone installing before #436 releases, and the template breaks with the same "did not return a result URL to poll" error.

The PR body already names the ordering dependency; `^0.22.1` makes it enforceable instead of procedural.

## Process

Seven templates' **runtime behavior** changes here with no automated coverage — the test plan is install + typecheck + format. The two riskiest changes are both pure logic and unit-testable: `stitch`'s control-flow rewrite (finding 3 is the kind of bug a test would have caught immediately) and the metered-fetch swap.

Also, only `scene-to-video` is coupled to #436's release; the other six aren't. Splitting it out would let those land now instead of waiting on an SDK release.

---

To be clear about what I'd actually block on: **2, 3, and 4**. Finding 1 needs a justification rather than a change if merge really is fast. The rest are comment/robustness fixes that can ride along or follow.

gwitwer (changes requested):
see review

oramada:
@claude review

gwitwer (dismissed):

gwitwer (approved):

PR #445, opened 2026-07-30
fix(harness): deploy links or creates the remote agent a template clone never had

ratataque:
## The bug

Apply a gallery template, click **Deploy**, and nothing ships.

`clone()` writes `sapiom.json` with the fork provenance and no `definitionId` (`agent-core/src/clone.ts:202`). That is deliberate — `agent-core/src/config.ts:17` already documents it:

> Absent right after a template `clone` (the definition is created at `deploy`, D6) and filled in by `link`.

**The "created at `deploy`" half was never built.** So `POST /api/workflows/:id/deploy` read the config, got `null`, and answered `409 "workflow is not linked to a Sapiom agent"`. The Deploy button carries no deploy-gate (only an auth gate), so a fresh template offered a live button that could not succeed.

## The fix

The route now resolves-or-creates the remote agent before building, via agent-core's existing `link({ name, create: true })` — which matches an existing definition by **name or slug before creating one**, so this is resync-*or*-create: re-deploying never duplicates, and a template already deployed from another machine re-attaches to the same definition.

```
POST /api/workflows/:id/deploy
  unparseable sapiom.json → 409 (now with a message that says so)
  unlinked → linking → [warning] → building → ready|error
  linked   →                       building → ready|error   (unchanged)
```

- **`linking`** (new, non-terminal) — the agent is being resolved or created.
- **`warning`** (new, non-terminal) — the agent was created but its id could not be written to `sapiom.json`. The build continues; the message says so verbatim.
- Terminal lines are unchanged: exactly one `ready` or `error` per stream.

The agent is named after its own declared `defineAgent({ name })` — read from the canvas extraction cache, which is normally already warm — falling back to `sapiom.json`'s cached name, the registry name, then the folder basename.

Scope is the harness route only. `agent-core` and `cli` are untouched, and `deploy()` still never writes `sapiom.json`. `sapiom agents deploy` keeps its `NOT_LINKED` behavior on purpose: a scripted or CI deploy must not silently create a remote agent because it ran in the wrong directory.

## Four things the review chain caught beyond the original bug

1. **Splitting "unlinked" from "unparseable config."** The old helper collapsed both into `null`. Under auto-link that orphans a resource: we would create the remote agent, then die when `writeConfig` re-read the invalid file. Bad config now 409s before anything is created.
2. **A failed `sapiom.json` write no longer costs you the deploy.** The file is a re-resolvable cache and `link` re-resolves by name, so a read-only checkout or EACCES emits the `warning` line and continues rather than throwing away a successful link.
3. **Double-clicking Deploy could create two remote agents.** `link({ create: true })` is a read-then-write with no uniqueness guard, and the button is not disabled while pending. An in-flight guard now lives at the state layer (following the existing `inFlightHistory` precedent), covered by a Playwright test that was mutation-verified: disabling the guard makes it fail.
4. **A successful first deploy still showed "Draft."** The registry's `list()` never re-reads `sapiom.json`, and the workspace watcher fires only when the *set* of marker directories changes — not on a content edit to an existing file. So the chip stayed **Draft** and Prod Run stayed disabled reading "Not deployed yet" until you opened a new session or restarted. A new `onLinked` seam rescans that one project and broadcasts `workflows.changed`, keeping the registry out of `actions.ts` the same way `resolveDefinitionName` does.

## Test plan

Verified on this tree:

- [x] `pnpm --filter @sapiom/harness test` — **103 files / 1594 tests**, plus perf config 1 file / 4 tests, all green
- [x] `pnpm --filter @sapiom/harness lint` — 0 errors (1 pre-existing warning in an untouched file)
- [x] `pnpm --filter @sapiom/harness typecheck` — clean on both projects
- [x] `pnpm --filter @sapiom/harness build` — exit 0
- [x] Playwright `direct-actions.spec.ts` — 7/7, including the new double-click guard test

New coverage: 8 deploy-route tests (linked, unlinked, bad-config, link failure, key-refresh-and-retry, seam name precedence, a 3-case name-fallback table, the warning path), 4 `onLinked` seam tests, 4 `resolveManifestName` unit tests, and the e2e double-click test.

Worth a manual pass before merge, since no automated test can reach a real tenant:

- [ ] Apply a gallery template into a fresh folder, click **Deploy**, confirm the agent is created, the build runs, and the chip flips to **Deployed** with Prod Run enabled
- [ ] Deploy that same project again — confirm it re-attaches to the same agent rather than creating a second one
- [ ] Confirm the same flow in the packaged desktop app (`pnpm --filter @sapiom/harness-desktop dist`)

## Known limitations, stated deliberately

- **Backend dependency.** `cli/src/commands/agents/link.ts:12` records that the `--create` path (`POST /definitions`) depended on tenant routes added in a parallel effort. If that route is not live, you now get a named failure with a hint instead of an opaque 409 — better diagnostics, but the button will not fully work until it is. **This is the one thing that could not be verified without a real tenant.**
- **Two clones of one template share a remote agent.** `link` matches by name/slug tenant-wide and both clones carry an identical `defineAgent({ name })`, so the second deploy replaces the first's build. Inherent to link-by-name and unchanged from `sapiom agents link --create`, but now reachable from one click. Recorded in the changeset.
- **The double-click guard is per-tab.** Two browser tabs, or a reload mid-deploy, could still race two `link` calls. Desktop is single-window so exposure is small; the real fix would be a server-side map around `ensureDefinitionId`.
- **The `warning` toast is only guaranteed visible at the end.** It is appended to the terminal toast, because the single toast slot is last-write-wins and `building` follows it in the same tick.
- **Out of scope:** `sapiom agents init` writes no `sapiom.json` at all, so a scaffolded (non-gallery) project is never discovered by the scanner and never gets a Deploy button. Separate bug, separate fix. (A `+ Connect`'d project with no marker *does* work end to end — `writeConfig` creates the file.)

Design and plan are committed under `docs/superpowers/`.

### DIFF
--- a/.changeset/deploy-auto-link.md
+++ b/.changeset/deploy-auto-link.md
@@ -1,0 +1,33 @@
+---
+"@sapiom/harness": patch
+---
+
+Deploy now links (or creates) the remote agent for a project that has never
+been linked, instead of failing.
+
+A gallery-template clone lands with `sapiom.json` carrying its fork provenance
+and no `definitionId` — by design, since the definition was always meant to be
+created at deploy. That half was missing, so `POST /api/workflows/:id/deploy`
+answered 409 "workflow is not linked to a Sapiom agent" and the Deploy button
+could not succeed on a fresh template.
+
+The route now resolves-or-creates the agent first (`link({ create: true })`,
+which matches an existing definition by name/slug before creating one, so
+re-deploying never duplicates it), caches the id in `sapiom.json`, and continues
+into the build. The stream gains a non-terminal `linking` line so the UI can say
+what it is doing; terminal lines are unchanged. The agent is named after its
+declared `defineAgent({ name })` where resolvable, falling back to the cached
+name or the workflow's own name. An unparseable `sapiom.json` still 409s — now
+with a message that says so, because creating a remote agent we could not
+record would orphan it.
+
+Caching the newly-linked id in `sapiom.json` is best-effort: the file is a
+re-resolvable cache and `link` re-resolves the same agent by name, so a failed
+write (read-only checkout, a permissions error, the config turning invalid
+between the initial check and this write) must not cost the user their build.
+On that path the stream emits a non-terminal `warning` line instead of failing
+— the agent was created on Sapiom but not recorded locally, so the Deployed
+chip will not flip and the next deploy re-links, but nothing is duplicated
+because `link` resolves the same agent again. The SPA renders both the
+`linking` and `warning` lines, and a double-click on Deploy can no longer
+create two remote agents for the same project.
--- a/.changeset/deploy-auto-link.md
+++ b/.changeset/deploy-auto-link.md
@@ -31,3 +31,9 @@ chip will not flip and the next deploy re-links, but nothing is duplicated
 because `link` resolves the same agent again. The SPA renders both the
 `linking` and `warning` lines, and a double-click on Deploy can no longer
 create two remote agents for the same project.
+
+Because linking matches by name, two gallery clones of the same template —
+which share the same declared `defineAgent({ name })` — resolve to the same
+remote agent, so deploying the second one replaces the first's build; this is
+inherent to link-by-name (unchanged from `sapiom agents link --create`), not
+a new bug, but it's now reachable from a single button click.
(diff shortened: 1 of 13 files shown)


gwitwer (approved):
nice!

PR #456, opened 2026-07-31
feat(examples): add orchestration + browser-automation gallery templates (SAP-2202)

gwitwer:
## What & why

[SAP-2202](https://linear.app/sapiom/issue/SAP-2202) quantified that the onboarding template gallery under-represents Sapiom's differentiated capabilities. Two capabilities had **zero** gallery presence:

- **Durable workflow orchestration** (`ctx.sapiom.agents.run` — "a workflow composes other workflows") — the one existing multi-workflow example (`the-brain`) uses raw child-launches and is being reduced, so there was no clean exemplar of the core differentiator.
- **Browser automation** (`ctx.sapiom.browserAutomation`, new in `@sapiom/tools@0.23.0`) — brand-new capability, no template.

This PR adds one flagship template for each gap.

### `fan-out-and-combine` — Fan Out and Combine
The canonical **workflow-composes-workflow** exemplar: split a goal into parts, run each part as its own child workflow in parallel via `ctx.sapiom.agents.run`, then merge the results (fan out → join → reduce).

- **Composes itself** by default — `childDefinition` defaults to `ctx.agentName`, so a deployed copy dispatches child runs of itself with no other setup. Point it at any other deployed slug to fan that out instead.
- **Bounded recursion** — children run in a `leaf` mode that does one unit of work and never fans out (depth is exactly one).
- **Never-fail** — a child that throws or doesn't complete becomes a `{ ok: false }` row and the reduce runs over the survivors; if nothing comes back it terminates with an honest account rather than a fabricated answer. `dryRun` returns the fan-out plan without dispatching.
- Runnable with `{}`.

### `logged-in-screenshots` — Logged-In Page Screenshots
The gallery's **first browser-automation** template: drive a real hosted browser to (optionally) log in and capture the pages you list as hosted images — the view a signed-in user sees, which a scraper can't reach.

- One `withSession` captures every page in a single session (in-session shots have no per-shot charge; the session always closes in a `finally`).
- The login (`identities.create` + `withSession({ identityId })`) is **optional and honest** — miss the `loginUrl`/`loginUsername`/`BROWSER_LOGIN_PASSWORD`, or let it fail, and it captures the public view and reports `authenticated: false`.
- Every requested URL yields a row (success or failure). Runnable with `{}`.

## Out of scope (deliberate)
SAP-2202 also suggested a flagship coding-agent template and de-emphasizing near-duplicate digest templates. Coding agents already have three gallery templates, and "de-emphasize" is a product/composition decision (and destructive), so both are left for a product call — this PR lands the two genuine zero-coverage gaps.

## Validation
- `pnpm examples:check` — passes (28 templates, schema-valid, sorted, all manifests valid; both new bands match the derived complexity score).
- `tsc --noEmit` — passes for both templates against the **published** SDKs (`@sapiom/agent@^0.7.1`, `@sapiom/tools@^0.23.0`), which is what the one-click build installs.
- Not run: a live deployed run (needs a live tenant) — the offline `run_local` paths and stub behavior were traced from the SDK stub client.

https://claude.ai/code/session_01VJuaJmQAwcKqvXqtoQUCRK

### DIFF
--- a/examples/fan-out-and-combine/AGENTS.md
+++ b/examples/fan-out-and-combine/AGENTS.md
@@ -1,0 +1,42 @@
+# Working in this agent
+
+This project defines exactly one Sapiom agent in `index.ts` — **Fan Out and
+Combine** — authored against `@sapiom/agent`. It splits a goal into parts, runs each
+part as its own child workflow in parallel, then merges the results:
+`plan` → `fanOut` → `reduce` → `done`, with a `solve` leaf and a `planned` (dry-run)
+off-ramp. Inside a step's `run`, Sapiom capabilities are pre-auth'd on `ctx.sapiom`
+(here: `ctx.sapiom.agents.run`, `ctx.sapiom.models.run`).
+
+## Authoring
+
+- An agent is `defineAgent({ entry, steps })`; each step is
+  `defineStep({ name, next, run })`. Keep exactly one `defineAgent(...)` export.
+- **Capabilities come from the types.** What's available on `ctx.sapiom` is defined
+  by `@sapiom/tools` — read the types / use autocomplete rather than guessing.
+- **One agent, two roles, chosen by `mode`.** The default (`coordinate`) fans out;
+  each child is launched with `mode: "leaf"`, and the leaf path (`solve`) does one
+  unit of work and terminates. A leaf NEVER fans out — that is what bounds the
+  recursion to a single level. Don't let `solve` call `agents.run`.
+- **It composes itself.** `fanOut` defaults `childDefinition` to `ctx.agentName`, the
+  run's own slug, so a deployed copy dispatches leaf runs of itself with no other
+  deployment. Point `childDefinition` at another slug to fan that out instead.
+- **`agents.run` blocks; that's the join.** `fanOut` uses `Promise.all` over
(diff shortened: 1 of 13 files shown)


gwitwer:
### Note on the red CI (pre-existing, not from this PR)

All jobs fail at the **`pnpm install --frozen-lockfile`** step, before any template validation runs:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because
pnpm-lock.yaml is not up to date with packages/cli/package.json
  - @sapiom/harness (lockfile: >=0.2.0, manifest: >=0.2.4)
```

This is a stale lockfile already on `main` — `packages/cli/package.json` requires `@sapiom/harness >=0.2.4` while `pnpm-lock.yaml` still pins `>=0.2.0`. It affects every PR's install, not this one: **this branch touches only `examples/`** (two new template dirs + the `registry.json` index) and does not modify `pnpm-lock.yaml` or any `packages/*/package.json`.

The actual gate for this change, `pnpm examples:check`, **passes** (28 templates, schema-valid, sorted, all manifests valid), and both new templates `tsc --noEmit` clean against the published SDKs. Regenerating the workspace lockfile is out of scope for this templates PR; it should be fixed on `main` separately (or this PR rebased once it is).

PR #464, opened 2026-07-31
fix(agent): manifest schema was stripping authored description + capabilities

BastienBoymond:
## The bug

Authored `description` (on `defineStep` / `defineAgent`) and step `capabilities` have **never rendered on the harness canvas** since they were introduced — because manifest validation silently drops them.

`agentManifestSchema` is a `z.object()`, which strips any key it doesn't declare. The introducing PR added `description`/`capabilities` to the **TS types** and to **`buildManifest`'s output**, but not to the **runtime Zod schema**. And `@sapiom/agent-core`'s `check()` — the exact pipeline the canvas uses to build a graph — runs `agentManifestSchema.parse()` on every manifest. So:

```
source authors `description`  →  buildManifest emits it  →  schema.parse() STRIPS it  →  canvas gets ""
```

Discovered while testing "Describe with AI": the agent wrote correct `description:` fields into a workflow's source, the workflow's own SDK read them, but the canvas stayed blank. Running the real extraction (`extractWorkflowGraph`) returned every description as `""`; adding the fields to the schema made all of them flow through.

## The fix

Add `description` (step + workflow) and `capabilities` (step) to `workflowStepManifestSchema` / `agentManifestSchema` as optional fields, so they survive validation and reach consumers.

## Test

New regression in `manifest.spec.ts`: build a manifest with an authored workflow + step `description` and step `capabilities`, `agentManifestSchema.parse()` it, and assert they're **preserved** (the previous schema silently dropped them). The existing suite only ever parsed manifests without these fields, which is how it slipped through.

Verified: `@sapiom/agent` unit **84 pass**. Changeset included (patch → publish).

## Impact / sequencing

This unblocks authored descriptions **and** the "Describe with AI" feature (PR #421), both of which were non-functional end-to-end without it. Needs a **publish** (like the field itself did) before the shipped harness renders authored descriptions.

### DIFF
--- a/.changeset/manifest-schema-description.md
+++ b/.changeset/manifest-schema-description.md
@@ -1,0 +1,9 @@
+---
+"@sapiom/agent": patch
+---
+
+Fix: authored `description` (on `defineStep` and `defineAgent`) and step `capabilities` were silently dropped during manifest validation.
+
+`agentManifestSchema` is a `z.object()`, which strips keys it doesn't declare. The previous release added `description`/`capabilities` to the TS types and to `buildManifest`'s output, but not to the runtime schema — so `agentManifestSchema.parse()` (which `@sapiom/agent-core`'s `check()` runs on every manifest) discarded them. The result: tooling that reads the validated manifest — the harness canvas — always saw empty descriptions, no matter what the source authored.
+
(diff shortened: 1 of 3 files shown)


BastienBoymond:
Folded into #421 (the Describe-with-AI feature is non-functional without it, so they ship together). Same commit, same changeset + regression test.

PR #467, opened 2026-07-31
feat(agent): make the agent input contract first-class via defineAgent({ inputSchema }) (SAP-2228)

gwitwer:
## Summary

Makes the agent input contract a first-class construct instead of a convention. `defineAgent` now accepts an optional agent-level `inputSchema` — **one obvious place** to declare "what this agent takes" — closing the drift described in SAP-2228/SAP-2226 where authors had to *know* that "the entry step's schema is the agent's contract" and hand-mirror it.

## What changed (`packages/agent`)

- **`agent.ts`**
  - `AgentDefinition` gains an optional `inputSchema?: ZodType<TInput>`.
  - `defineAgent` **folds** the agent-level schema onto the entry step when the entry step declares none. Everything downstream (`buildManifest`, `workflowInputContract`, `stepInputContract`, and therefore the engine + dashboard) keeps reading `steps[entry].inputSchema` unchanged — it now simply carries the contract.
  - Declaring a **different** schema at both the agent level and the entry step is a **build error** with a clear message. Declaring the *identical* schema object in both places is allowed (no-op).
  - Typing `inputSchema` as `ZodType<TInput>` infers `TInput` from the schema, tying the `defineAgent<TInput>` generic (hence the `run(def, input)` call site) to the same runtime schema that becomes the contract — the TS annotation and runtime validation can no longer diverge.

## Design note — scope

The manifest **shape is intentionally unchanged**: folding onto the entry step satisfies the acceptance criteria ("manifest whose entry step carries that JSON Schema") without a manifest schema bump, so **no `apps/workflows-engine` change is required** and existing agents keep working byte-for-byte. The two *Nice-to-have* items in the ticket (an explicit manifest-level input contract field to remove the ordinal-vs-entry dependency, and a `template.json` `defaultInput` generator) are deliberately left as follow-ups — each is additive and carries its own cross-repo/tooling surface.

## Acceptance criteria

- ✅ An agent declaring only `defineAgent({ inputSchema })` builds a manifest whose entry step carries that JSON Schema (dashboard renders its fields).
- ✅ A conflicting agent-level + entry-step schema fails the build with a clear message.
- ✅ Existing agents that declare the schema on the entry step keep working unchanged.

## Testing

- `sdk.spec.ts` — fold onto schemaless entry, identical-schema no-op, conflict throws, non-entry steps untouched, back-compat no-op.
- `manifest.spec.ts` — manifest's entry step carries the folded JSON Schema and still validates against `agentManifestSchema`.
- `type-enforcement.spec.ts` — `TInput` is inferred exactly from `inputSchema` (checked by `tsc --noEmit`).
- Verified locally: `pnpm --filter @sapiom/agent build`, `pnpm typecheck`, `pnpm lint`, and the three specs (66 tests pass).

A `minor` changeset is included (`@sapiom/agent`).

Closes SAP-2228

https://claude.ai/code/session_017dbrv98JjCQ1wXbvmpTsR4

### DIFF
--- a/.changeset/agent-level-input-schema.md
+++ b/.changeset/agent-level-input-schema.md
@@ -1,0 +1,11 @@
+---
+"@sapiom/agent": minor
+---
+
+Make the agent input contract first-class: `defineAgent({ inputSchema })`.
+
+`defineAgent` now accepts an optional agent-level `inputSchema` — one obvious place to declare "what this agent takes". When the entry step declares no `inputSchema` of its own, `defineAgent` folds the agent-level schema onto it, so the built manifest's entry step carries the JSON Schema (and the dashboard renders its fields) without any downstream change. Declaring a *different* schema at both the agent level and on the entry step is now a build error with a clear message; declaring the identical schema object in both places is allowed.
+
+The schema is typed `ZodType<TInput>`, so the `defineAgent<TInput>` generic (hence the `run(def, input)` call site) is inferred from the same runtime schema that becomes the contract — the TS annotation and the runtime validation can no longer drift apart (SAP-2226).
+
+Existing agents that declare the schema on the entry step keep working unchanged.
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -1,3 +1,7 @@
+// `zod/v4` subpath (present in zod 3.25.x AND zod 4.x): the v4 `ZodType` surface
+// while the `zod` peer can resolve to v3 or v4. See step.ts / introspection.ts.
+import type { ZodType } from 'zod/v4';
+
 import { UnknownStepError } from './errors.js';
 import type { StepDefinition } from './step.js';
 
@@ -33,6 +37,17 @@ export const AGENT_DEFINITION_BRAND = Symbol.for('sapiom.models.definition');
  *     steps: { gather, summarize, ... },
  *   });
  *
+ * The input contract can instead be declared once, at the agent level, via
+ * `inputSchema` — `TInput` is then inferred from it (no explicit generic
+ * needed) and `defineAgent` folds it onto the entry step:
+ *
+ *   export const cycleAgent = defineAgent({
+ *     name: 'cycle',
+ *     entry: 'gather',
+ *     inputSchema: z.object({ companyId: z.number() }),
+ *     steps: { gather, summarize, ... },
+ *   });
+ *
  * Steps inside `steps` are typed Step<unknown, unknown, TShared>. The
  * primitive doesn't (and can't) statically enforce that each step's TIn
  * matches its predecessor's TOut — that's a design tradeoff for the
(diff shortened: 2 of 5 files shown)


gwitwer:
## Review — one blocking bug, otherwise sound

**Verdict: request changes (1 real bug).** Fold-logic branch coverage, generics/`TInput` inference, and downstream-reader consistency all check out.

### 🔴 Blocking — `defineAgent` mutates the caller's `steps` map in place
`packages/agent/src/agent.ts:184-187`

```ts
(def.steps as Record<string, StepDefinition<TShared>>)[def.entry] = {
  ...entryStep,
  inputSchema: def.inputSchema as ZodType<unknown>,
};
```

`AgentDefinition.steps` is typed `Readonly<Record<...>>` and treated read-only everywhere else in this function. This casts the `Readonly` away and rewrites the entry step on the object the **caller passed in**, not a copy. Consequences:

- Two `defineAgent` calls sharing a `steps` object literal (common step map, reused test fixture) → the first call's fold silently rewrites the second's entry step. Order-dependent, hard to trace.
- An author who `Object.freeze()`s their steps map (reasonable, given the `Readonly` type) gets a runtime `TypeError` instead of the intended graceful fold.

**Fix** — replace `def.steps` with a fresh object instead of mutating the input:

```ts
(def as { steps: Record<string, StepDefinition<TShared>> }).steps = {
  ...def.steps,
  [def.entry]: { ...entryStep, inputSchema: def.inputSchema as ZodType<unknown> },
};
```

Please add a regression test: two `defineAgent` calls from the **same shared `steps` object** (one with agent-level `inputSchema`, one without) → assert the second is unaffected.

### 💬 Nit — conflict detection is reference-identity only
`agent.ts:173` (`entryStep.inputSchema !== def.inputSchema`). By design ("identical schema *object*"), so not a bug — but two structurally-identical schemas built separately (`z.object({...})` twice) are flagged as a conflict. Consider wording the error as "declare it once and import it in both places" so authors don't read it as "the schemas differ."

gwitwer:
Thanks — good catch, fixed in 4d498e5.

**🔴 Blocking (in-place mutation):** switched the fold to copy-on-write, essentially your suggested fix — a fresh `steps` map with a fresh entry step, so the caller's `Readonly` object is never written to:

```ts
(def as { steps: Record<string, StepDefinition<TShared>> }).steps = {
  ...def.steps,
  [def.entry]: { ...entryStep, inputSchema: def.inputSchema as ZodType<unknown> },
};
```

Added two regression tests:
- two `defineAgent` calls from the **same shared `steps` object** (one with agent-level `inputSchema`, one without) → the second is unaffected and the shared object is untouched;
- folding onto an `Object.freeze()`d steps map does **not** throw and leaves the original frozen map as-is.

**💬 Nit (conflict wording):** reworded so it reads as a "declare once" instruction rather than "the schemas differ":

> Agent '<name>' declares a different inputSchema at the agent level and on its entry step '<entry>'. Declare the input contract once and reference that single schema object in both places (or remove one).

I kept reference-identity for the conflict check by design (comparing two structurally-identical zod schemas for equality is unreliable), but the message now makes the "same object" expectation explicit.

Local verification on `@sapiom/agent`: typecheck + lint + build clean, 68 tests pass (incl. the 2 new ones).

PR #468, opened 2026-07-31
fix(examples): approval-chain — drop gate pause timeoutMs so a slow approval isn't reaped (SAP-2223)

gwitwer:
## Problem

`approval-chain` was authored against an engine with **no paused-run reaper** and wired `timeoutMs: reminderMs` (24h default) onto each approval `pauseUntilSignal`, expecting a lapsed pause to **resume** `decide` with an empty payload — a reminder tick.

The engine now **has** a reaper (confirmed while verifying SAP-2206 / PR #458):

- `recorder.service.ts` maps `timeoutMs` → `pausedUntil` on the execution row.
- `workflows-sweep.processor.ts` runs a 60s repeatable sweep → `findTimedOutPaused` → `expirePause`.
- `runner.service.ts` `expirePausedExecution` → `failExecution(PauseTimeoutError)`. The sweep filters only on `status=PAUSED` + past `pausedUntil` — **no pause-kind/signal filter** — so signal pauses with `timeoutMs` are swept and **terminated**.

**Impact:** on the common one-click Approve/Reject path (SAP-948, no external cron), any approval slower than `reminderMs` (24h) **silently failed the run** with `PauseTimeoutError` instead of sending a reminder — and never reached the graceful `escalate` step (a `PauseTimeoutError` fails the run, it does not route to a step).

## Fix (option 1)

- **Drop `timeoutMs`** from both the `present` and `remind` pauses → gates wait **indefinitely at $0**. Reminders/escalation are driven purely by the `approval.decision` signal convention the header already described: the run-detail one-click UI, a cron firing `remind`/`timeout`, or a `run_local` auto-resume.
- **Removed the now-dead `reminderMs` plumbing** (input field, `Shared` slot, `clampReminderMs`, `DEFAULT_REMINDER_MS`) — its only consumer was the pause `timeoutMs`. `maxReminders` (default 2) still bounds reminder ticks before escalation.
- **Corrected the stale "no paused-run reaper exists today" comment** and the matching README / AGENTS / template.json copy.
- **Added a regression note** at both pauses and in the header: a `timeoutMs` here now *terminates* the run, so it must not be re-added without switching to a terminal model. This mirrors `wait-for-webhook` (SAP-2206), which *wants* the terminal timeout and opts in — same engine mechanism, opposite intent.

## Acceptance criteria

- [x] `approval-chain` no longer auto-fails a legitimately-slow approval; the reminder/escalation path behaves as documented.
- [x] Inline comments match live engine behavior (a reaper exists; `timeoutMs` = terminal failure).
- [x] A note captures the intended timeout semantics so a future engine change can't silently re-break it.

## Verification

- `pnpm examples:check` ✓ (registry + 27 manifests schema-valid; approval-chain not flagged)
- `pnpm examples:check:test` ✓ (73/73)
- `prettier --check examples/approval-chain/**/*.ts` ✓
- `tsc --noEmit` (in `examples/approval-chain`) ✓

Closes SAP-2223.


gwitwer:
## Review — fix is correct; flagging a sibling bug it doesn't cover

**Verdict: approve with nits.** Dropping `timeoutMs` from both `pauseUntilSignal` calls in `present`/`remind` means the reaper's `findTimedOutPaused` sweep never matches these pauses, so a slow approval can no longer be silently killed by `PauseTimeoutError`. All reminder/escalation logic in `decide` is driven by signal-payload content and is untouched. Traced every `reminderMs`/`DEFAULT_REMINDER_MS`/`clampReminderMs` reference — no orphaned refs left.

### ⚠️ Follow-up (not this PR) — `cold-outreach-engine` has the exact same footgun, live
`examples/cold-outreach-engine/index.ts:693` — `send` calls `pauseUntilSignal({ signal: REPLY_SIGNAL, resumeStep: "advance", correlationId, timeoutMs: dripMs })`. Same pattern approval-chain just fixed: a signal pause with `timeoutMs`, where `advance` expects an empty-payload resume "on timeout." Under the live reaper this terminates the run with `PauseTimeoutError` on the first drip interval with no reply — silently failing any campaign with >1 touch. The header comment (~line 46, "resumes on its own when the interval passes") makes the same now-false claim. **Recommend a follow-up ticket so this doesn't ship as a second incident.**

### ✅ Verified correct nearby (no action)
- `wait-for-webhook` (`index.ts:150-215`) is the intentional mirror case — its terminal `PauseTimeoutError` on lapse is the *desired* honest terminal, not a silent bypass. Good contrast.
- `scene-to-video:441`, `dependency-upgrade:204,287`, `eval-gate:106` use `timeoutMs` as a step/capability-call timeout, not a `pauseUntilSignal` timeout — different mechanism, unaffected.

### 💬 Nit
No artifact in this repo exercises the fixed behavior end-to-end (the reaper lives in the engine, and `run_local` auto-resumes regardless of `timeoutMs`). Consider a backend-side integration test for `PauseTimeoutError` semantics on signal pauses if one doesn't already exist. Out of scope for this examples PR — not blocking.

PR #472, opened 2026-07-31
feat(examples): examples:sync-setup — populate registry setup blocks from manifest zeroSetup (SAP-2234)

gwitwer:
## What & why

The gallery **shelf reads the `setup` block in `registry.json`, not the manifest `zeroSetup`** — and `registry.json` carried no `setup` blocks at all, so the authored zero-setup truth (from #470 / SAP-2080) never reached the shelf. The `pnpm examples:sync-setup` codemod the schema comment references didn't exist. This adds it, plus a CI guard.

## Changes

- **`deriveSetup(manifest)` + `pnpm examples:sync-setup`** — generates each template's registry `setup` block from its co-located `template.json`:
  - `runsWithNoSetup` — `true` only where the manifest carries a `zeroSetup` with a *meaningful* terminal (`completed` / `completed_partial`); a suspend or an absent `zeroSetup` ⇒ `false`.
  - `connectionCount` ← `requiredSecrets.length`, `settingCount` ← `settings.length`.
  - `provisions` ← distinct sorted resource kinds (reuses `deriveProvisions`), only when non-empty.
  - `degradedWithoutSetup` ← `zeroSetup.narrative` verbatim (renders on the card).
- **`checkSetupSync` wired into `examples:check`** — CI fails when a committed `setup` block drifts from what its manifest implies (verify-don't-trust; supersedes the narrower provisions-only check).
- **Regenerated `registry.json`** setup blocks for all 27 templates.
- **Unit tests** for `deriveSetup` + `checkSetupSync`.

## Scoping notes

- **Stacked on #470** (base `feat/SAP-2080`) because it consumes the manifest `zeroSetup` that PR adds — **merge #470 first**.
- Templates with no authored `zeroSetup` get `runsWithNoSetup: false` and no narrative — correct for `the-brain` (excluded), and honest for the recently-fixed `proposal-generator` / `research-to-microsite` until their `zeroSetup` is authored (follow-up / re-baseline).
- `runsWithNoSetup: true` coexists with `connectionCount ≥ 1` for honest-degrade templates (e.g. `slack-notifier`): it *runs* immediately (composes + prints), and `connectionCount` signals the credential needed to do the real thing — both drive the shelf.

## Validation
- `pnpm examples:check` — 27 templates, schema-valid; drift guard verified to fail on tampering and clear after re-sync ✅
- `pnpm examples:check:test` — 79/79 (6 new) ✅
- `prettier --check` ✅

Unblocks SAP-2063 (shelf labels). Ticket: SAP-2234.

### DIFF
--- a/examples/registry.json
+++ b/examples/registry.json
@@ -63,7 +63,14 @@
           "capability": "email.send",
           "terminal": true
         }
-      ]
+      ],
+      "setup": {
+        "runsWithNoSetup": true,
+        "connectionCount": 0,
+        "settingCount": 2,
+        "provisions": ["postgres"],
+        "degradedWithoutSetup": "Runs the saga with no approvers supplied, so nothing is committed and it escalates."
+      }
     },
     {
       "id": "cold-outreach-engine",
@@ -140,7 +147,14 @@
           "kind": "compute",
           "terminal": true
         }
-      ]
+      ],
+      "setup": {
+        "runsWithNoSetup": true,
+        "connectionCount": 0,
+        "settingCount": 1,
+        "provisions": ["postgres"],
+        "degradedWithoutSetup": "Runs the outreach engine with no leads, so there is nobody to write to and nothing is sent."
+      }
     },
     {
       "id": "content-repurposing-pipeline",
@@ -201,7 +215,13 @@
           "capability": "email.send",
           "terminal": true
         }
-      ]
+      ],
+      "setup": {
+        "runsWithNoSetup": true,
+        "connectionCount": 0,
+        "settingCount": 1,
+        "degradedWithoutSetup": "Repurposes a built-in sample post into a newsletter, tweet thread, video script, and LinkedIn post."
+      }
     },
     {
       "id": "dependency-upgrade",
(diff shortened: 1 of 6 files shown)


gwitwer:
## Review — solid; two small cleanups

**Verdict: approve with nits.** `deriveSetup` mapping is correct: `MEANINGFUL_ZEROSETUP_TERMINALS = {completed, completed_partial}` covers exactly the two non-suspend `terminalState` enum values; `paused_for_approval` and absent `zeroSetup` both correctly fall to `false` (no unhandled enum, no inversion). `connectionCount`/`settingCount`/`provisions`/`degradedWithoutSetup` all derive correctly. `checkSetupSync` does a full canonical key-sorted JSON compare (catches any field-level drift, incl. an absent `setup` block) and is confirmed wired into `.github/workflows/test.yml` via `examples:check`, so it actually guards in CI. Base-branch scoping vs #470 confirmed clean (510/-51 across the 6 PR files, no #470 content leaking in).

### 💬 Nits
1. **Dead code** — `checkSetupProvisions` (+ its ~5 tests in `examples-manifest-check.test.mjs`) is no longer called anywhere; its only call site in `examples-check.mjs` was swapped to `checkSetupSync`, which the PR body says "supersedes the narrower provisions-only check." Looks like missed cleanup — remove in this PR or a fast follow before it lingers as confusing dead code.
2. **Test symmetry** — no direct assertion that `terminalState: "completed"` (only `completed_partial` is tested) → `runsWithNoSetup: true`. Same `Set.has` check, low risk, but it's the other literal the PR names — worth one line.

### Note (not a bug)
`canonical()` sorts object keys but not arrays, so `provisions` order matters in the raw compare. Since `deriveProvisions` always sorts and the sync script always writes the derived block, this only bites if someone hand-edits `provisions` into a different valid order — which *should* fail the drift check anyway (generated, never hand-edited). Just stricter than the old order-insensitive `checkSetupProvisions`.

PR #475, opened 2026-07-31
feat(harness): instrument the Studio with client PostHog + tier consent

gwitwer:
## Summary

Instruments the Studio (`@sapiom/harness`) with client-side PostHog — journeys, click tracking (autocapture), and UI-driven outcome events — mirroring the web app's SAP-2117 analytics into the **same PostHog project (291192)** so studio and web journeys sit side by side. Ships with a privacy-forward, two-tier consent model appropriate to a tool that runs on someone's desktop. **No session/screen recording, ever.**

Motivation: Yash & Oded got stuck on vault/secrets in the Studio (a docs/usability miss, not a bug), and we couldn't see it — the Studio streams to our BigQuery transcript pipeline but emits **nothing to PostHog**, so its UI-interaction layer (clicks, buttons, journeys) is dark. This closes that gap.

### Client PostHog module (`web/src/lib/analytics/`)
- `journeys.ts` — journey is classified from **App view state**, not a URL (the harness has no router); same `Journey` vocabulary as the web app.
- `before-send.ts` — redaction choke point: strips the boot token from URLs, truncates click text, and **drops it entirely on secrets/vault surfaces**. Replay/`$snapshot` logic removed (no recording).
- `events.ts` — typed outcome-event registry + `track()`; journey/view ride as super-properties via `registerViewContext`.
- `tracking-attrs.ts` + `TrackScope.tsx` — declarative autocapture attribution.
- `posthog.ts` — init / identify / org-group / consent-gate. Initializes lazily from the first `AppState` so the consent decision is atomic (no pre-consent capture window). `disable_session_recording`, `persistence: localStorage` for offline buffering.

### Identity + config
- `tenantId` exposed on `AppState` / `GET /api/state`; `identify(userId)` + `group('organization', tenantId)`.
- PostHog key/host injected at **runtime** via `window.__HARNESS__` (`server/static.ts`) — one bundle across CLI + desktop; `SAPIOM_POSTHOG_KEY=""` disables.

### Two-tier consent
- **PostHog (light: clicks/journeys/usage)** — **on by default**, honors `SAPIOM_TELEMETRY_DISABLED` / `DO_NOT_TRACK` / `--no-telemetry`, opt-out in Settings (`productAnalyticsOptIn`).
- **Sapiom→BQ (invasive: session detail)** — **flipped to opt-out by default** (the pre-release TODO), benefit-framed opt-in on the WelcomePanel setup surface + desktop `setup.html`.
- Surfaced in existing surfaces only (SettingsPopover + WelcomePanel + TelemetryNotice) — no new screens.

## Testing
- [x] `pnpm --filter @sapiom/harness typecheck` (server + web) green
- [x] `pnpm --filter @sapiom/harness test` — **1634 unit tests pass** (15 new: journeys mapping + redaction/secrets-drop + boot-token strip)
- [x] lint clean (0 errors)
- [ ] Manual: run the Studio, confirm events land in PostHog 291192 with journey/org group; verify consent gating + offline flush (follow-up — see ticket)

## Related
- Refs: SAP-1988
- Pattern mirrored from SAP-2117 (web journeys); harness telemetry lineage: SAP-1449 / SAP-1442 / SAP-1423.

## Notes / follow-ups (in-ticket)
- Richer named events on deploy/run/secrets controls + broader `trackingAttrs` (autocapture already captures these clicks; this labels them).
- Live PostHog verification pass.
- Org-level grouping only for now (`userId === tenantId` today).

### DIFF
--- a/packages/harness-desktop/src/renderer/setup.html
+++ b/packages/harness-desktop/src/renderer/setup.html
@@ -44,8 +44,8 @@
 
       <section id="consent" class="consent" hidden>
         <label class="consent-row">
-          <input type="checkbox" id="telemetry" checked />
-          <span>Share anonymous usage data to help improve Sapiom (optional).</span>
+          <input type="checkbox" id="telemetry" />
+          <span>Share your session details with Sapiom so we can optimize your experience (optional, off by default).</span>
         </label>
         <button id="consent-continue" class="btn-primary">Continue</button>
       </section>
--- a/packages/harness/package.json
+++ b/packages/harness/package.json
@@ -88,6 +88,7 @@
     "@xterm/addon-web-links": "^0.11.0",
     "@xterm/xterm": "^5.5.0",
     "lucide-react": "^1.23.0",
+    "posthog-js": "^1.363.5",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "tsx": "^4.19.0",
--- a/packages/harness/src/cli/consent.test.ts
+++ b/packages/harness/src/cli/consent.test.ts
@@ -57,17 +57,19 @@ describe("ensureConsent", () => {
     expect(settings.telemetryOptIn).toBe(false);
   });
 
-  it("defaults to ON and persists on first run when stdin is not a TTY", async () => {
+  it("defaults to OFF (opt-out) and persists on first run when stdin is not a TTY", async () => {
+    // SAP-1988: the invasive Sapiom→BQ tier is opt-out by default — a non-TTY
+    // first run (CI/Docker/headless) must never silently opt the user in.
     const wasTTY = process.stdin.isTTY;
     Object.defineProperty(process.stdin, "isTTY", { value: false, configurable: true });
 
     try {
       const result = await ensureConsent({ noTelemetry: false });
-      expect(result.telemetryOptIn).toBe(true);
+      expect(result.telemetryOptIn).toBe(false);
       expect(result.source).toBe("default-silent");
 
       const settings = await loadSettings();
-      expect(settings.telemetryOptIn).toBe(true);
+      expect(settings.telemetryOptIn).toBe(false);
     } finally {
       Object.defineProperty(process.stdin, "isTTY", { value: wasTTY, configurable: true });
     }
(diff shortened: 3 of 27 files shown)


gwitwer:
## Claude review

The two-tier consent split (light PostHog on-by-default vs. invasive BQ opt-out) is coherent, the `before-send` redaction is well-tested, and the default flip is correctly propagated through the CLI/tests. No migrations and no throw-on-missing env (the PostHog key falls back to a default, empty disables cleanly), so **no deployment risk**. Three findings.

### 🟠 Medium — `inspectingDeadSession` is defined, tested, and read, but never populated → the observe-dead-session journey is unreachable

`journeys.ts` and `coarseViewLabel` both branch on `view.inspectingDeadSession`, and `journeys.test.ts` asserts it maps to `agent_observe` / `session_history`. But the only place a `HarnessView` is actually built — `App.tsx` (`registerViewContext` effect) — omits the field:

```ts
const view: HarnessView = {
  firstRun: st.firstRun === true,
  settingsOpen,
  templatesOpen,
  hasLiveSession: st.sessions.some((s) => s.status !== "exited"),
  rightTab,
};   // ← no inspectingDeadSession
```

**What breaks:** a user reviewing a finished/historical session (all sessions exited, so `hasLiveSession` is false) falls through every rule to `journey: "unknown"`, `view: "unknown"`. So the `agent_observe` funnel undercounts, and — worse for this design — those views pollute `unknown`, which the module's own doc says is meant to be *the signal that a rule is missing*. The passing unit test masks it because it feeds the field the App never sets. Either wire `inspectingDeadSession` into the App's view descriptor or drop the branch + its test.

### 🟡 Low — the init-time `$pageview` can leak once for a previously-opted-in user who is now env-forced-off

`initAnalytics` sets `opt_out_capturing_by_default: optOutToStart` and relies on `syncConsent(state)` *after* `posthog.init` to enforce the hard kill-switch. But `persistence: "localStorage"` means posthog restores a prior opt-in preference, and a stored preference takes precedence over `opt_out_capturing_by_default`. So a user who opted in on a previous boot and then sets `DO_NOT_TRACK`/`--no-telemetry` will fire the `capture_pageview` during `init()` — before `syncConsent` opts them out. The payload is sanitized (token stripped, localhost origin+path only), so the leak is minimal, but it contradicts the PR's explicit "hard kill-switch always wins / never fires" guarantee. Consider calling `posthog.opt_out_capturing()` before init when `isHardOff(state)`, or gating the `posthog.init` call itself.

### 🟡 Low — the privacy-critical consent gate has no test

`before-send.ts` and `journeys.ts` are well covered, but `posthog.ts` — `shouldCapture`, `syncConsent`, and the env-forced-off gating that is the whole privacy contract — has zero unit tests (only `resetAnalyticsForTest` is exported for it). This is the highest-consequence logic in the PR; a regression here silently captures from opted-out users. Worth a small test around the two-tier gate.

---

**Nit (folded, since findings exist):** the event registry (`agent.run_*`, `secrets.*`, `agent.turn_stuck`) is fully typed but only `session.started` is emitted — the motivating "see where Yash & Oded got stuck on secrets" signal isn't captured by this PR, just scaffolded. This is disclosed in the PR's follow-ups, so it's expectation-setting, not a defect.

**Verdict:** Solid, privacy-forward plumbing; the `inspectingDeadSession` wiring gap is the one thing I'd fix before merge (it's a silent measurement hole in a PR whose entire point is measurement). The other two are low-cost hardening.

gwitwer:
Thanks — all three addressed in c2e3a88.

**🟠 `inspectingDeadSession` wiring gap (the merge-blocker):** fixed. App's `registerViewContext` effect now computes it from the active session's status (`active?.status === "exited"`), so reviewing a finished session maps to `agent_observe` / `session_history` instead of polluting `unknown`. Added `harness.activeSessionId` to the effect deps.

**🟡 init-time `$pageview` leak:** fixed, and I took the stronger route you hinted at — `initAnalytics` now **returns before `posthog.init()` when consent is off** (`!shouldCapture`), rather than relying on `opt_out_capturing_by_default` (which, as you noted, a localStorage-restored opt-in overrides). So an env-forced-off / opted-out user never loads posthog at all — nothing fires. When consent later flips on, the Settings toggle re-runs the effect (`initialized` still false) and inits then. Removed the now-moot `opt_out_capturing_by_default`.

**🟡 consent gate untested:** added `posthog.test.ts` (6 cases) covering exactly the privacy contract — env-forced-off and product opt-out never init, on-consent inits + `identify` + org `group`, missing key disables, and consent flips opt in/out correctly.

**Nit (only `session.started` emitted):** correct — the `agent.run_*` / `secrets.*` / `agent.turn_stuck` registry is scaffolded, and wiring those onto the deploy/run/secrets controls is the next enrichment pass (tracked on SAP-1988). Autocapture already captures those clicks today with journey/view context; this adds the named-outcome layer on top.

Suite is green: typecheck + 1640 unit tests (6 new) + lint.

PR #478, opened 2026-08-01
Templates: golden gallery curation — remove 3, promote 3 Fixables (SAP-2301)

gwitwer:
Curates the onboarding template gallery to the golden bar from the 2026-08-01 golden-path audit (`plans/onboarding-templates/TEMPLATE-QUALITY-BAR-2026-08-01.md`): **every card must produce a real, inspectable artifact on a zero-setup run.**

Closes SAP-2301.

## Removals (registry/gallery config)
Dropped from `examples/registry.json` (27 → 24 templates). Example dirs stay as pattern references; the harness reads the live catalog, so no fallback list needs touching.

- **durable-backfill** — durability/restart mechanics demo; a `{}` run counts a synthetic integer, writes nothing, reports green `complete:true`. No "what is Sapiom" artifact.
- **the-brain** — a *pattern* (agents-managing-agents), not a template.
- **wait-for-webhook** — a *pattern* (durable pause/resume), not a template.
- **hello-agent** kept as the starter/smoke-test.

## Promotions (Fixable → Golden)
Each satisfies the audit's **default-not-dryRun** and **honest-terminal-classification** bars.

- **content-repurposing-pipeline** — `dryRun` is now an explicit author opt-in (no longer inferred from a defaulted source). A zero-setup run renders the real quote graphics via `images.create` (surfaced as inspectable links in the terminal), then stops short of the pricey teaser clip — matching the golden precedent (`personalized-media-at-scale` keeps video off the zero-setup path). No-recipient terminal is declared `completed_partial` with `unmet:[deliverTo]`. A caller-supplied `source` renders the full pack incl. the clip.
- **scheduled-research-brief** — with no `deliverTo`, delivers to the agent's own Sapiom-hosted inbox (via the existing `resolveSenderInbox` machinery) and terminates `delivered:true` with an inspectable message, so the "email" headline fires zero-setup. Real `deliverTo` stays the upgrade; `dryRun` stays the explicit no-send opt-in.
- **news-roundup** — default `companyName` swapped to a reliably-covered brand (`Nvidia`) so a zero-input run finds selectable news and publishes; the no-news / empty-selection cases now route to a **distinct degrade terminal** (`published:false` + note) instead of a green `completed` that shipped nothing; `parseSelection`'s `.min(1)` softened. `zeroSetup.expect` tightened to assert a real published artifact.

## Verification
- `pnpm examples:check` ✅ (registry schema/sort/sourcePaths + 24 manifests)
- `pnpm examples:check:test` ✅ (87 tests; manifest-check floor updated 25 → 24)
- Per-example `tsc --noEmit` ✅ for all three edited templates
- Registry `setup` blocks regenerated via `examples:sync-setup` (narratives mirror the new zeroSetup)

The workspace `build`/`typecheck`/`lint` gates target `packages/*` only; this diff is entirely under `examples/` (outside the pnpm workspace) plus one `scripts/*.test.mjs` threshold, so those gates are unaffected. CI covers them.

https://claude.ai/code/session_01SSDSxQn7dFZhpZdwMLNHS8

### DIFF
--- a/examples/content-repurposing-pipeline/index.ts
+++ b/examples/content-repurposing-pipeline/index.ts
@@ -102,6 +102,8 @@ interface Shared extends Record<string, unknown> {
   deliverTo: string | null;
   aspectRatio: string;
   model?: string;
+  /** Whether to render the (pricey) teaser clip. Off for the built-in sample run. */
+  renderClip: boolean;
   pack: Pack;
   graphics: Graphic[];
   clip: Clip | null;
@@ -493,6 +495,12 @@ const repurpose = defineStep({
     ctx.shared.set("deliverTo", input.deliverTo?.trim() || null);
     ctx.shared.set("aspectRatio", ASPECT_RATIO);
     if (input.model) ctx.shared.set("model", input.model);
+    // The teaser clip (image-to-video) is the priciest, slowest leg. A run nobody
+    // configured — the built-in sample — renders the quote graphics but stops short of
+    // the clip, so a zero-setup run still fires a real visual artifact without becoming
+    // the most expensive one in the gallery. Supply your own `source` and the full pack
+    // (clip included) renders.
+    ctx.shared.set("renderClip", !usedSampleSource);
     ctx.shared.set("pack", pack);
     ctx.shared.set("graphics", []);
     ctx.shared.set("clip", null);
@@ -503,11 +511,10 @@ const repurpose = defineStep({
       quotes: pack.quoteGraphics.length,
     });
 
-    // dryRun: trace the graph and read the copy without paying for graphics/clip.
-    // It defaults on ONLY when the source was defaulted too — image and video are
-    // the priciest capabilities here, and a run nobody configured should not be the
-    // most expensive one in the gallery. Supply a source and the full pack renders.
-    if (input.dryRun ?? usedSampleSource) {
+    // dryRun is an explicit author opt-in — never inferred from a defaulted source —
+    // to trace the graph and read the copy without paying for any generated media. A
+    // zero-setup run (no dryRun) falls through to `graphics` and renders real artwork.
+    if (input.dryRun) {
       ctx.logger.info("dryRun — returning copy only");
       return terminate({
         dryRun: true,
@@ -515,7 +522,7 @@ const repurpose = defineStep({
         pack,
         note: [
           ctx.shared.get("note"),
-          "No quote graphics or teaser clip were rendered, and nothing was emailed — this is the copy only. Pass `dryRun: false` for the full pack.",
+          "No quote graphics or teaser clip were rendered, and nothing was emailed — this is the copy only. Omit `dryRun` for the full pack.",
         ]
           .filter(Boolean)
           .join(" "),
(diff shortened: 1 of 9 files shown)


gwitwer:
### CI note: the one red check (`playwright-mock`) is a pre-existing `main` failure, not a regression from this PR

`playwright-mock` fails on a single test — `web/e2e/dismiss.spec.ts:47 › settings popover › closes on a click anywhere outside` (30s timeout at line 57, `.brand-name` click). 291 other Playwright tests pass, including every template/catalog test (`templates.spec.ts`, `welcome.spec.ts` "shows the live catalog with facets").

This exact test is already failing on `main`:
- `main` HEAD run [30720362622](https://github.com/sapiom/sapiom-js/actions/runs/30720362622) — same test, same line, `1 failed / 291 passed`.
- The last 5 consecutive pushes to `main` all show the Test workflow red.

This PR's diff is entirely under `examples/` (outside the pnpm workspace) plus one `scripts/*.test.mjs` threshold — it touches no harness/web code, so it cannot affect a settings-popover dismiss test. I re-ran the failed job twice; it fails deterministically, matching `main`.

Every check that this change can influence is green: **examples**, **test (20.x)**, **test (22.x)**, **desktop-smoke**, **CodeQL**, **Analyze**.

The `dismiss.spec.ts` failure looks like harness-UI test drift on `main` and should be addressed by the harness owners in a separate change — it's out of scope for this template-gallery curation ticket (SAP-2301).

gwitwer:
**Review — PR #478 · golden gallery curation (SAP-2301)** · round 1 of at most 2 automatic rounds

### 🟠 Medium — `news-roundup` zeroSetup is now coupled to live web data, so a correct run can fail the gate

The contract flips from `status: nonEmptyString` (always green) to `status: "published"` + `siteUrl` + `articles: nonEmptyArray`. That now requires, on the day the UAT runs, that a live `web.search` for the default `"Nvidia"` returns hits **and** the LLM selects ≥1 as recent news. When coverage is thin or the model returns `[]`, the run correctly routes to the new `noNews` terminal (`published:false`) — and the zero-setup gate goes red for a run that behaved exactly as designed. Nvidia is well-covered so it will usually pass, but the gate is now nondeterministic where it wasn't. The template change (honest no-news degrade) is good; it's the *assertion* that's now flaky. Consider asserting the union (`published` **or** the `noNews` degrade terminal) rather than only the happy path, or accept the known flake explicitly.

### 🟡 Low — deployed `scheduled-research-brief` with no `deliverTo` now self-emails on every cron tick, indefinitely

Previously a no-recipient run skipped sending. Now the non-dry path always sends — to the agent's own Sapiom-hosted inbox when `deliverTo` is unset. That's the right call for a one-shot zero-setup demo, but this template is meant to be **deployed on a `schedule`**: a user who deploys it and attaches the cron without setting `deliverTo` gets an email to the demo inbox every tick, forever, where before it was a no-op. Worth bounding (self-deliver only on a manual/first run) or at least calling out in `notes`.

### What checks out (verified, expected to be broken)

- The zero-setup skip-clip path in `content-repurposing` does **not** crash: `must()` throws only on `undefined`, `clip` is `null` on that path, and `value?.fileId` is null-safe. The `graphics` array rides the no-recipient terminal via `...summary`, so the flipped `graphics: nonEmptyArray` assertion holds, and the new `graphics → package` edge targets a real step.
- Registry removals are config-only: the three dropped dirs remain as pattern references, 27 → 24, and the manifest-check floor 25 → 24 matches.

No deployment risk: examples-only diff plus one test threshold and an e2e click-coordinate fix — no env vars, migrations, or backend surface.

**Verdict:** Approve. The gallery is genuinely tighter and the terminal-honesty changes are the right direction. The one real follow-up is the `news-roundup` gate flakiness (Medium); everything else is sound.

PR #481, opened 2026-08-02
feat(templates): declare & honor per-resource DB reuse for every Postgres template (SAP-2319)

gwitwer:
## What & why

Runtime **foundation** for the DB-reuse epic ([SAP-2303](https://linear.app/sapiom/issue/SAP-2303)). Makes every gallery Postgres resource honestly **runtime-repointable** and self-describing, so the reuse picker ([SAP-2320](https://linear.app/sapiom/issue/SAP-2320)) can offer *"use a database you already have"* for any declared Postgres — with zero author pre-wiring — and **never show a picker the next run silently ignores** (`design-v2.md` § Landmines).

The [SAP-2191](https://linear.app/sapiom/issue/SAP-2191) handle-read is opt-in per template. Audit of the 8 Postgres templates: 3 read the injected handle **and** declared a key (picker works today), 2 read it but declared nothing, and 3 hardcoded. A picker built on a hardcoded template would be a dishonest affordance — so the runtime must honor the handle **before** the UI may offer it.

Carries no delete risk (delete is already reference-counted, SAP-2190) and no UI, so it ships ahead of the picker.

## Interface contract (owned here, consumed by SAP-2320)

`resources[]` entry gains **`reuse?: { key: string }`**:
- **present** ⇒ user-repointable; `key` is the top-level entry-input path the run reads via `resolveResourceHandle` (default `dbHandle`).
- **absent** ⇒ internal state, not user data — no picker.

## Changes

- **Schema** (`examples/template.schema.json`): add the `reuse` descriptor.
- **Runtime migration** — route the handle read through the canonical `resolveResourceHandle` seam:
  - `approval-chain` → key `ledgerHandle`
  - `nl-db-query-endpoint` → key `dbHandle`
  - Both preserve opt-in / conditional-default semantics via `fallback: ""` — a zero-setup run is **never** forced to provision Postgres, and a user-named handle that 404s is still rejected rather than silently reprovisioned.
- **Declarations** — add `reuse` to the 7 reusable manifests (`meeting-notes-crm`, `cold-outreach-engine`, `error-triage-digest`, `personalized-media-at-scale`, `scheduled-db-insight-report`, `approval-chain`, `nl-db-query-endpoint`).
- **`the-brain`: non-reusable.** Its DB is the brain's own `fleet_events` state, not user data, and it is excluded from the gallery by design (T0.2). The **absence** of a `reuse` descriptor is the marker — index.ts is left hardcoded, deliberately.
- **Lint** (`examples-check` §5c, `manifest-resource-reuse`): `reuse` is postgres-only, and its declared `key` must actually be read via `resolveResourceHandle` in the same `index.ts` — so declaration and code can never drift and the picker can never advertise a control the run ignores. Rejects a malformed descriptor (schema) and a hardcoded-handle drift (semantic).
- **Tests**: fixture cases for the schema field and the semantic check, plus a pass over **every real reusable manifest against its index.ts**.

## Verification

- `pnpm examples:check` ✅ — 24 templates, schema-valid.
- `pnpm examples:check:test` (`node --test`) ✅ — 98/98 pass (incl. new reuse cases).
- Confirmed the lint **rejects** a malformed `reuse` (missing/extra/dotted key) and a drift (reuse key not read via `resolveResourceHandle`).
- Prettier-clean on all changed `.mjs`/`.ts`. No behavior change for the 3 already-reusable templates.

> **Follow-up (not blocking):** the acceptance criterion "verified against a run/trace" is best exercised in the funded-tenant baseline pass — deploy each reusable template, set the injection key to a second handle, and confirm the trace opens it. The static `manifest-resource-reuse` check mechanically enforces the honesty invariant (every declared key is read via the seam) in the meantime.

Closes SAP-2319.

### DIFF
--- a/examples/approval-chain/index.ts
+++ b/examples/approval-chain/index.ts
@@ -3,6 +3,7 @@ import {
   defineStep,
   goto,
   pauseUntilSignal,
+  resolveResourceHandle,
   terminate,
   type AgentExecutionContext,
 } from "@sapiom/agent";
@@ -409,9 +410,17 @@ const start = defineStep({
       "escalateTo",
       input.escalateTo?.trim() || config.ESCALATION_EMAIL || null,
     );
+    // The durable ledger is opt-in: absent a handle the chain runs off `ctx.shared`
+    // alone and provisions no Postgres. Read the chosen handle through the canonical
+    // injection seam (declared as `resources[].reuse.key: "ledgerHandle"`) so a
+    // deploy-time reuse picker can point it at a ledger DB the tenant already owns.
+    // Fallback is `""`, not a default handle: an empty result must degrade to null,
+    // never force a managed database onto a run that asked for none.
     ctx.shared.set(
       "ledgerHandle",
-      input.ledgerHandle?.trim() || config.LEDGER_HANDLE || null,
+      resolveResourceHandle(input, { key: "ledgerHandle", fallback: "" }) ||
+        config.LEDGER_HANDLE ||
+        null,
     );
     ctx.shared.set("maxReminders", input.maxReminders ?? DEFAULT_MAX_REMINDERS);
     ctx.shared.set("dryRun", input.dryRun === true);
--- a/examples/approval-chain/template.json
+++ b/examples/approval-chain/template.json
@@ -18,7 +18,8 @@
       "kind": "postgres",
       "handle": "approval-chain-ledger",
       "duration": "7d",
-      "ephemeral": false
+      "ephemeral": false,
+      "reuse": { "key": "ledgerHandle" }
     }
   ],
   "settings": [
--- a/examples/cold-outreach-engine/template.json
+++ b/examples/cold-outreach-engine/template.json
@@ -18,7 +18,8 @@
       "kind": "postgres",
       "handle": "cold-outreach-engine",
       "duration": "7d",
-      "ephemeral": false
+      "ephemeral": false,
+      "reuse": { "key": "dbHandle" }
     }
   ],
   "settings": [
--- a/examples/error-triage-digest/template.json
+++ b/examples/error-triage-digest/template.json
@@ -18,7 +18,8 @@
       "kind": "postgres",
       "handle": "error-triage-digest",
       "duration": "7d",
-      "ephemeral": false
+      "ephemeral": false,
+      "reuse": { "key": "dbHandle" }
     }
   ],
   "settings": [
--- a/examples/meeting-notes-crm/template.json
+++ b/examples/meeting-notes-crm/template.json
@@ -18,7 +18,8 @@
       "kind": "postgres",
       "handle": "meeting-notes-crm",
       "duration": "7d",
-      "ephemeral": false
+      "ephemeral": false,
+      "reuse": { "key": "dbHandle" }
     }
   ],
   "settings": [
--- a/examples/nl-db-query-endpoint/index.ts
+++ b/examples/nl-db-query-endpoint/index.ts
@@ -2,6 +2,7 @@ import {
   defineAgent,
   defineStep,
   goto,
+  resolveResourceHandle,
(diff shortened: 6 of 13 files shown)


gwitwer:
## Claude review — round 1

Solid PR. The runtime migration is behavior-preserving (`resolveResourceHandle` trims and falls back exactly as the old `?.trim() || … || null` chains did), all 7 declared templates genuinely read their handle through the seam with a matching key, and `the-brain`'s omission of `reuse` is the correct honesty marker. No deployment risk: SDK/examples/lint only — no env vars, migrations, or infra.

One finding worth acting on before this becomes the foundation SAP-2320 builds a picker on.

### Medium — the honesty check can be satisfied by prose, not code (`scripts/examples-manifest-check.mjs:231`)

`readsInjectedHandle` is the load-bearing guarantee of this PR — "declaration and code can never drift." But it doesn't verify the declared key is read by an actual `resolveResourceHandle` call; it does two looser things:

1. **Default key (`dbHandle`)** — returns `true` if *any* `resolveResourceHandle(` call exists anywhere in the file (`if (key === "dbHandle") return true`). A template declaring `reuse.key: "dbHandle"` whose only call is `resolveResourceHandle(input, { key: "somethingElse" })` passes while never reading `dbHandle`.

2. **Non-default key** — matches `key: "<key>"` against the *entire source, including comments and unrelated object literals*. This is already live in your own diff: `approval-chain/index.ts` now contains the comment ``…reuse.key: "ledgerHandle"``, whose substring `key: "ledgerHandle"` satisfies the regex on its own. Delete the real `resolveResourceHandle(input, { key: "ledgerHandle" })` call and the check still passes on the comment alone.

Concrete failure: a future reusable template passes the lint with a `resolveResourceHandle` call reading a *different* handle than declared, plus a mention of the declared key in a comment — and the picker then advertises a control the run silently ignores. That is exactly the dishonest affordance the check exists to reject.

Neither loophole is exercised by any current template, so this is not a CI break today — but the check is the mechanism the whole epic leans on, and the tests only cover the happy path. Tightening it to confirm the key appears inside a `resolveResourceHandle(...)` argument (rather than anywhere in the source), and dropping the unconditional `dbHandle` short-circuit in favor of "a bare call OR an explicit `key: 'dbHandle'`", would close both. Worth a fixture case for each loophole too.

**Verdict:** Approve on the migration and declarations; recommend hardening `readsInjectedHandle` before SAP-2320 depends on its guarantee.

gwitwer:
Good catch — this is the load-bearing check for the epic, so both loopholes are worth closing. Fixed in 2086795.

Replaced the whole-source regex in `readsInjectedHandle` with `keysReadViaResolver`: it locates each `resolveResourceHandle(` call, extracts that call's argument list by a balanced-paren scan, and reads the key from the call's own `key: "…"` option (or `dbHandle` for a bare call). Then `readsInjectedHandle(source, key) = keysReadViaResolver(source).has(key)`.

This closes both:
1. **Default-key short-circuit** — gone. `dbHandle` is credited only by a call that actually reads it (a bare call, or an explicit `key: "dbHandle"`), so a lone `resolveResourceHandle(input, { key: "somethingElse" })` no longer satisfies a declared `dbHandle`.
2. **Comment/whole-source match** — gone. Only text inside a real call's parens is inspected, so the `reuse.key: "ledgerHandle"` doc comment (and any unrelated object literal) can't forge a read. Delete the real `resolveResourceHandle(input, { key: "ledgerHandle" })` call and the check now fails on `approval-chain`.

Also, per your note, a non-literal key value (a variable) is intentionally **not** credited — the picker needs a statically declared key, so "compute the key at runtime" should fail the declaration rather than pass it.

Added a fixture for each loophole (declared `dbHandle` + call reading another key → fails; key present only in a comment → fails). `examples:check:test` now 100/100, `examples:check` green.

PR #491, opened 2026-08-03
feat(tools): searchindex namespace (ctx.sapiom.searchindex) + search.map() (SAP-2255)

ynadge:
## Summary

Adds the provider-neutral `searchindex` namespace to `@sapiom/tools` and `ctx.sapiom.searchindex`, covering the full provisioned-index control and data planes. It also adds structured, gateway-direct `search.map({ url })` support for the corpus-refresh workflow.

This revision includes the release-blocking contract corrections and independent-review fixes from SAP-2255: canonical management envelopes, valid range/query defaults, fail-closed response parsing, positional fetch integrity, read/write type separation, nullable regions, logical meter documentation, and a production-faithful stateful stub.

## Changes

- Adds control-plane `create/get/list/update/delete`; returned indexes bind `upsert/query/range/fetchDocuments/deleteDocuments` to their own Sapiom data-plane URL.
- Parses canonical `{ databases: [...] }` list responses and only the explicitly documented compatibility shapes. Malformed successful list/query/range/fetch responses throw `SearchIndexContractError` instead of becoming empty data.
- Verifies non-null `fetchDocuments` results preserve the requested ID at each position; wrong or swapped IDs fail closed while positional `null` misses remain valid.
- Splits required write payloads (`SearchDocumentInput`) from read documents whose `content`/`metadata` fields remain absent unless included. `region` is `string | null`.
- Defaults queries to 5 results and maps public `limit` to raw `topK`; query payloads explicitly request content and metadata. Defaults range to `{ cursor: "0", limit: 100 }`.
- Shares SearchIndex and map input validation between the live client and stub. Invalid map inputs reject before transport or stub resolution.
- Makes successful `search.map` responses fail closed with exported `SearchContractError` unless they contain a valid links array; every link requires a non-empty URL and typed optional title/description fields.
- Coerces plain-JSON SearchIndex `create/get/list/update` stub overrides into registry-backed bound handles, preserving stateful data-plane behavior.
- Generates deterministic stub resource IDs matching `^res_[a-z0-9]{20}$`, so local workflows can pass IDs unchanged to the management and MCP surfaces.
- Documents the routing boundary: web search, scrape, and email use Core's routed capability seam; map calls the Firecrawl gateway directly at `/v2/map`.
- Documents provider-invisible SearchIndex meters and current catalog rates. The backing provider remains an implementation detail of the customer contract.
- Adds the `./search-index` package subpath, barrel/client wiring, smoke coverage, and a minor changeset.

## Contract evidence

- Unified gateway DTO/list contract: [`SearchDatabaseResponse` and `{ databases }`](https://github.com/sapiom/Sapiom/blob/4ba50cd31f7ec36d141f28174a22e211ddc19bd4/x402/apps/unified/src/providers/upstash/management/search.service.ts#L67-L104), [`SearchService.list()`](https://github.com/sapiom/Sapiom/blob/4ba50cd31f7ec36d141f28174a22e211ddc19bd4/x402/apps/unified/src/providers/upstash/management/search.service.ts#L257-L264).
- Official Search client wire mapping, pinned to `upstash/search-js@27f0c045`: public `limit` defaults to 5 and is sent as raw `topK` with `includeData: true` and `includeMetadata: true` ([source lines 76-116](https://github.com/upstash/search-js/blob/27f0c045cb7c776091c1c27571d937710ee0a6d6/src/search-index.ts#L76-L116)).
- Range fixtures use the canonical live `{ result: { nextCursor, vectors } }` shape captured on 2026-08-03; the only retained compatibility page is the documented Search SDK `{ nextCursor, documents }` shape.

## Verification

- Rebased on `sapiom-js/main@bbc3bcfa9ffcd4e97e1e310e8f9523fe3c3718e6`; `git range-diff` confirms the four SAP-2255 commits are patch-equivalent.
- `pnpm --filter @sapiom/tools exec jest --runInBand` — 29/29 suites, 583/583 tests.
- `pnpm --filter @sapiom/tools typecheck`.
- `pnpm --filter @sapiom/tools lint`.
- `pnpm --filter @sapiom/tools build` — CJS + ESM builds.
- CJS and ESM root/`./search`/`./search-index` runtime import smokes.
- Emitted declaration and package export-map inspection, including root and `./search` `SearchContractError` exports.
- `pnpm examples:check` — 11 templates and 50 registered assets valid.
- `pnpm terminology:check` — 373 files valid with no stale allowlist entries.
- `pnpm changeset status --since=origin/main` — `@sapiom/tools` minor.
- `npm pack --dry-run --json` — expected runtime, types, and docs present.

## Release boundary

The changeset schedules the next minor release. This PR does **not** publish the package, update the workflow engine, deploy, merge, or mutate production. Publication remains gated on SAP-2386 proving the authoritative SearchIndex metering rollout; engine pickup is separate follow-up work.

Closes: SAP-2255

### DIFF
--- a/.changeset/sap-2255-searchindex-namespace.md
+++ b/.changeset/sap-2255-searchindex-namespace.md
@@ -2,24 +2,26 @@
 "@sapiom/tools": minor
 ---
 
-Add the `searchindex` namespace (`ctx.sapiom.searchindex`) and `search.map()` (SAP-2255).
+Add the `searchindex` namespace (`ctx.sapiom.searchindex`) and `search.map()`
+(SAP-2255).
 
-`searchindex` covers the full priced Upstash Search surface behind the Sapiom
+`searchindex` covers the complete provisioned-search surface behind the Sapiom
 gateway — control plane `create`/`get`/`list`/`update`/`delete`, with
 `create`/`get`/`list` returning a bound `SearchIndex` handle carrying the
 data-plane operations (`upsert`/`query`/`range`/`fetchDocuments`/
-`deleteDocuments`) against the index's own `*.search.data.sapiom.ai` URL.
-Documents are `{ id, content, metadata? }` — `content` is auto-embedded
-server-side, `metadata` is stored verbatim (put URLs/content-hashes there).
-`range` is the enumeration/reconciliation primitive; data-plane calls are
-priced per request ($0.000050; `query` reranking +$0.001). Errors throw the new
-`SearchIndexHttpError`.
+`deleteDocuments`) against the index's own Sapiom data-plane URL. Write inputs
+require `{ id, content, metadata? }`; read results preserve omitted content and
+metadata. `range` defaults to a valid first page (`cursor: "0"`, `limit: 100`).
+Malformed successful envelopes fail closed with `SearchIndexContractError`.
 
-`search.map({ url })` exposes Firecrawl site mapping ($0.009 flat) with
-structured `{ links: [{ url, title?, description? }] }` output — gateway-direct,
-since `/v2/map` is not a routed capability.
+The public contract names provider-invisible meters: `searchindex.index`,
+`searchindex.upsert`, `searchindex.query`, `searchindex.query_rerank`,
+`searchindex.range`, `searchindex.fetch_documents`, and
+`searchindex.delete_documents`.
+
+`search.map({ url })` exposes site mapping with structured
+`{ links: [{ url, title?, description? }] }` output.
 
 Both are wired into the `Sapiom` client interface, `bind()`, the barrel, the
-`./search-index` subpath export, and the stub client (stateful in-memory
-indexes, so `run_local` exercises upsert → query/range/fetch/delete flows
-deterministically).
+`./search-index` subpath export, and a stateful, validation-faithful stub client
+so `run_local` exercises upsert → query/range/fetch/delete flows.
--- a/.changeset/sap-2255-searchindex-namespace.md
+++ b/.changeset/sap-2255-searchindex-namespace.md
@@ -19,8 +19,9 @@ The public contract names provider-invisible meters: `searchindex.index`,
 `searchindex.range`, `searchindex.fetch_documents`, and
 `searchindex.delete_documents`.
 
-`search.map({ url })` exposes site mapping with structured
-`{ links: [{ url, title?, description? }] }` output.
+`search.map({ url })` exposes gateway-direct site mapping with structured
+`{ links: [{ url, title?, description? }] }` output. Malformed successful map
+responses fail closed with `SearchContractError`.
 
 Both are wired into the `Sapiom` client interface, `bind()`, the barrel, the
 `./search-index` subpath export, and a stateful, validation-faithful stub client
--- a/.changeset/sap-2255-searchindex-namespace.md
+++ b/.changeset/sap-2255-searchindex-namespace.md
@@ -1,0 +1,25 @@
+---
+"@sapiom/tools": minor
+---
+
+Add the `searchindex` namespace (`ctx.sapiom.searchindex`) and `search.map()` (SAP-2255).
+
+`searchindex` covers the full priced Upstash Search surface behind the Sapiom
+gateway — control plane `create`/`get`/`list`/`update`/`delete`, with
+`create`/`get`/`list` returning a bound `SearchIndex` handle carrying the
+data-plane operations (`upsert`/`query`/`range`/`fetchDocuments`/
+`deleteDocuments`) against the index's own `*.search.data.sapiom.ai` URL.
+Documents are `{ id, content, metadata? }` — `content` is auto-embedded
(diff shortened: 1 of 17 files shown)


gwitwer:
## Review — verdict: REVISE before merge

The surface is largely correct: endpoint paths, hosts, headers, the `ttl` (`/^\d+[mhd]$/`) and `indexName` (`/^[a-zA-Z0-9_-]+$/`) regexes, the `/v2/map` body shape, and gateway host resolution for the `upstash`/`firecrawl` subdomains all match the backend contract (`backend/src/mcp/tools/upstash.tool.ts`, `firecrawl.tool.ts`, `config/env.ts`) exactly. Build + lint + 524 package tests pass. But there's one confirmed production-breaking bug.

### 🔴 `searchindex.list()` returns `[]` on every real call — the response is wrapped, not a bare array
`packages/tools/src/search-index/index.ts:1241`

```js
const raw = (await res.json()) as RawSearchIndexResponse[];
return Array.isArray(raw) ? raw.map(...) : [];
```

`GET /v1/search/indexes` does **not** return a bare array — the gateway wraps it as `{ databases: [...] }`. Confirmed in `backend/src/mcp/tools/upstash.tool.ts:70` (`// Gateway returns search indexes under the 'databases' key`) and the backend regression test `tools.spec.ts` — *"lists search indexes returned under the gateway databases key (AGENT-230)"*. Since `raw` is an object, `Array.isArray(raw)` is `false` and `list()` silently returns `[]` on every live call. The PR's spec masks this by mocking a bare array (`jsonResponse([indexDto(), ...])`), which doesn't match production — so the README's headline "resolve an index by name" example (`(await searchindex.list()).find(...)`) is broken as written.

**Fix:** unwrap `raw.databases ?? raw` (mirroring the backend's own `formatDatabaseListResponse` fallback) before mapping, and add a spec case using the real `{ databases: [...] }` shape.

### 🟡 Nit — wrong id-prefix in doc comment
`packages/tools/src/search-index/index.ts:828` documents the index id as `res_…`. The backend's actual prefix is `srch_` (`upstash.tool.ts:787,811`; `srch_abc` fixtures). Harmless (id is opaque `string`) but it's a wrong example in public docs — likely copied from the `database` capability, which does use `res_…`.

Nicely done on the defensive `{ result: [...] } | [...]` unwrapping in `query`/`fetch`/`delete-documents` — that same discipline just needs to reach `list()`.

ynadge:
Automated Claude/gwitwer review disposition for current head `a10fd9495041d1dbfcf1f3c26c7d97810613a274`:

1. **Wrapped `{ databases }` list response — valid finding on the reviewed revision; resolved.** Commit `881a9c9` introduced the canonical unwrap. The current parser accepts `{ databases: [...] }`, retains only the explicitly documented bare-array compatibility shape, and fails closed on unknown envelopes ([implementation](https://github.com/sapiom/sapiom-js/blob/a10fd9495041d1dbfcf1f3c26c7d97810613a274/packages/tools/src/search-index/index.ts#L405-L413), [regressions](https://github.com/sapiom/sapiom-js/blob/a10fd9495041d1dbfcf1f3c26c7d97810613a274/packages/tools/src/search-index/index.spec.ts#L248-L284)). Fresh targeted run: 3/3 list-contract tests passed.

2. **`srch_` prefix nit — stale/incorrect; no SDK change.** The authoritative gateway generator uses `res_` plus 20 lowercase alphanumeric characters ([generator](https://github.com/sapiom/Sapiom/blob/99cbac760d4c2579c7276a929e2d9514285bf080/x402/apps/unified/src/providers/upstash/shared/resource-id.ts#L9-L32)), and Search creation calls that generator ([service](https://github.com/sapiom/Sapiom/blob/99cbac760d4c2579c7276a929e2d9514285bf080/x402/apps/unified/src/providers/upstash/management/search.service.ts#L116-L123)). The current MCP boundary independently enforces `^res_[a-z0-9]{20}$` ([schema](https://github.com/sapiom/Sapiom/blob/99cbac760d4c2579c7276a929e2d9514285bf080/backend/src/mcp/tools/upstash.tool.ts#L17-L27)). Remaining `srch_` Swagger examples/legacy fixtures on `origin/dev` are stale documentation, not the generated resource contract. The SDK's `res_…` wording is therefore correct.

No unresolved code finding remains from this automated review, so no additional commit was required.

ynadge:
Final CI cleanup: the branch is now rebased onto `sapiom-js/main@bbc3bcfa9ffcd4e97e1e310e8f9523fe3c3718e6`.

The previous `examples` failure was not in the SAP-2255 implementation: the old base expected two stale private Canvas terminology counts. Current `main` contains the upstream allowlist correction. `git range-diff` confirms all four SAP-2255 commits remain patch-equivalent after the rebase.

Fresh verification on the rebased head:

- `@sapiom/tools`: 29 suites / 583 tests
- typecheck, CJS+ESM build, and lint passed
- examples check passed (11 templates, 50 assets)
- terminology check passed (373 files, no stale entries)
- `git diff --check` passed

No package publication, merge, deployment, or production mutation was performed.

PR #492, opened 2026-08-03
feat(content-generation): add images.launch async dispatch to fix 30s router 503s

gwitwer:
## Summary

Image generation was failing with **503 Heroku "Application Error"** pages under any fan-out. The routed synchronous `images.create` holds its HTTP request open for the full generate+store, which meets Core's **30s router cap (Heroku H12)** — a `Promise.all` over N rows drove every request in the batch past 30s, so the step 503'd on every retry.

The backend already supported async image dispatch (`dispatch: 'async'`, SAP-1802) over the **same** fal-queue → webhook → resume rail as video, but the SDK never exposed it. This was an asymmetry (video got the full launch/pause/webhook treatment; images didn't), not an outage. This PR closes the gap **SDK-only** — no backend/gateway/engine change is required, because the async completion→resume path is already media-agnostic (resumes on `correlationId == requestId`, delivers a generic `outputs[]` payload).

### `@sapiom/tools` (new public API — minor)
- Add `contentGeneration.images.launch` mirroring `video.launch`: submits with `dispatch: 'async'`, forwards `x-sapiom-workflow-token`, returns an `ImageLaunchHandle` (`requestId`, `dispatch`, inline `wait()`). The submit returns as soon as the job is enqueued, so the 30s wall no longer applies.
- Add `IMAGE_RESULT_SIGNAL`, the `ImageResultPayload` shape a resumed step receives, and `toImageResumePayload`.
- Extend the `capabilityCall` seam to forward request headers (so the routed async verb can carry the resume token).
- Wire the client namespace, barrel exports, and the run-local stub.
- `images.create` is unchanged.

### Agent (`personalized-media-at-scale`)
- Rewrite the blocking `renderImages` fan-out into a `renderImage` ⇄ `collectImage` launch + `pauseUntilSignal` loop — a one-to-one mirror of the video path (`renderClip` ⇄ `collectClip`).
- Images now render **sequentially** (a paused step waits on a single `(signal, correlationId)` pair — the same reason video is sequential). Tradeoff vs. the old all-at-once fan-out, but the fan-out was the 503 trigger.

## Testing
- [x] `@sapiom/tools`: typecheck, lint, **517 tests pass** (+21 new for `launchImage` / `toImageResumePayload` / prompt-guard)
- [x] `@sapiom/agent` (92) and `@sapiom/agent-core` (133) suites pass — stub change is safe
- [x] Example typechecked against the locally-built SDK (caught a stale step-name reference); `defineAgent` graph validation passes with all 6 steps
- [ ] End-to-end deploy of agent 451 (see below)

## Notes
- A full `run_local` of this example's image path isn't a clean signal: the `fetch` step needs a real Postgres connection that stubs don't provide (affects the video path identically, predates this change).
- To ship agent 451: publish the SDK change, then `link` → `deploy` the example from a checkout with the updated `@sapiom/tools`.

## Related
- Builds on SAP-1802 (backend async image dispatch) / SAP-1116 (routed capability seam)

### DIFF
--- a/.changeset/images-launch-async-dispatch.md
+++ b/.changeset/images-launch-async-dispatch.md
@@ -1,0 +1,20 @@
+---
+"@sapiom/tools": minor
+---
+
+`contentGeneration.images.launch` — a dispatchable async surface for image generation, mirroring
+`video.launch`.
+
+The routed synchronous `images.create` holds its HTTP request open for the full generate+store,
+which meets Core's 30s router cap: a concurrent fan-out (`Promise.all` over N rows) drove every
+request in the batch past 30s, so the whole step 503'd on every retry. The backend already supported
+async image dispatch (`dispatch: 'async'`, SAP-1802) over the same fal-queue → webhook → resume rail
+as video, but the SDK never exposed it.
+
+`images.launch` submits with `dispatch: 'async'`, forwards the workflow resume token, and returns an
+`ImageLaunchHandle` (`requestId`, `dispatch`, and an inline `wait()`) — so the submit returns as soon
+as the job is enqueued and the 30s wall no longer applies. Pass the handle to
+`pauseUntilSignal(handle, { resumeStep })` to suspend a workflow step until the image is ready, or
+`await handle.wait()` to poll inline. Also exported: `IMAGE_RESULT_SIGNAL`, the `ImageResultPayload`
+shape a resumed step receives, and `toImageResumePayload`. `images.create` is unchanged. No backend
+change is required — the async completion→resume path is media-agnostic.
--- a/examples/personalized-media-at-scale/index.ts
+++ b/examples/personalized-media-at-scale/index.ts
@@ -7,7 +7,12 @@ import {
   terminate,
   type AgentExecutionContext,
 } from "@sapiom/agent";
-import { VIDEO_RESULT_SIGNAL, type VideoResultPayload } from "@sapiom/tools";
+import {
+  VIDEO_RESULT_SIGNAL,
+  IMAGE_RESULT_SIGNAL,
+  type VideoResultPayload,
+  type ImageResultPayload,
+} from "@sapiom/tools";
 import postgres from "postgres";
 import { z } from "zod/v4";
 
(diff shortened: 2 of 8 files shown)


gwitwer:
🤖 **Claude review** — run manually with this repo's CI review prompt (`.github/workflows/claude-code-review.yml`). SDK-only, no migrations/env/infra, so no deployment-risk block.

### `handle.wait()` polls the routed result URL with the wrong auth header

`launchImage` submits through the routed capability seam, which forces `authHeader: "x-api-key"` (`packages/tools/src/_client/capability-call.ts:81`). But the inline poll — `transport.fetch(responseUrl, { method: "GET" })` at `packages/tools/src/content-generation/index.ts:421` — passes no `authHeader`, so it falls back to `DEFAULT_AUTH_HEADER = "x-sapiom-api-key"` (`packages/tools/src/_client/index.ts:104`).

Video's `launch` is internally consistent: its submit *and* its poll both use the default `x-sapiom-api-key`, because video is gateway-direct. Images broke that symmetry — the submit is Core-routed (`x-api-key`) but the poll reverts to the gateway header. If the Core `responseUrl` requires `x-api-key` the way the submit does, every `wait()` poll 401s and the loop spins until the 2-minute deadline throws `Image generation did not complete within 120000ms` — a hang, not a clear error.

Two things make this easy to miss:
- The agent's real path uses `pauseUntilSignal`, not `wait()`, so the primary flow is unaffected. `wait()` is only hit by the documented inline path ("call `handle.wait()` to block inline").
- All 3 new `wait()` tests use a mock `fetch` that ignores auth, so the header is never asserted — this asymmetry is completely untested.

Please confirm the routed result endpoint accepts `x-sapiom-api-key`. If it doesn't, thread the routed header through to the poll (e.g. `transport.fetch(responseUrl, { method: "GET" }, { authHeader: "x-api-key" })`) and add a test that asserts the poll's auth header, not just the submit's.

---

Verdict: **One thing to confirm before merge** (the `wait()` poll auth). The rest is clean — I checked the claims that would bite: `launchImage`'s body is byte-for-byte identical to `createImage` plus `dispatch:'async'`; `renderImage`⇄`collectImage` faithfully mirrors the video launch/collect loop (`next:[]` + `pause`, `mediaIndex` shared cursor); and `must(mediaIndex)` is safe at `0` since `must` tests `=== undefined`.

PR #496, opened 2026-08-03
[SAP-2362] Fix Studio starter scaffolding

ynadge:
## Summary

- let Agent Core scaffold safely into a folder containing only Agent Studio private state
- preserve and ignore .sapiom, write the sapiom.json discovery marker, and retain the non-empty-directory safety boundary
- route Studio starter prompts through sapiom_dev_agents_scaffold with exact arguments instead of relying on the desktop-only CLI
- add package and browser regression coverage plus patch changesets

## Verification

- 18 Agent Core scaffold tests
- 27 Harness template prompt tests
- 28 Chromium template and Add Workspace journey tests
- Agent Core, MCP, and Harness typechecks
- Harness dependency graph production build
- Agent Studio terminology gate
- clean-room source dogfood: Studio-shaped scaffold, npm install, full check, registry discovery, and local start-to-finish run with no warnings

Linear: SAP-2362

### DIFF
--- a/.changeset/brave-starters-scaffold.md
+++ b/.changeset/brave-starters-scaffold.md
@@ -1,0 +1,7 @@
+---
+"@sapiom/agent-core": patch
+"@sapiom/harness": patch
+"@sapiom/mcp": patch
+---
+
+Make bundled starters work from a fresh Agent Studio session by preserving Studio-owned state, writing the project discovery marker, and handing Claude the local scaffold MCP tool directly.
--- a/packages/agent-core/src/__tests__/scaffold.test.ts
+++ b/packages/agent-core/src/__tests__/scaffold.test.ts
@@ -7,6 +7,7 @@
 import { execFileSync } from "node:child_process";
 import {
   existsSync,
+  mkdirSync,
   mkdtempSync,
   readFileSync,
   readdirSync,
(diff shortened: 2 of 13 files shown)


gwitwer:
## Review — verdict: SHIP

Clean, tightly-scoped fix. Verified by building the PR commit in an isolated worktree and running the suites: 18 agent-core scaffold + 27 harness templates + 28 Playwright e2e = 73 passing, typecheck clean.

Two changes, both correct:
1. `agent-core`'s `scaffold()` now tolerates a pre-existing real `.sapiom/` directory (Studio's private session state) as if the target were empty, while still hard-rejecting any other pre-existing content; writes the `sapiom.json` discovery marker and adds `.sapiom/` to both bundled templates' `.gitignore`. The `isScaffoldableTarget`/`walk` skip logic, symlink rejection, and `writeConfig(targetDir, {})` marker write all match test expectations.
2. The web app's prompts now tell the coding agent to call the `sapiom_dev_agents_scaffold` MCP tool with `{dir, template}` instead of running `sapiom agents init . -t <name>` — a desktop-only CLI command that isn't available in a browser Studio session. Correct fix.

### 🟢 Minor (non-blocking, not a bug)
`packages/harness/web/src/App.tsx` `handleScaffoldInSession` derives `cwd` from `state.sessions` and falls back to `"."` if the session isn't found. Under the old shell-command flow `.` resolved to the session's PTY cwd; under the new MCP-tool flow `.` would resolve relative to the MCP server's own cwd. The fallback should be unreachable (the session is always present when this fires), so not flagging as a bug — just noting the `"."` fallback is slightly fragile and a real absolute `cwd` would be safer.

Nice — no stale `sapiom agents init` references left in the touched files, scope is tight, changeset correctly scoped to `agent-core`/`harness`/`mcp`.

PR #500, opened 2026-08-03
[SAP-2358] Align Studio workspace discovery

ynadge:
## Summary

- share one agent-project marker, depth, and ignored-directory contract across registry scans, live watching, and the folder picker
- reconcile confirmed removed or malformed markers, preserve entries through transient read failures, and clear genuinely stale live-session bindings
- preserve Studio-owned `.sapiom/` while cloning a gallery template through an isolated staging checkout
- replace false bundled-starter offline claims with the actual account, spend, and npm boundary

## Dogfood

A disposable filesystem tree reproduced the original three-way disagreement: the registry admitted generated trees, the watcher admitted malformed markers, and the picker badged them. Against this branch, registry and watcher both return only the valid projects at depth 1 and 3, while the picker reports false for ignored and malformed candidates.

A permissions regression also reproduces the reconciliation edge case: when a registered project directory becomes temporarily unreadable, a rescan now preserves both its in-memory and persisted registry entries. Confirmed missing or malformed markers still reconcile away.

Pre-merge Studio acceptance exposed a debounce race when an unreadable project was restored and immediately became malformed. The watcher now fingerprints unreadable subtrees as an opaque state, so the subsequent invalid state always triggers reconciliation instead of leaving a stale rail entry and session binding.

## Review status

- Removed the unrelated partial MCP alias rename; the complete convention is handled in [sapiom-js #513](https://github.com/sapiom/sapiom-js/pull/513) and [Sapiom #4041](https://github.com/sapiom/Sapiom/pull/4041).
- Registry reconciliation now distinguishes valid, absent, invalid, and unreadable marker states. An unreadable subtree is opaque for that scan, preventing transient filesystem failures from pruning persisted entries or exposing nested projects incorrectly.
- Live watcher snapshots preserve the unreadable boundary as a distinct fingerprint, preventing an unreadable-to-invalid transition from being swallowed by event debouncing.
- The CodeQL path-expression finding is fixed: marker reads resolve the fixed `sapiom.json` beneath the selected directory, prove containment, and reject marker symlinks. The alert is closed.

## Verification

- full Harness suite: 113 files / 1,668 tests passed, plus 4 performance tests
- `pnpm --filter @sapiom/harness build`
- `pnpm --filter @sapiom/agent-core typecheck`
- `pnpm --filter @sapiom/mcp typecheck`
- lint for Agent Core, Harness, and MCP (one pre-existing Harness warning, zero errors)
- Agent Core clone suite: 10 passed
- Playwright Add Workspace + templates: 32 passed
- no-delay live Studio acceptance: discovery, rail reconciliation, binding removal, and context clearing passed
- exact locally published MCP package cloned a production definition while preserving `.sapiom/harness-context.json` and leaving no staging directories
- `pnpm terminology:check`

Restacked onto current `main` on 2026-08-04. The PR contains only the five SAP-2358 commits (20 files); inherited #496 history and the `App.tsx` conflict are gone. GitHub reports the branch mergeable.

Linear: https://linear.app/sapiom/issue/SAP-2358/docs-document-studio-workspaces-projects-templates-and-sessions


gwitwer:
## Review — verdict: NEEDS CHANGES

The core of this PR is solid: the shared `agent-project-discovery.ts` (marker/ignore-dir/depth contract unified across registry, watcher, and folder picker), the scan reconciliation (drops scan-sourced entries whose marker disappeared/became malformed, never touches connect-sourced ones), and the `.sapiom/`-preserving staged clone all check out and pass their tests (51 harness + 10 agent-core, typecheck clean). One blocker.

### 🔴 The Claude Code MCP alias rename collides with the reserved remote-capability name
`packages/harness/src/core/adapters/claude-code-info.ts:14`, `packages/mcp/README.md:32`

The PR changes the setup instruction from `claude mcp add sapiom-dev -- npx -y @sapiom/mcp` to `claude mcp add sapiom -- …`. But `sapiom` is explicitly the reserved name of the **remote paid capability** MCP everywhere else:

- `packages/harness/src/core/inject/mcp-config.ts:57` auto-injects `sapiom` = `{type: "http", url: "https://api.sapiom.ai/v1/mcp"}` (remote) and `sapiom-dev` = this local dev MCP, side by side, in the same file.
- `packages/mcp/src/index.ts:42` — the server's own wire name is still `"sapiom-dev"`.
- Sibling adapters for the *same* Claude Code CLI were left as `sapiom-dev`: `conductor.ts:21`, `codex-info.ts:17`, `opencode.ts:17`, `pi.ts:16`.
- `docs/mcp-servers.md` (untouched) exists specifically to prevent this: *"conflating them is the most common source of confusion."*
- A prior changeset in this same package (`packages/mcp/CHANGELOG.md` 0.8.1) renamed this MCP **to** `sapiom-dev` precisely "to stop it being conflated with the remote `sapiom` capability MCP."

Result: `packages/mcp/README.md` is now self-contradictory (opens with "under the server name `sapiom-dev`", shows `"sapiom-dev"` JSON keys, then tells Claude Code users to register it as `sapiom`), and a user following the new instruction in a Studio/Conductor session ends up with two registrations that disagree on which server is `sapiom`. This reintroduces the exact conflation a previous fix in this package removed — the opposite of "aligning discovery."

**Fix:** revert `claude-code-info.ts` + the README's `claude mcp add` line to `sapiom-dev` (matching every other adapter, the wire name, and the docs). If a project-wide rename to `sapiom` is genuinely wanted, it has to be done consistently — `mcp/src/index.ts` name, the injected key, the remote entry (to avoid collision), all sibling adapters, README, and `docs/mcp-servers.md` — with tests asserting the new contract.

### 🟡 Low — non-atomic clone move (optional)
`packages/agent-core/src/clone.ts` — if `renameSync` throws partway through moving multiple top-level entries out of the staged clone, the `finally` deletes the staged dir regardless, so not-yet-moved entries are lost while moved ones remain (partial checkout beside the preserved `.sapiom/`). Narrow failure mode (same-device rename), not blocking — worth a comment or a cleanup-on-failure branch if full atomicity matters.

ynadge:
The review blocker is resolved at this PR's current head: `ccf7ce7` removed the partial alias rename, so #500 keeps `sapiom-dev` consistently. The approved public alias policy is implemented later as one coordinated cutover in #514, where local authoring becomes `sapiom` and hosted direct access becomes `sapiom-direct` across the affected package/UI/docs surfaces.

I also fixed the CodeQL path-injection finding in `23af979`: marker reads now resolve the fixed `sapiom.json` beneath the selected directory, prove containment, and reject marker symlinks before reading. CodeQL now passes and the alert is closed. Local verification passed 113 Harness files / 1,666 tests, four performance tests, typecheck, and lint (one pre-existing warning only).

gwitwer (dismissed):

PR #501, opened 2026-08-04
[SAP-2359] Make Studio workbench state truthful

ynadge:
## Summary

- distinguish a linked agent definition from a runnable cloud build by consuming the existing active-build projection
- keep Test, Deploy, and Run independent of Claude readiness while preserving authentication and ready-build gates
- make Canvas, Steps, Code, the workflow rail, and the dashboard link agree on draft/linked/building/failed/ready state
- enrich the bound workflow before explicit and automatic Canvas renders so its deployment badge uses current cloud build state
- correct integration snippets, local-run boundaries, retry evidence, and deterministic Canvas failure handling

## Dogfood findings fixed

- a failed first deploy could retain a definition ID and incorrectly enable Run and Code
- direct actions were disabled while Claude was starting even though those actions bypass the coding-agent session
- signing in after a signed-out boot did not refresh deployed-agent metadata
- Canvas used the raw workflow registry during rendering, so a ready deployment could still appear merely linked
- Canvas run evidence could bleed across agent rebinding or show an older retry attempt
- Canvas render failures never reached the actionable error overlay, and Retry only reloaded the same failed document
- TypeScript and cURL snippets had different runtime semantics but shared one inaccurate hint; cURL also did not encode the definition path

## Verification

- `pnpm build` at the repository root — passed
- `pnpm typecheck`, `pnpm lint`, and `pnpm test` at the repository root — passed (lint has existing warnings only)
- `pnpm test` in `packages/harness` on the final stacked base — 117 files / 1,699 tests plus 4 performance tests passed
- `pnpm typecheck` in `packages/harness` on the final stacked base — passed
- `pnpm test:ui -- --workers=4` in `packages/harness` — 298 Playwright tests passed
- Canvas build-status regression exercised red/green and passes through the live server render route

Linear: [SAP-2359](https://linear.app/sapiom/issue/SAP-2359/docs-document-canvas-steps-code-and-direct-actions)

Stacked on #500.

PR #502, opened 2026-08-04
[SAP-2364] Make local authoring contracts truthful

ynadge:
## Summary

- make local execution parse every step input through the authored Zod schema, matching production defaults and key stripping
- normalize relative project paths in `check`, and add a real-compiler regression
- clone gallery repositories through an isolated staging directory and retry only Git branch-propagation failures
- correct Local Run, stub, retry, template-count, and gallery-lifecycle guidance across Agent Core, MCP, Harness, bundled starters, and Hello Agent
- add a patch changeset for Agent Core, MCP, and Harness

## Why

SAP-2364 dogfooding found three runtime/documentation gaps: relative `check` paths could resolve twice, local step inputs bypassed their schemas, and a newly seeded gallery fork could briefly be visible through GitHub REST but not smart HTTP. It also found public product copy that described Local Run as fully offline/free even though arbitrary author code and usage analytics remain real; only Sapiom capability traffic is stubbed.

## Dogfood evidence

- scaffolded, installed, typechecked, checked, and locally ran both bundled starters with telemetry disabled
- exercised built-in and explicit coding-pause stubs, including the failure branch
- cloned the live Hello Agent template into Yash's organization and completed a zero-input local run with `{ "greeting": "Hello, world!" }`
- reproduced the Git propagation failure before the clone correction, then completed the same live path after the fix
- copied the exact proposed docs example into a fresh scaffold and completed both default and explicit-name runs

## Conflict resolution

- merged current `main` at `5cf3160` with a normal merge commit (no history rewrite)
- retained the original 41-file SAP-2364 scope and added one guardrail-only allowlist correction after CI exposed drift already present on `main`
- discarded three stale inherited deltas that were not part of the SAP-2364 commit

## Verification after resolving conflicts

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint` (0 errors; existing repository warnings only)
- Agent Core: 158 tests passed
- MCP: 92 tests passed
- Harness: 1,717 unit tests and 4 performance tests passed
- PR-specific template browser journey: 17 Playwright tests passed
- full UI suite: 298/300 passed on the parallel run; the two unrelated Canvas zoom-readiness failures both passed on an immediate serial rerun (4/4 Links-block tests)
- examples check: 11 manifests and 50 registered authoring/source assets passed
- `git diff --check`
- the pre-follow-up CI run passed Node 20, Node 22, full Playwright mock/canvas suites, desktop smoke, and CodeQL

## CI follow-up

- corrected two stale terminology allowlist counts for private `launched-workflow` compatibility identifiers introduced by #501
- reran `pnpm examples:check` and `pnpm terminology:check`; both pass locally
- refreshed CI passed all eight checks on the allowlist-only follow-up commit, including Node 20/22, examples/terminology, Playwright, desktop smoke, and CodeQL

Linear: [SAP-2364](https://linear.app/sapiom/issue/SAP-2364/docs-publish-build-and-local-test-guides)


gwitwer:
## Review — verdict: SHIP WITH NITS

The three functional fixes are real, correctly implemented, and covered by passing tests (158 agent-core + 92 mcp, full build green): (1) `check()` now resolves `sourceDir` to an absolute path before using it as both execFile path and cwd; (2) `runLocal`'s dispatcher now runs a step's Zod `inputSchema` before `step.run`, so local runs apply the same defaults/key-stripping as production; (3) `clone()` stages into a temp dir and retries only GitHub's branch-propagation-window error. The doc-truthfulness corrections ("offline and instant", "26 templates", `agent.coding.launch` → `models.coding.launch`, the never-implemented stub-array semantics) are legitimate. Two things to tidy.

### 🟡 The new regression test for the relative-`sourceDir` fix doesn't test the fix
`packages/agent-core/src/check.spec.ts:855` ("runs the project compiler when sourceDir is relative")

It calls `runTypecheck()` directly — not `check()`, where the actual `path.resolve(opts.sourceDir)` fix lives — with a relative dir computed via `path.relative(cwd, mkdtempSync(tmpdir()))`, which is dominated by `../../..` segments. Reproduced: that `..`-heavy path **passes even without the fix** (the double-resolution cancels out). The real bug reproduces reliably with a plain non-`..` subdir (e.g. `check({ sourceDir: "my-agent" })` from its parent). So this test gives false confidence — it wouldn't catch a regression of the exact fix, and doesn't touch the changed code path.

**Fix:** test `check()` end-to-end with `process.chdir()` into a parent and `sourceDir: "<subdir>"` (no `..`), or create the temp project under `process.cwd()` so the relative path has no `..` segments.

### 🟡 Large unrelated reformatting bundled in
Diffed against the true base (`c5f8c3b6`): the overwhelming majority of changed lines in `web/src/lib/api.ts` (+379/-112), `use-harness-state.ts` (+412/-182), and several agent-core files are pure single→double-quote conversion and re-wrapping to satisfy `.prettierrc.json` (never enforced before — sibling files like `git.ts`/`deploy.ts` remain single-quoted). This inflates a ~300-400 line change into 2656/-1272 and reformats files unrelated to "authoring contracts." Not a bug, but please split the drive-by formatting into its own PR/commit — or at minimum flag those hunks in the description so reviewers skim rather than line-review them.

### 🟢 Nit — changeset note
`StepStubs` narrowing from `StubResponse | StubResponse[]` to `StubResponse` is a correct "truthful contract" fix (array-as-sequence was never implemented — confirmed in `tools/src/stub/index.ts`), but it's a public type change; one sentence in `.changeset/honest-local-runs-parse.md` would help TS consumers who typed `StubResponse[]`.

PR #508, opened 2026-08-04
fix(studio): preserve local work through auth failure

ynadge:
## Summary

- preserve signed-out local Agent Studio startup when the desktop browser-auth flow is cancelled, times out, or fails
- keep cached desktop authentication silent and retain smoke mode's no-browser contract
- make the shared MCP/Studio callback page host-neutral
- return an honest failure page for invalid state, missing code, and failed token exchange instead of reporting authentication complete early
- add a patch changeset and focused regressions

## Dogfood

- verified the real environment-aware Studio/MCP credential cache without printing or changing its credential
- deployed a one-step configuration canary, then proved write-only secret listing, production secret injection, saved-default updates without redeploy, caller override, reset, and cleanup
- confirmed Local Run does not fetch cloud secrets but does inherit ordinary host environment

## Verification

- dependency builds through `@sapiom/mcp` and `@sapiom/harness-desktop`
- MCP suite: 92 tests passed
- desktop suite: 58 tests passed
- MCP ESLint and typecheck
- desktop typecheck and production build
- changed-file Prettier check and `git diff --check`

Stacked on [sapiom-js #505](https://github.com/sapiom/sapiom-js/pull/505).

Linear: [SAP-2361](https://linear.app/sapiom/issue/SAP-2361/docs-document-authentication-secrets-and-runtime-defaults)

### DIFF
--- a/.changeset/calm-studios-auth.md
+++ b/.changeset/calm-studios-auth.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/mcp": patch
+---
+
+Make browser-auth completion pages host-neutral and report failed callbacks honestly. Agent Studio desktop now preserves signed-out local work when account authentication is unavailable or cancelled.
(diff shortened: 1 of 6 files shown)


ynadge:
Lineage corrected without rewriting commits: this PR now targets #505 head `yashnadge/sap-2366-correct-schedule-and-signal-contracts`, its direct parent. GitHub now shows one intended commit (`5f843c8`) and a clean, mergeable diff.

PR #512, opened 2026-08-04
feat(harness): Studio brand redesign + frameless macOS window; desktop 0.2.0

ratataque:
## What

Redesigns the Sapiom Studio web app (`@sapiom/harness`) to the new brand, adds a **frameless macOS window** to the desktop app (`@sapiom/harness-desktop`), and cuts a **0.2.0** desktop release.

Built across several sessions; this PR also **syncs 26 commits of `main`** (the product-wide "workflow" → "agent" rename, `@sapiom/agent-studio` npx package, `/agents` routes) into the redesign.

## Redesign (web app)

- **Brand & tokens** — the shared design-system `sapiom-studio` preset (Geist type, compact IDE density, scarce green reserved for live/success/on/confirmed, neutral ink focus/selection), vendored into the public `ds-neutral` fallback so OSS + packaged builds render with no private-registry dependency.
- **Rail lockup** — the Sapiom wordmark + `agent.studio` (lowercase mono), matching a new terminal masthead.
- **Three flat blocks** — grey rail + grey terminal shell + a raised white graph pane, no permanent dividers; the affirmative CTA is the ink button, green only on genuine state.
- **One header** — session bar + tab lane + actions merge; the active session is a title dropdown, sibling live sessions are switch chips with a trailing `+`; actions are Prod · Test · Run · Deploy.
- **Depth** — the raised graph column (dotted board included) is a lighter white lifted off a slightly darker shell by a left-edge shadow that paints the **full** column height. It must sit above the terminal's opaque, absolutely-positioned `xterm-viewport` (which had been overpainting it from the terminal's top down) — scoped to desktop (`min-width: 769px`) so the mobile bottom sheet's fixed positioning is untouched.
- **Canvas render** — light-theme nodes/success use the brand green, not an off-brand cyan; the board is white.

## Frameless macOS window (desktop)

- Drops the native title bar (`titleBarStyle: hiddenInset`) and insets the OS traffic lights into the rail's 56px top line, which becomes the window's drag handle (every control on it opts back out, so buttons/tabs still click).
- The SPA detects the host via a `frame=macos` query param and splits the rail header across two rows — wordmark + theme/collapse tools beside the lights, `agent.studio` on the line below. Entirely inert in the `npx` browser app.

## Merge with `main`

Kept the redesign's structure/visuals, layered main's terminology + functional changes:
- `lib/urls.ts` centralized helper renamed to `agentUrl` / `SAPIOM_AGENTS_URL`, pointed at `/agents`.
- **Completed the bound-agent chip** the redesign had left wired as a dead prop (`refine.css` already styled it), so main's canvas "Open agent" flow and its e2e render it.
- e2e assertions moved to the redesign DOM (`agent.studio` product, footer-anchored settings popover, session-menu dot) with main's renamed fixtures (`rfq-agent`, `Agent Studio` title).

## Desktop release 0.2.0

Minor bump (`0.1.5` → `0.2.0`) for the rebrand + new window mode (the desktop is private — no changeset). **The release is cut after this merges** by pushing the `harness-desktop-v0.2.0` tag on the merged commit — the tag must equal `package.json`'s version, which `desktop-release.yml`'s `prepare` job enforces.

## Testing

- `pnpm --filter @sapiom/harness build` (tsc + vite) — pass
- e2e — **296 passed**
- unit — **1657 passed**
- Verified the frameless window + all visual refinements (full-height lift shadow, darker shell, green canvas nodes, two-row mac header) in the running Electron app.


gwitwer:
## Review — Studio brand redesign + frameless macOS window

Solid PR. The load-bearing pieces hold up under scrutiny: the Electron↔web frame contract is correct on both sides, the z-index lift genuinely raises the graph pane above xterm's opaque viewport, and the e2e suite was migrated to the new DOM without losing coverage. No high/medium defects. What follows is all low-severity.

### 🟡 Low — the raised graph pane's surface depends on CSS import order; reorder it and the flagship depth effect silently disappears
`styles.css` declares `.right-pane`/`.session-bar`/`.right-pane-tabs` on the grey base surface; `refine.css` re-declares the same three selectors at the same specificity with the raised-white + left-edge-shadow treatment. The redesign only wins because `main.tsx` imports `styles.css` **then** `refine.css`. Reorder those two imports (or a bundler that reorders) and `.right-pane` falls back to `--surface-base` grey — the entire depth effect this PR is built around vanishes, with no error. The base-layer declarations are now dead and contradict the redesign; collapse the redesign values into one place rather than layering them over stale ones by import luck.

### 🟡 Low — `RailSort = "recent"` doesn't actually sort agent rows by recency
`lib/workspace-tree.ts` — `agentOrder(sort)` special-cases only `"name"`; every other value including `"recent"` falls through to `a.path.localeCompare(b.path)` (alphabetical by path). `WorkflowInfo` carries no timestamp, so agent rows within a group *cannot* be recency-sorted — only folder order honors recency. The `RailSort` doc comment promises "newest activity first," which the control can't deliver for agents. Either document that agent rows are path-stable, or drop the pretense.

### 🟡 Low — dead code + a weak e2e assertion
- `lib/mock-terminal.ts` — `visibleLength` is now unused (its only caller `welcomeBox()` was deleted). It won't fail CI because `lint` only covers `src/`, not `web/`, and `noUnusedLocals` is off — which is itself worth noting: nothing lints `web/` for unused symbols. Delete it.
- `e2e/portable-continue.spec.ts` — `await expect(context).not.toHaveAttribute("data-session-id", "")` also passes when the attribute is *absent*, so on its own it's a weak guard; it's saved by the paired visible/not-`sess-pricing` assertions. Tighten or drop the empty-string check.

### One thing to confirm
The drag-region `no-drag` escape hatch only releases `button`/`a`/`input`/`[role=...]`. If the session-title dropdown or the trailing `+` in `SessionBar` is a plain `<div onClick>` without `role="button"`, the drag strip will swallow its clicks on the desktop app. Worth a quick check that those are real buttons or carry the role.

### Verified (non-obvious, actively checked)
- Frame contract end-to-end: main process sets `hiddenInset` + `trafficLightPosition` + appends `?frame=macos` **only** under `process.platform === "darwin"`; the SPA reads the param and defaults to `"web"`, does not OS-sniff, and the two-row rail split is CSS-only keyed on `[data-window-frame="macos"]` — a browser cannot get desktop chrome. Windows/Linux keep the native frame.
- z-index: `.right-pane { position: relative; z-index: 1 }` scoped to `min-width: 769px` (exact complement of the mobile `max-width: 768px`), lifts above the absolute `.xterm-viewport` without trapping the portaled flyout or the splitters. All new custom props resolve in both light and dark ramps.
- e2e: removed smoke tests were each replaced with a renamed new-DOM equivalent (net 83 tests, no coverage loss); the deploy-failed test was strengthened; no `waitForTimeout`/`.only`/`.skip`. `harness-desktop` 0.1.5→0.2.0 bumped manually — correct, since it's `private: true` and changesets won't publish it.

**Verdict:** Ship-able. Fold in the import-order consolidation (Low #1) since it protects the whole redesign, and confirm the SessionBar controls aren't drag-swallowed; the rest are cleanup.

<sub>🤖 Reviewed with Claude Code following the repo's CI review instructions.</sub>

gwitwer (dismissed):
small nits, lgtm!

PR #521, opened 2026-08-04
fix(content-repurposing-pipeline): mint quote-graphic download URLs before packaging

gwitwer:
## Summary
`contentGeneration.images` returns only a `fileId` (plus a `downloadUrlUnavailable` signal), not a `downloadUrl`. The per-graphic URL was minted **only** inside the `clip` step, so a run that skips the clip (`renderClip: false` — the built-in sample path) reached `package` with every graphic URL unset. The pack markdown omitted the graphic links, the run output reported `downloadUrl: null`, and the run still terminated as **`completed`** — a false success delivering a pack with no viewable graphics.

Observed in prod run `agents/487/runs/156924`: both quote graphics `downloadUrl: null` while the pack file itself got a valid signed URL.

## Changes
- `package` step: mint each graphic's URL from its `fileId` via `ctx.sapiom.fileStorage.getDownloadUrl` (the same resolution the `clip` step already uses) before `renderPackMarkdown`, best-effort — a mint hiccup drops one link rather than failing the run.

## Testing
- [x] `tsc --noEmit` clean, prettier clean, `node --test index.test.mjs` 6/6 pass.

## Checklist
- [x] No hardcoded secrets
- [x] Self-reviewed

https://claude.ai/code/session_01WruzmfS2a9gXcJtKhRLMUF

### DIFF
--- a/examples/content-repurposing-pipeline/index.ts
+++ b/examples/content-repurposing-pipeline/index.ts
@@ -620,6 +620,28 @@ const packageStep = defineStep({
     const graphicsList = must(ctx.shared.get("graphics"), "graphics");
     const value = must(ctx.shared.get("clip"), "clip");
 
+    // `contentGeneration.images` returns only a `fileId` (with a
+    // `downloadUrlUnavailable` signal), so a graphic's usable URL must be minted
+    // from its fileId here — the same resolution the `clip` step already does.
+    // Without this, a run that skips the clip (renderClip=false) packages and
+    // delivers null graphic links yet still reports success. Best-effort: a mint
+    // hiccup drops that one link rather than failing the run (the pack ships inline).
+    for (const g of graphicsList) {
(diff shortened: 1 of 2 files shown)


gwitwer:
Updated to `main` and added the CI fix.

The red `examples` job was **not** the content-pack change — `examples:check` passed; it was `terminology:check` failing on a **stale allowlist on `main`** (the harness redesign added references to the persisted `"launched-workflow"` Canvas node-kind literal without bumping `scripts/agent-studio-terminology-allowlist.json`). Bumped `canvas-body-node-kind` 2→3 and `web-canvas-node-kind` 1→2 — these are the stable compatibility identifier the entries already except, not new drift.

Verified locally: `terminology:check` passes (373 files, no stale entries) and `examples:check` passes. Bundled here so merging this greens the tree for future example PRs.

PR #531, opened 2026-08-05
feat(harness): observable Run / Test / Deploy in the Steps view

ratataque:
## What

Makes the **Test / Run / Deploy** actions observable. Before, clicking one changed almost nothing on screen — no view switch, no reveal, no success feedback, and the demo prod run snapped straight to "completed". Now each action reveals the right pane and switches it to the **Steps** view (one unified activity surface), the run visibly progresses, and the relevant final data surfaces up front.

## Behaviour

- **Auto-switch + reveal** — Run/Test/Deploy each open the pane to the Steps view. Gated so a click that would only toast (a prod-run with no ready build) never yanks the view.
- **See it move** — steps advance pending → running → passed in view; the acting button carries a `data-running` brand pulse tied to the *real* run status (not just the hand-off ring); the demo prod run now progresses across polls on a wall clock. (Progressive mock is gated strictly to `run()`-minted `exec-mock-prod-*` ids, so the existing terminal-fixture specs don't regress.)
- **Run-summary card** headlines the Steps surface — outcome, live "N of M steps", total duration (Σ step latencies). The run's **produced result is the primary CTA** (a filled brand button); the deployed dashboard link + dev-server preview drop to secondary flat links; when the run produced nothing, the dashboard takes the primary slot.
- **Step detail is an inline dropdown** — clicking a step expands its full detail *in place* (Agent run status/duration/IO/logs, capability calls, contract, and Leads-to/Reached-from lineage) with the per-step coding-agent actions (Ask coding agent / Ask to modify / Copy name). No separate slide-in view.
- **Payload CTAs** — Input / Output / Logs / Result share one disclosure with a rotating caret, readable label, hover row, and a Copy button; the Result is collapsed by default.
- **Deploy as an action, not a toast** — Deploy lands in the same Steps surface with a live linking → building → deployed banner, then a completion state that links to the dashboard and jumps to the "Trigger from your code" snippet.

## Honesty

No fabricated data: no cost anywhere (per-step cost isn't in the run model — Studio stays cost-agnostic; `cost-removed` still asserts no `$`), no latency on a still-running step, payload blocks gated on `!== undefined` so a truly-absent value renders nothing.

## Not in scope

Animating the **canvas board diagram** node-by-node is deferred — the demo board doc is stale and would need regenerating; the Steps-list observability delivers the "see it move" without it. Tracked as a follow-up.

## Tests

New `web/e2e/run-observability.spec.ts` (auto-switch, `data-running`, run-summary + primary CTA, copyable result, deploy banner → code, no `$`) + `progressiveLeasingRun` unit tests. Reworked smoke / direct-actions / direct-action-feedback / rich-step-detail / run-inspector / popover-crop / canvas-inspector for the Steps-view switch and the inline step-detail dropdown.

Gates: **typecheck · e2e 286 · canvas 10 · unit 1729 · terminology · build** all green. Rebased clean on `origin/main`.

### DIFF
--- a/.changeset/observable-run-actions.md
+++ b/.changeset/observable-run-actions.md
@@ -1,0 +1,25 @@
+---
+"@sapiom/harness": minor
+---
+
+Make Test / Run / Deploy observable in Studio: clicking one now reveals the
+right pane and switches it to the Steps view (the unified activity surface),
+instead of the action landing silently.
+
+- **See it move.** A run's steps advance pending → running → passed in view; the
+  acting button carries a `data-running` pulse tied to the real run status (not
+  just the brief hand-off ring); and the demo prod run now progresses across
+  polls on a wall clock rather than snapping to "completed".
+- **Relevant final data up front.** A run-summary card headlines the Steps
+  surface — outcome, live progress, total duration, and the single most relevant
+  result CTA (the deployed agent's dashboard link → a dev-server preview →
+  URLs the run produced → the final step's output). Honest-absence throughout:
+  no cost fields, no latency on a still-running step, no fabricated values.
+- **Better payload CTAs.** Input / Output / Logs / Result share one disclosure
+  with a Copy button (the final Result renders expanded); nothing renders for a
+  payload a step never carried.
+- **Deploy as an action, not a toast.** Deploy lands in the same Steps surface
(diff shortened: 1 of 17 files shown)


LucasMarsala (approved):
lgfm

PR #538, opened 2026-08-05
fix(harness): repair the Templates info popover and let nav dismiss the screen

ratataque:
## What

Two fixes on the Studio Templates screen, both reported from the desktop app.

### 1. Info (i) popover rendered transparent
The shared "floating panel" CSS recipe (background + border + shadow + `z-index`) is opted into by class name. The template spec-sheet class `.template-facts` was never added to that list, so the popover rendered with **no surface** — the Steps / Trigger / Complexity / Capabilities list and the Preview / Use buttons painted directly over the card text behind them.

**Fix:** add `.template-facts` to the recipe (and its `prefers-reduced-motion` group). Its own later rule keeps its richer padding/width; the recipe only supplies the missing surface.

### 2. Couldn't leave the Templates screen via other nav
Templates is a sibling view gated by a `templatesOpen` boolean that hides the workbench via CSS — not a modal. The rail's other nav rows fired their handlers, but none reset `templatesOpen`, so the screen stayed painted on top and appeared to do nothing. Only the back arrow escaped.

**Fix:** clear `templatesOpen` in the navigation entry points (`openSession`, `selectTab`, `reviewPastSession`, `handleFocusAgent`, and the Create new / Overview rail actions), consistent with how they already reset `composing`. Any navigation now dismisses the screen the way the back arrow does.

## Testing
- Typecheck clean
- Full e2e suite: **287 passed** (includes a new regression test, `templates.spec.ts` — "the rail navigates away from templates")
- Verified visually in both the mock SPA and the packaged-renderer **desktop** app

## Notes
- Bug-fix `patch` changeset for `@sapiom/harness` included.

### DIFF
--- a/.changeset/template-info-and-nav.md
+++ b/.changeset/template-info-and-nav.md
@@ -1,0 +1,14 @@
+---
+"@sapiom/harness": patch
+---
+
+Fix two issues on the Templates screen in Studio:
+
+- The info (i) spec sheet popover rendered with no surface — its background,
+  border, and shadow were missing, so the Steps / Trigger / Complexity /
+  Capabilities list and the Preview / Use buttons painted directly over the card
+  behind it. `.template-facts` now opts into the shared popover elevation recipe.
+- The Templates destination hid the workbench, but the rail's other nav actions
+  never cleared it, so clicking Create new, Search, or an agent/session row left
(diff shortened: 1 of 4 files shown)


LucasMarsala (approved):
lgfm

PR #543, opened 2026-08-05
feat(harness): route sapiom://templates/<id> deep links to the gallery

gwitwer:
## Summary
Extends the existing `sapiom://` deep-link scheme (which already handled `sapiom://agent/<id>`) with a **template** target, so the web app's template-detail **"Open in Studio"** opens that template *inside* the desktop app instead of only bringing the window forward. Builds directly on the agent deep-link machinery shipped in #540.

### Desktop main process
- `DeepLinkTarget` is now a discriminated union — `DeepLinkAgentTarget { kind: "agent" }` | `DeepLinkTemplateTarget { kind: "template"; templateId }` (`ipc.ts`).
- `parseDeepLink` recognizes `sapiom://templates/<id>` (with `template` alias, trailing slash, percent-decoding), alongside the existing `agent`/`agents` (`deep-link.ts`).
- Cold-start threading generalized (`withAgentParam` → `withDeepLinkParams`): agent links ride in as `?agent=`, template links as `?template=` (`boot.ts`). Query-only, preserving the `isTrustedSender` pathname==`/` invariant.

### SPA (renderer)
- Bridge + cold-start mirrors updated to the union (`lib/desktop.ts`, `lib/deep-link.ts`; `agentFromSearch` → `deepLinkFromSearch`).
- `applyDeepLink` branches on `kind`: agent → existing focus/clone; **template → open the templates browser on that id**.
- `TemplatesPanel` takes an `openTemplateId` prop and resolves it against the live catalog (+ bundled starters), opening its detail; an unknown/not-yet-loaded id falls through to the gallery. Keyed on the id so navigating back isn't undone.

The **agent deep-link path is unchanged** — only extended.

## Testing
- [x] `deep-link.test.ts` extended (14 tests pass): agent cases now assert `kind: "agent"`; new template cases cover the alias, trailing slash, percent-decoding, optional slug, and rejection of unknown hosts / missing ids.
- [x] Changed files typecheck clean locally; full `tsc`/build deferred to CI (worktree deps not fully built).
- [ ] Manual desktop deeplink smoke (needs a packaged build with the scheme registered) — follow-up.

## Related
- Builds on the agent deep-link scheme from #540.
- Pairs with the web app's template "Open in Studio" button (separate Sapiom monorepo PR).

## Checklist
- [x] No hardcoded secrets
- [x] Self-reviewed
- [x] One logical change

https://claude.ai/code/session_0143Q8XqKSp34LyAK1Lnhhqd

### DIFF
--- a/packages/harness-desktop/src/main/boot.ts
+++ b/packages/harness-desktop/src/main/boot.ts
@@ -406,7 +406,7 @@ export async function boot(setupWin: BrowserWindow, mode: BootMode): Promise<Boo
   });
 
   // 9. Load the SPA in the main window; close setup once it renders.
-  const url = withAgentParam(`http://127.0.0.1:${server.port}/?token=${bootToken}`, mode.deepLink);
+  const url = withDeepLinkParams(`http://127.0.0.1:${server.port}/?token=${bootToken}`, mode.deepLink);
   const mainWindow = createMainWindow(url);
   mainWindow.webContents.once("did-finish-load", () => {
     if (!setupWin.isDestroyed()) setupWin.close();
@@ -420,14 +420,20 @@ export async function boot(setupWin: BrowserWindow, mode: BootMode): Promise<Boo
  * Thread a cold-start deep link's target onto the SPA load URL as query params.
  * Query only — never a path segment: `isTrustedSender` (trusted-sender.ts) fails
  * closed unless the top frame's pathname is exactly "/", so a path-based route
- * here would break the update + folder-picker IPC.
+ * here would break the update + folder-picker IPC. The SPA reads these back in
+ * `web/src/lib/deep-link.ts`.
  */
-function withAgentParam(loadUrl: string, target: DeepLinkTarget | undefined): string {
+function withDeepLinkParams(loadUrl: string, target: DeepLinkTarget | undefined): string {
   if (!target) return loadUrl;
   try {
     const u = new URL(loadUrl);
-    u.searchParams.set("agent", target.definitionId);
-    if (target.slug) u.searchParams.set("agentSlug", target.slug);
+    if (target.kind === "agent") {
+      u.searchParams.set("agent", target.definitionId);
+      if (target.slug) u.searchParams.set("agentSlug", target.slug);
+    } else {
+      u.searchParams.set("template", target.templateId);
+      if (target.slug) u.searchParams.set("templateSlug", target.slug);
+    }
     return u.toString();
   } catch {
     return loadUrl;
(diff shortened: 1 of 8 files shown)


gwitwer:
**Review round 1** — extends the `sapiom://` deep-link scheme with a `template` target. No deployment risk: desktop/SPA only, no env vars, migrations, or infra. The `parseDeepLink` refactor is clean and the discriminated union is the right call. Two things worth acting on.

### Medium — a template deep link is one-shot-per-id, so re-clicking "Open in Studio" after leaving the panel is a no-op

`applyDeepLink` handles a template by `setDeepLinkTemplateId(id)` + `setTemplatesOpen(true)` (`App.tsx`), and `TemplatesPanel` only opens the detail when `openTemplateId` *changes* (`openedByDeepLinkRef` guard). Sequence: open template A via a link → hit back-to-session (`onExit`, `templatesOpen=false`) → click the same "Open in Studio" (A) again. `setDeepLinkTemplateId("A")` is the same value so it no-ops, the ref still equals `"A"`, and the panel reopens showing the **gallery**, not A's detail.

The agent path doesn't have this: it calls `focusExistingRef.current?.(id)` on every invocation, so re-clicking an agent link always works. The keying is deliberate for the navigate-*back*-within-the-panel case (per the comment), but it also swallows a legitimate repeat of the same warm link. Consider clearing `openedByDeepLinkRef` / `deepLinkTemplateId` on panel exit, or driving the open off a per-event token rather than the bare id, so identical consecutive links still resolve.

### Medium — the new SPA-side deep-link logic is untested

`deep-link.test.ts` covers the desktop `parseDeepLink` template cases well, but the mirrored SPA parse and the panel wiring have no tests:
- `deepLinkFromSearch` (`web/src/lib/deep-link.ts`) — the new `?template=`/`templateSlug` branch and the agent-wins-if-both precedence. It's a pure function; a table test mirroring the desktop one is cheap and guards the read side of the boot→SPA contract.
- The `TemplatesPanel` `openTemplateId` effect — the resolve-against-gallery-or-starters path and the ref keying (the exact behavior in the finding above).

Given `withDeepLinkParams` writes `templateSlug` and `deepLinkFromSearch` reads it back, an untested rename on either side breaks the cold-start path silently.

### Nits
- `withDeepLinkParams` (`boot.ts`) branches on `kind === "agent"` with a bare `else` for template; an explicit `else if (target.kind === "template")` keeps it exhaustive-by-inspection when a third kind lands.

**Verdict:** Approve-with-changes — no blockers; the two mediums are a real UX gap and a coverage gap, both cheap to close.

PR #548, opened 2026-08-05
fix(agent): align local and hosted validation

oramada:
## Summary

- preserve native ESM bundle imports in the CommonJS agent-core build so `check` and `runLocalFromDir` work after publication
- register standard JSON Schema formats so manifest preflight rejects invalid URI values
- treat dispatched `StepInputValidationError` completions as terminal instead of consuming the retry budget
- add patch changesets for `@sapiom/agent-core` and `@sapiom/agent-runtime`

## Testing

- `pnpm build`
- `pnpm --filter @sapiom/agent-core test --runInBand` (160 tests)
- `pnpm --filter @sapiom/agent-runtime test --runInBand` (29 tests)
- typecheck and lint for both affected packages
- built CommonJS `loadDefinition` smoke test against the default agent template

## Follow-up

The production workflows engine has its own retry/validation boundary and needs a separate parity patch. Capability metadata normalization is also excluded because it requires a coordinated SDK/production contract change.

### DIFF
--- a/.changeset/friendly-tools-rest.md
+++ b/.changeset/friendly-tools-rest.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/agent-core": patch
+"@sapiom/agent-runtime": patch
+---
+
+Fix CommonJS local agent loading, enforce JSON Schema URI formats, and make dispatched input-validation failures terminal instead of retryable.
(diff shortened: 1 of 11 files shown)


oramada:
@claude review

oramada:
CJS is not expected to be used

PR #549, opened 2026-08-05
fix(agent-core): resolve source directory before typecheck

UfukNode:
This fixes the relative `sourceDir` path case in `runTypecheck()`.

I hit this through the existing `check.spec` coverage: when `sourceDir` is relative, the function builds a relative `node_modules/.bin/tsc` path and also uses that same relative directory as the child process cwd. That makes the tsc path resolve from inside the source directory again, so the check reports a typecheck failure even though the fixture compiler exits cleanly.

The fix is just to resolve the project directory once, then use that absolute path for both the tsc lookup and the cwd.

Checked locally:

```bash
pnpm --filter @sapiom/agent-core exec jest src/check.spec.ts --runInBand --no-cache
pnpm --filter @sapiom/agent-core typecheck
pnpm --filter @sapiom/agent-core build
```

### DIFF
--- a/.changeset/fix-test-stability.md
+++ b/.changeset/fix-test-stability.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/agent-core": patch
+---
+
+Fix local agent typechecking when the source directory is passed as a relative path.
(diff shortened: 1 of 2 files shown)


UfukNode:
Looks like this was covered by #554, which also handles the Windows tsc shim case. Closing this to avoid keeping a conflicting duplicate open.

PR #553, opened 2026-08-06
fix(build-manifest): validate pause resumeStep against steps map

### LINKED ISSUE #552: bug(build-manifest): pause resumeStep not validated missing step silently passes graph check
### Summary

`validateGraph` / `buildForwardEdges` validates `continue` targets against the steps map and pushes an error for unknown ones, but does **not** do the same for `pause.resumeStep`. An agent with a `pause` that points to a non-existent step passes `check` and `deploy` without error, then fails at runtime.

### Reproduction

```ts
const wait = defineStep({
  name: 'wait',
  pause: { signal: 'x.done', resumeStep: 'nonExistentStep' }, // typo
  async run(_i) {
    return pauseUntilSignal({ signal: 'x.done', resumeStep: 'nonExistentStep' });
  },
});

export const agent = defineAgent({ name: 'test', entry: 'wait', steps: { wait } });
```

`npx @sapiom/cli agents check` → ✅ passes  
Runtime → 💥 crashes when signal arrives

### Root Cause

`packages/agent/src/build-manifest.ts`, `buildForwardEdges`:

```ts
// continue → error pushed ✅
if (t.kind === "continue") {
  if (names.has(t.target)) targets.add(t.target);
  else errors.push(`step '${name}' has a continue target '${t.target}' that is not in the steps map`);
}

// pause → NO error when resumeStep is missing ❌
} else if (t.kind === "pause" && names.has(t.resumeStep)) {
  targets.add(t.resumeStep);
}
```

### Fix

Add the same error for `pause` as exists for `continue`:

```ts
} else if (t.kind === "pause") {
  if (names.has(t.resumeStep)) targets.add(t.resumeStep);
  else errors.push(
    `step '${name}' has a pause resumeStep '${t.resumeStep}' that is not in the steps map`
  );
}
```

Sertug17:
Fixes #552

`buildForwardEdges` was validating `continue` targets but silently skipping invalid `pause.resumeStep` values. An agent with a misspelled resumeStep would pass `check` and `deploy`, then crash at runtime when the signal arrived.

This PR mirrors the same validation that already exists for `continue` targets.

### DIFF
--- a/packages/agent/src/build-manifest.ts
+++ b/packages/agent/src/build-manifest.ts
@@ -169,8 +169,12 @@ function buildForwardEdges(
           errors.push(
             `step '${name}' has a continue target '${t.target}' that is not in the steps map`,
           );
-      } else if (t.kind === "pause" && names.has(t.resumeStep)) {
-        targets.add(t.resumeStep);
+      } else if (t.kind === "pause") {
+        if (names.has(t.resumeStep)) targets.add(t.resumeStep);
+        else
+          errors.push(
+            `step '${name}' has a pause resumeStep '${t.resumeStep}' that is not in the steps map`,
+          );
       }
     }
     if (step.transitions.length === 0) {


ratataque (approved):
nice ty

PR #555, opened 2026-08-06
docs(contributing): fix clone URL, pnpm floor, and stale package list

memosr:
## Description

`CONTRIBUTING.md` has drifted from the repository. A contributor following
Getting Started either clones a URL that 404s or installs with a pnpm major
that cannot read the committed lockfile.

Four corrections:

- **Clone URL** pointed at `sdk.git`; the repository is `sapiom-js`.
- **pnpm floor** said 8.0.0, but `engines.pnpm` is `>=10.0.0` and
  `packageManager` pins `pnpm@10.34.3`. pnpm 8 fails against the pnpm 10
  lockfile that CI installs with `--frozen-lockfile`.
- **Project Structure** listed 5 of 18 packages, and 3 of those 5 (`axios`,
  `node-http`, `langchain`) carry a Deprecated banner in their own READMEs,
  with the langchain packages unpublished since #442. None of `agent`,
  `tools`, `cli`, `mcp`, `agent-core`, `agent-runtime` or the harness
  packages appeared, so the onboarding doc pointed contributors exclusively
  at legacy surface. Now lists all 18, grouped, deprecations flagged.
- **Worked examples** (`cd packages/core`, `pnpm --filter @sapiom/core …`)
  now use `@sapiom/agent`.

## Testing

- `prettier --check CONTRIBUTING.md` passes
- Every `@sapiom/*` name in the doc resolves to a real `packages/*/package.json`
  (18 cited, 18 on disk, no drift either direction)
- Each deprecation claim checked against the corresponding package README
- `packages/agent` verified to expose the `dev`, `test`, `test:watch`, `build`,
  `lint`, `format` scripts the doc now points contributors at

## Changeset

Omitted deliberately: no package source is touched. Matches #68 and #97, which
were also root docs-only changes shipped without one.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ Thank you for your interest in contributing to the Sapiom SDK! We welcome contri
 ### Prerequisites
 
 - Node.js 18.0.0 or higher
-- pnpm 8.0.0 or higher
+- pnpm 10.0.0 or higher (the repo pins `pnpm@10.34.3` via `packageManager`)
 
 ### Setup
 
@@ -15,8 +15,8 @@ Thank you for your interest in contributing to the Sapiom SDK! We welcome contri
 2. Clone your fork:
 
    ```bash
-   git clone https://github.com/YOUR_USERNAME/sdk.git
-   cd sdk
+   git clone https://github.com/YOUR_USERNAME/sapiom-js.git
+   cd sapiom-js
    ```
 
 3. Install dependencies:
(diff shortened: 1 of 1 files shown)


memosr:
Closing this: upstream rewrote CONTRIBUTING.md and three of the four corrections here are now fixed there independently. The clone URL points at sapiom-js, the pnpm floor reads 10.0.0 with the 10.34.3 pin noted, and the stale package list is gone in favour of pointing contributors at packages/ directly, which is a better answer than the enumerated list I proposed since it cannot drift again.

One small thing remains, but it is minor enough that I do not think it warrants keeping this open: the pnpm filter examples still use @sapiom/core. That package does exist, so nothing is broken, and the doc says to substitute the package you are changing. Happy to send a one-line PR if you would prefer the example point at a more representative package.

Thanks for the rewrite.

PR #559, opened 2026-08-07
fix(agent-core): install deps on scaffold + actionable Canvas bundle-error hint (SAP-2462)

gwitwer:
## Problem

A newly-created (or freshly-cloned) agent fails its very first Studio **Canvas render** with a raw esbuild wall — `Could not resolve "@sapiom/agent"` / `Could not resolve "zod"` — which the "Render failed" panel relays verbatim. Nothing is wrong with the agent; its dependencies were just never installed.

The Canvas extracts the step graph by **fully esbuild-bundling** `index.ts` from disk (`runManifestCheck` → `check()`, `bundle: true`, no `packages: 'external'`), so it must resolve `@sapiom/agent`/`zod` from the **project's own `node_modules`**. This auto-render fires unprompted on session boot and every source save. Two gaps let it reach users:

1. `scaffold()` never installs deps, yet the rail renders the project immediately. The one helper that did install (`tryInstallDependencies` in the harness `example-seed.ts`) is demo-only and no longer wired into the app.
2. The error surfaced is raw esbuild output, not the actionable "run `npm install`" cause — despite the render path being full of comments that already know this is the missing-install case.

Fixes SAP-2462.

## Changes

- **Install on create.** `scaffold()` gains an opt-in `installDependencies` flag (returns `dependenciesInstalled`); the `sapiom_dev_agents_scaffold` MCP tool — the Studio create path — passes it. Best-effort and non-fatal (missing/offline npm still yields a successful scaffold). `installProjectDependencies` now lives in `@sapiom/agent-core` and is shared with the example seed, removing the demo-only duplication.
- **Actionable error.** New `describeBundleFailure()` maps "no `node_modules` + unresolved import" to `Dependencies are not installed. Run \`npm install\` in <dir>, then try again.` (raw esbuild detail preserved). Wired into both `check()` and `loadDefinition()`. Every other bundle failure's message is unchanged.
- CLI `init` behavior unchanged (still prints `npm install` as an explicit next step); deploy path (`bundle.ts`, `packages: 'external'`) untouched.

## Tests

- `bundle-error.spec.ts` — maps the missing-deps case; leaves a genuine bad import / syntax error / non-Error untouched.
- `install-deps.spec.ts` — returns `false` (never throws) when the install can't run.
- Existing `scaffold.test.ts`, `example-seed.test.ts`, canvas render/graph, and MCP agents-tool suites pass. Full `@sapiom/agent-core` suite green (166); `@sapiom/mcp` green (97). Typecheck + lint clean across the three packages.

## Follow-up (not in this PR)

Optional hardening noted on SAP-2462: resolve the SDK from the harness's own `node_modules` via esbuild aliases so common-case extraction needs no project install at all (the manifest brand is a global `Symbol.for`).

### DIFF
--- a/.changeset/canvas-install-and-npm-hint.md
+++ b/.changeset/canvas-install-and-npm-hint.md
@@ -1,0 +1,14 @@
+---
+"@sapiom/agent-core": patch
+"@sapiom/mcp": patch
+"@sapiom/harness": patch
+---
+
+Install a new agent's dependencies on scaffold, and turn the Canvas's "Could not resolve …" render error into an actionable "run npm install" hint
+
+The Canvas step-graph extraction (`check` / `loadDefinition`) esbuild-bundles a project's `index.ts` resolving its imports — `@sapiom/agent`, `zod`, … — from the project's own `node_modules`. A newly-scaffolded (or freshly-cloned) agent whose deps were never installed therefore failed its very first, unprompted Canvas render with a raw esbuild wall (`Could not resolve "@sapiom/agent" … Could not resolve "zod"`), which the failure panel relayed verbatim.
+
+Two fixes:
+
+- `scaffold()` gains an opt-in `installDependencies` flag (returned as `dependenciesInstalled`), and the `sapiom_dev_agents_scaffold` MCP tool — the Studio's create path — now passes it, so a new agent opens with a working Canvas. Best-effort and non-fatal: a missing/offline npm still yields a successful scaffold. The `installProjectDependencies` helper (previously demo-only inside the harness's example seed) now lives in `@sapiom/agent-core` and is shared by both.
+- `check` and `loadDefinition` now route bundle failures through `describeBundleFailure`, which detects the "no `node_modules` + unresolved import" case and returns `Dependencies are not installed. Run \`npm install\` in <dir>, then try again.` (preserving the raw esbuild detail). Every other bundle failure's message is unchanged.
--- a/packages/agent-core/src/bundle-error.spec.ts
+++ b/packages/agent-core/src/bundle-error.spec.ts
@@ -1,0 +1,71 @@
+/**
+ * Unit tests for describeBundleFailure — the mapping that turns esbuild's raw
+ * "Could not resolve …" (by far the most common cause: a project whose deps
+ * were never installed) into a one-line "run npm install" instruction, while
+ * leaving every other bundle failure's message untouched.
+ */
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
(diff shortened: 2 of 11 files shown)


gwitwer:
**Review round 1** (via our CI review guidelines, run manually). SDK/backend-only change — no deployment risk (no env vars, migrations, or infra), design-system section N/A.

### Medium — `npm install` runs synchronously and blocks the MCP event loop for up to 120s

`installProjectDependencies` uses `execFileSync` (`install-deps.ts`), and the `sapiom_dev_agents_scaffold` MCP tool now calls `scaffold({ installDependencies: true })` on the interactive create path. `execFileSync` blocks the Node event loop for the entire install — up to the 120s `INSTALL_TIMEOUT_MS`. While it runs, the MCP server cannot service any other JSON-RPC message, honor a `notifications/cancelled`, or answer a ping — the whole server is frozen, not just that one call.

`scaffold()` is already `async`, so this is a clean swap: make `installProjectDependencies` async over `promisify(execFile)` (or `spawn`) and `await` it. The 120s timeout, the `false`-on-failure contract, and the Windows `shell` handling all carry over unchanged. The example-seed path (`installDependencies ?? true`) shares the same synchronous helper, so fixing it there fixes both callers.

Worth naming precisely because it's on the newly-added, on-demand, user-facing path (create-in-Studio) rather than the boot-time seed where a one-time stall is less visible.

### What checks out
- `describeBundleFailure` gating on `!existsSync(node_modules) && /Could not resolve/` correctly leaves genuine bad-import and syntax-error messages untouched (node_modules present, or non-resolution error) — verified against the three spec cases.
- `dependenciesInstalled` is returned all the way through the MCP tool result via `ok(await scaffold(...))`, so a caller can see a failed install rather than being misled by the tool description — good.
- `execFileSync` import in `example-seed.ts` is still live (`tryGit`), so removing `tryInstallDependencies` didn't orphan it.

**Verdict:** Solid, well-tested cleanup with a genuinely better error path. Non-blocking — recommend making the install async before merge; nothing else stands in the way.

gwitwer:
Addressed round 1 in af9d8387.

**Medium — sync install blocking the MCP event loop:** fixed. `installProjectDependencies` is now `async` over `promisify(execFile)` and both callers (`scaffold`, the harness example seed — both already `async`) `await` it, so the create path no longer freezes the MCP server for the duration of the install. The 120s timeout, `false`-on-failure contract, and Windows `shell` handling carry over unchanged; added a 10MB `maxBuffer` since async `execFile` buffers output (npm stays near-silent under `--loglevel=error`). Test updated to `await …resolves.toBe(false)`.

Thanks for the catch — agreed it mattered specifically because it's on the newly-added, on-demand, user-facing path rather than boot-time seeding.

PR #575, opened 2026-08-08
fix(agent): avoid mutating caller's def object in defineAgent

### LINKED ISSUE #572: bug(agent): defineAgent mutates the caller's def object in-place
### Summary

`defineAgent` modifies the object passed in by the caller: it reassigns `def.steps` when folding an agent-level `inputSchema`, and attaches `AGENT_DEFINITION_BRAND` via `Object.defineProperty` directly on `def`. Both mutations are visible on the caller's original reference.

### Root Cause

`packages/agent/src/agent.ts`:

```ts
// Mutation 1: reassigns def.steps on the CALLER'S object
(def as { steps: ... }).steps = {
  ...def.steps,
  [def.entry]: { ...entryStep, inputSchema: def.inputSchema },
};

// Mutation 2: brand attached to caller's object
Object.defineProperty(def, AGENT_DEFINITION_BRAND, { value: 1, enumerable: false });
```

### Reproduction

```ts
const myDef = {
  name: 'x', entry: 'start',
  inputSchema: z.object({ name: z.string() }),
  steps: { start: myStep },
};
const originalSteps = myDef.steps;
defineAgent(myDef);

console.log(myDef.steps === originalSteps); // false mutated
console.log(isAgentDefinition(myDef));      // true brand on original object
```

### Fix

Operate on a shallow copy instead of mutating the input:

```ts
const result = { ...def };
result.steps = { ...def.steps, [def.entry]: { ...entryStep, inputSchema: ... } };
Object.defineProperty(result, AGENT_DEFINITION_BRAND, { value: 1, enumerable: false });
return result;
```

Sertug17:
## Problem
`defineAgent` mutates the caller's `def` object in two ways:
1. Directly reassigns `def.steps` via a type cast when the entry step lacks an `inputSchema`
2. Attaches `AGENT_DEFINITION_BRAND` as a non-enumerable property on the original `def` reference

This corrupts a shared `steps` literal across multiple `defineAgent` calls and throws on `Object.freeze()`d inputs.

## Fix
- Replace the direct `.steps` assignment with `def = { ...def, steps: { ... } }` (rebinds local variable, original untouched)
- Create `const result = { ...def }` before attaching the brand symbol
- Return `result` instead of `def`

Fixes #572

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -185,21 +185,25 @@ export function defineAgent<TInput = unknown, TShared extends Record<string, unk
       // an `Object.freeze()`d map. The step's `run` and routing declarations are
       // carried over by the spread, so downstream readers of
       // `steps[entry].inputSchema` pick up the contract without special-casing.
-      (def as { steps: Record<string, StepDefinition<TShared>> }).steps = {
-        ...def.steps,
-        [def.entry]: { ...entryStep, inputSchema: def.inputSchema as ZodType<unknown> },
-      };
+      def = {
+        ...def,
+        steps: {
+          ...def.steps,
+          [def.entry]: { ...entryStep, inputSchema: def.inputSchema as ZodType<unknown> },
+        },
+      } as AgentDefinition<TInput, TShared>;
     }
   }
   // Attach the brand as a non-enumerable property so it:
   //   - survives esbuild bundling (symbol properties pass through)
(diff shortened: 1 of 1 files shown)


gwitwer:
Closing in favour of #585, which fixes the same defineAgent mutation with a single shallow copy, a regression test, and a changeset. Thanks for the report and fix.

PR #576, opened 2026-08-08
fix(scaffold): append to .npmrc instead of overwriting

### LINKED ISSUE #573: bug(scaffold): template .npmrc overwritten when non-default registry is configured
### Summary

`scaffold` renames `_npmrc` → `.npmrc` from the template (via `copyTemplate`), then unconditionally overwrites that `.npmrc` when a non-default npm registry is detected. Any content the template had in `_npmrc` is silently lost.

### Root Cause

`packages/agent-core/src/scaffold.ts`:

```ts
// Step 1: renames template's _npmrc → .npmrc
copyTemplate(templateDir, targetDir, replacements);

// Step 2: overwrites it entirely
if (sapiomRegistry !== DEFAULT_REGISTRY) {
  writeFileSync(
    path.join(targetDir, ".npmrc"),
    `@sapiom:registry=${sapiomRegistry}\n`, // ❌ template content gone
  );
}
```

`DOTFILE_NAMES` explicitly includes `"_npmrc"`, so templates are expected to carry npmrc content. For any developer using a custom/private registry (Verdaccio, private npm), that content is silently dropped.

### Fix

Append to the existing file instead of overwriting:

```ts
const npmrcPath = path.join(targetDir, ".npmrc");
const existing = existsSync(npmrcPath) ? readFileSync(npmrcPath, "utf8") : "";
const line = `@sapiom:registry=${sapiomRegistry}\n`;
if (!existing.includes(line)) {
  writeFileSync(npmrcPath, existing + line);
}
```

Sertug17:
## Problem
When `scaffold` detects a non-default `@sapiom` registry, it calls `writeFileSync` to write `.npmrc` unconditionally overwriting the file. If the template already copied a `_npmrc` → `.npmrc` with user-defined entries (e.g. a private registry for other scopes), those entries are silently erased.

## Fix
- Read the existing `.npmrc` content (or empty string if absent)
- Only append the `@sapiom:registry=` line if it isn't already present (idempotent)
- `existsSync` and `readFileSync` are already imported in the file

Fixes #573

### DIFF
--- a/packages/agent-core/src/scaffold.ts
+++ b/packages/agent-core/src/scaffold.ts
@@ -343,10 +343,12 @@ export async function scaffold(opts: ScaffoldOptions): Promise<ScaffoldResult> {
   if (!opts.versions) {
     const sapiomRegistry = registryFor("@sapiom/tools");
     if (sapiomRegistry !== DEFAULT_REGISTRY) {
-      writeFileSync(
-        path.join(targetDir, ".npmrc"),
-        `@sapiom:registry=${sapiomRegistry}\n`,
-      );
+      const npmrcPath = path.join(targetDir, ".npmrc");
+      const existing = existsSync(npmrcPath) ? readFileSync(npmrcPath, "utf8") : "";
+      const line = `@sapiom:registry=${sapiomRegistry}\n`;
+      if (!existing.includes(line)) {
+        writeFileSync(npmrcPath, existing + line);
+      }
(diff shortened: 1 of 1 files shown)


gwitwer:
Closing in favour of #628, which fixes the same .npmrc overwrite and adds a test, a changeset, and handling for a missing trailing newline. Thanks.

PR #594, opened 2026-08-11
fix(agent): drop nested defaulted fields from manifest schema required

erhnysr:
## Problem

`z.toJSONSchema()` (Zod v4) marks a field with a `default` as required, even though `z.object()` supplies that default on parse when the caller omits it. The AJV pre-check therefore rejects input that the subsequent Zod parse would have accepted.

`buildManifest` already compensates for this — but only at the top level. A defaulted field nested inside an object, inside `items`, or inside a `oneOf`/`anyOf`/`allOf` branch is still marked required, so omitting it fails the pre-check with a spurious "must have required property" error.

This is the same class of mismatch #83 fixed for `additionalProperties: false`, and for the same stated reason: the pre-gate must not be stricter than Zod. That normalization was made recursive; this one was left at the top level.

## Fix

`dropDefaultedFromRequired` now recurses, mirroring the traversal `relaxAdditionalProperties` already uses — nested object properties, `items`, and composition branches.

Applied in both places that carry the normalization:

- `packages/agent/src/build-manifest.ts` — so newly built manifests are correct
- `packages/agent-runtime/src/manifest-validation.ts` — so a manifest built by an older SDK is corrected at validation time

## Tests

`packages/agent/src/manifest.spec.ts` — a defaulted field nested in an object is no longer in that object's `required`. Verified failing-first: with the fix reverted, this test fails with `Expected value: not "tone" / Received array: ["tone"]`, which is exactly the reported behaviour.

`packages/agent-runtime/src/manifest-validation.spec.ts` — three cases covering the runtime path: the pre-check accepts input omitting a nested defaulted field, still rejects a genuinely missing field that has no default, and is a no-op on a null schema.

The "still rejects" case is the important one — it confirms the relaxation is narrow and doesn't weaken validation for fields that really are required.

Typecheck and lint pass for both packages.

## Changeset

`patch` for both `@sapiom/agent` and `@sapiom/agent-runtime`, matching the bump #83 used for the comparable fix.

## Formatting note

I matched each file's local style rather than running prettier — the repo isn't uniformly formatted (the two packages differ in quote style, and both touched files are already flagged by the root config), so reformatting would have buried the change in noise.

### DIFF
--- a/.changeset/agent-drop-nested-defaulted-required.md
+++ b/.changeset/agent-drop-nested-defaulted-required.md
@@ -1,0 +1,10 @@
+---
+"@sapiom/agent": patch
+"@sapiom/agent-runtime": patch
+---
+
+Generated step-input JSON Schemas no longer force a nested field with a Zod `default` to be required.
+
+`z.toJSONSchema()` (Zod v4) marks a defaulted field as required at every level, but a `z.object()` supplies the default on parse when the caller omits it. `buildManifest` already dropped such fields from the top-level `required`; it now does so at every level (nested objects, `items`, and `oneOf`/`anyOf`/`allOf` branches), matching how `additionalProperties: false` is already relaxed recursively. The runtime AJV pre-check applies the same relaxation, so a manifest built by an older SDK is corrected at validation time too.
+
+Additive and non-breaking. A step whose `inputSchema` nests a defaulted field now accepts input that omits it — the same input its Zod parse already accepted — instead of failing the pre-check with a spurious "must have required property" error.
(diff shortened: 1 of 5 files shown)


erhnysr:
Superseded by #852 (SAP-3218), which landed this fix upstream via the shared `normalizeInputJsonSchema` in `packages/agent/src/introspection.ts` — it drops `additionalProperties: false` and defaulted keys from `required` (nested included), and `build-manifest.ts`/`manifest-validation.ts` now route through it. Closing this in favor of that implementation.

PR #627, opened 2026-08-13
fix(core): setHeader/removeHeader no longer coerce undefined headers to empty string

### LINKED ISSUE #625: [Bug]: core header helpers convert undefined values into empty headers
### Affected package or area

@sapiom/core header utilities (`packages/core/src/utils/utils.ts`)

Current `main` at `sapiom/sapiom-js` as of 2026-08-13.

### Environment

Node.js: 20+ / 22+
Package manager and version: pnpm 10.34.3
Operating system: platform-independent
Runtime or framework: header maps passed through SDK HTTP integrations

The helpers accept `Record<string, string | string[] | undefined>`, where `undefined` represents an omitted header in Node/HTTP-style maps.

```ts
import { setHeader, removeHeader } from "@sapiom/core";

const headers = {
  Authorization: undefined,
  "Content-Type": "text/plain",
};

setHeader(headers, "X-Sapiom-Transaction-Id", "tx_123");
removeHeader(headers, "Content-Type");
```

### Expected behavior

Unrelated headers with `undefined` values should remain omitted when copying headers into the returned object. The returned header map should not contain `Authorization` in the example above.

### Actual behavior

Both `setHeader` and `removeHeader` copy unrelated headers with this expression:

```ts
newHeaders[key] = Array.isArray(val) ? val[0] || "" : val || "";
```

That coerces `undefined` to `""`, turning an absent header into a present-but-empty header. Downstream HTTP clients can then send an empty `Authorization` or similar header instead of omitting it.

This is separate from #10, which normalized a completely missing `request.headers` object in `@sapiom/node-http`. This report is about individual `undefined` values inside an existing header object in `@sapiom/core` utilities. A focused fix would skip `val === undefined` while copying non-target headers and add regression tests in `packages/core/src/utils/utils.test.ts`.

### Acknowledgements

- [x] This report does not describe a suspected security vulnerability; I will use the Security Policy for private reporting.

ygd58:
Fixes #625

Both `setHeader` and `removeHeader` accept `Record<string, string | string[] | undefined>`, where `undefined` represents an omitted header in Node/HTTP-style maps. The copy loop used `val || ""` as a fallback, which turned any `undefined` value into an empty string, so an absent header (e.g. `Authorization: undefined`) came back as present-but-empty. Downstream HTTP clients could then send an empty header instead of omitting it.

**Fix:** unrelated headers with an `undefined` value are now skipped while copying, keeping them omitted in the returned map, matching the documented type.

**Tests:** added regression tests for both `setHeader` and `removeHeader`, per the issue's suggested fix (`packages/core/src/utils/utils.test.ts`).

**Changeset:** added (`@sapiom/core`, patch).

Verified locally: `pnpm --filter @sapiom/core test` (134/134 passing), `pnpm --filter @sapiom/core typecheck` clean, `eslint` clean on both changed files.

### DIFF
--- a/.changeset/header-utils-undefined-values.md
+++ b/.changeset/header-utils-undefined-values.md
@@ -1,0 +1,16 @@
+---
+"@sapiom/core": patch
+---
+
+Fix `setHeader` and `removeHeader` coercing unrelated `undefined`-valued
+headers into empty strings
+
+Both helpers accept `Record<string, string | string[] | undefined>`, where
+`undefined` represents an omitted header in Node/HTTP-style maps. When copying
+unrelated headers, the previous `val || ""` fallback turned any `undefined`
+value into `""`, so an absent header (e.g. `Authorization: undefined`) came
+back as a present-but-empty header. Downstream HTTP clients could then send an
+empty header instead of omitting it.
+
+Unrelated headers with an `undefined` value are now skipped while copying, so
+they stay omitted in the returned header map, matching the documented type.
--- a/packages/core/src/utils/utils.test.ts
+++ b/packages/core/src/utils/utils.test.ts
@@ -71,6 +71,20 @@ describe("Header Utilities", () => {
       expect(result["Authorization"]).toBe("old"); // Other headers preserved
     });
 
+    it("should keep unrelated undefined-valued headers omitted, not coerce to empty string", () => {
+      const headers = {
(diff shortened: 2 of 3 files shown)


gwitwer:
Closing in favour of #626, which fixes the same undefined-header coercion and was opened first. Both had tests and changesets. Thanks.

PR #630, opened 2026-08-13
feat(harness): finish Studio PostHog instrumentation — attribution, run funnel, redaction

gwitwer:
## Summary

The Studio has captured clicks since SAP-1988, but not legibly. Autocapture is on and working (~6.4k clicks / 81 people in 30d), yet a third of those clicks arrive with no label, the rest have no surface attribution, and the run funnel — the Studio's core action — was never wired. This closes those gaps.

**Review round 1 found that the redaction this PR added was itself leaky, and untested in the way that mattered.** Every finding reproduced; all are fixed in `91e9c90e`. The headline change from that round: element attributes are now an **allowlist**, so a surface nobody tagged is safe by default rather than leaky by default. Details in "What review changed" below.

## What was actually wrong

Verified against production PostHog before writing any code:

| Gap | Evidence |
|---|---|
| `trackingAttrs`/`TrackScope` spread on **zero** components | built + tested + documented in SAP-1988, never applied |
| ~34% of clicks have no `$el_text` | 1,931 clicks + 231 changes of 6,444, across 71 people |
| `harness_host` absent from the taxonomy | declared in `AppContextProperties`, never registered |
| Run funnel never fired | 3 registry entries, 0 call sites |
| `isSecretSurface()` unreachable | it reads `surface`/`object` from `trackingAttrs`, applied nowhere |
| User content in `$el_text` | `fetch-recent-weather`, `newsletter-autopilot`, `Ewan's Organization` |
| Heatmaps starved | random port per boot ⇒ every session is its own URL |

### Click attribution
- `surface` on the top-level regions; `object` on entity rows.
- **New `dialog` dimension.** posthog-js merges capture attributes walking *outward*, so the outermost element wins — verified in `dist/module.no-external.js`, not taken from the doc comment. `StartDialog` is a DOM child of the agent rail, so a `surface` on it would be silently discarded. `dialog` composes: you get "the add-agents dialog" **and** "opened from the rail".
- `TemplatesPanel` carries no panel-level `surface` — it hosts two mutually exclusive modes and an outer surface would collapse gallery and detail into one bucket.

### Icon-only buttons
`before-send` promotes the nearest `aria-label` into `$el_text` when there is no text, tagged `el_text_source: "aria_label"`. One rule instead of ~30 hand-added attributes.

### Session/host context
`harness_host` (`cli`|`desktop`), `studio_boot_id` (per SPA load — PostHog's `$session_id` can't answer this, since the random port makes every boot a separate origin with its own `localStorage`), and `harness_kind` for the active session. Registered before the first `$pageview`, and re-registered after a sign-out `reset()` clears super properties.

### Run funnel
`agent.run_started` / `run_succeeded` / `run_failed` at both choke points — `startRunPolling` (prod, button or CLI-detected) and the local-run stream. Carries `target`, `duration_ms`, message-free `error_kind`.

Judgement calls, each commented at the call site: `unobservable` is its own `error_kind` so old servers don't inflate the failure rate; a `paused` local run emits nothing (still alive); a superseded poller emits nothing (still running server-side, we just stopped watching).

### Privacy
Allowlist attributes (below), `drop_name` mode for user-authored labels, escape-aware chain regexes, and the port collapse so URL breakdowns aggregate.

## What review changed (`91e9c90e`)

**1. Attributes → allowlist.** The denylist was wrong on arrival: `data-testid={`workflow-${workflow.name}`}` sits on the very element carrying `object: "agent"`, and `title={isDirectory ? cwd : label}` ships the absolute path including the OS username. So the flagship path leaked the agent name on every rail click while its test asserted `attr__class` survived and never checked the testid. Now only `class`/`id` survive, in **every** mode including `truncate`, in both carriers. `data-testid` is dropped rather than special-cased — interpolated with names today, and a poor Actions selector regardless. Match on `class`.

**2. Tagged the surfaces promotion was unsafe on.** An `aria-label` is not automatically safe: `Copy path for ${label}`, `Focus ${label}`, `More actions for ${detailStep.label}`, `${field.name}, ${field.type}` all interpolate user data, and promotion copied them into `$el_text`. `workspace` and `run` were in `USER_NAMED_OBJECTS` and applied nowhere — now applied, plus the canvas actions header (**both** render branches), terminal masthead, session rows and bar, canvas field lists.

**3. The gate.** `redaction-gate.test.ts` asserts on whole payloads with a needle rather than on fields a test author remembered to check — the exact failure mode that let the original leak through. It caught `TemplateCard` on first run (safe — gallery names are ours — now recorded with a reason).

**4. Two run-funnel logic bugs.** The redelivery guard seeded `outcomeEmitted` from the dedupe flag, inverting its own invariant: a replayed `execution.started` clears the old interval, so the new poller was the only live one and started latched shut — one `run_started`, zero terminals, forever. Now two independent latches. And a single failed poll declared `run_failed` on a live run; `/api/runs/:id/state` returns 503 whenever the harness holds no credential while the coding agent runs with its own, so it was reachable in normal use. Now tolerates 5 consecutive failures (~10s), which is also what makes `unobservable` honestly mean "old server".

**5. Smaller.** Escape-aware chain regexes (posthog writes `\"`, so `[^"]*` left the tail of a quoted name behind); pane-level `object` moved down in `DeadSessionPane`/`DirectoryPicker`, where it was blanking every button label and reintroducing the blank-row problem; `announcedRuns` bounded; `harness_kind` typed as `HarnessKind`; `journey` doc no longer contradicts `CommandPalette`.

**6. Heatmap claim corrected, not defended.** The PostHog toolbar queries by the page's real URL, which is still the ephemeral-port one — so the port collapse fixes breakdowns for certain and heatmaps only maybe. Says so at the call site now.

### Known limit of the static gate

The static half is **file-level**, and that's a real limit: a component with two render branches passes as soon as one tags an object. That case existed here — `WorkflowActionsHeader`'s detail branch was still leaking after I tagged its steps branch. Fixed, but the check would have waved it through, so the test documents what it cannot see. The durable version needs a rendered DOM and belongs in `web/e2e`; this runner has none by design (see `vitest.config.ts`).

## Deliberately NOT in scope

- **`secrets.*`** — the Studio has no secrets panel. Every set-a-secret flow is in the coding agent's pty or on the dashboard. Now documented as blocked at the registry rather than reading like a working metric.
- **`agent.turn_stuck`** — needs live per-tool-call outcomes; `BusMessage` carries nothing tool-shaped.
- **Session recording** — still `disable_session_recording: true`. Needs its own consent tier and a harder masking posture; separate decision.

## Breaking-ish

- `$current_url`/`$host` change shape (`127.0.0.1:57070` → `localhost`). Insights filtering on a specific port stop matching — none should, the port was random.
- `attr__data-testid` and all other non-`class`/`id` attributes no longer reach PostHog. Any Action matching on a testid needs re-pointing at `class`.
- **`attr__href` is dropped, not sanitized.** Previously it flowed through `sanitizeUrl`; the allowlist now removes it in every mode, so autocapture no longer records *where* a link click went. `$el_text` still carries the visible label on `truncate` surfaces, so most link analysis survives. Verified nothing regresses: the only saved insight mentioning `href` is a web-app one on `/settings/plans` whose own description already reads "Always 0 today".

## Testing

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean (one pre-existing warning in untouched `rest.test.ts`)
- [x] `vitest run` — **2067 passed / 140 files**, incl. 100 analytics cases
- [x] `pnpm build:web` succeeds
- [ ] DOM-level redaction gate in `web/e2e` — follow-up, see "Known limit"
- [x] run-funnel logic pinned by `runFunnelStep` unit tests (mutation-tested: 4/4 reverts caught)
- [ ] e2e coverage for the run funnel — follow-up

New coverage: whole-payload needle tests per name-bearing surface (agent row, workspace row with absolute path, icon-only copy-path button, canvas header, terminal masthead, session bar, directory row), the untagged-surface case proving the allowlist holds with no tag, quote-escaping, port collapse, aria promotion (nearest-first, never overwriting real text, suppressed on secret/user-named), `harness_host` both ways, boot-id stability across sign-out, register-before-pageview ordering, and the run/local outcome classifiers.

## Related

- Refs: SAP-2117
- Refs: SAP-1988

## Checklist
- [x] All tests passing
- [x] No hardcoded secrets
- [x] Self-reviewed

https://claude.ai/code/session_0138o3fwi1KqMJSvuLLk2uMZ


gwitwer (commented):
Reviewed with our CI bot review instructions, plus an independent pass. Core assumptions in the PR body were **verified against the vendored `posthog-js@1.409.5` bundle** and hold: the outermost element does win a capture-attribute key conflict, the chain is serialized nearest-first (both nearest-first comments are correct), `attr__aria-labelledby`/`aria-hidden` can't match the aria-label regex, and `TrackScope`'s `display: contents` breaks no CSS. The `dialog` dimension works — `trackingAttrs` is generic over keys. CI is 9/9 green.

The architecture is genuinely good: declarative ancestor-inherited attribution, one `before-send` rule instead of ~30 hand-kept attributes, and the blocked-registry-entry comments are the right fix for "an entry with no emitter reads like a working metric."

**But the safety property has the same trap one layer down, and it isn't merely ungated — it's untested.** Redaction now depends on: the right `object` value, on the right ancestor, at the right depth, with no attribute anywhere in the subtree interpolating a user string. Three of those four are already violated in this PR, on the surfaces it was written to fix, with all 33 new tests passing green.

Blocking (inline):

1. **`attr__data-testid` ships the exact names `drop_name` exists to redact** — on the very element carrying `object: "agent"`. The flagship redaction path leaks, and its test can't see it.
2. **The aria-label promotion copies user names into `$el_text`** on every untagged surface — including two beyond the rail (`WorkflowActionsHeader`, `CanvasStepDetail`).
3. **`workspace` and `run` are in `USER_NAMED_OBJECTS` but never applied** — workspace rows ship the full absolute path, OS username included, via `attr__title`.
4. **The redelivery guard suppresses the terminal event rather than the duplicate start** — its own scenario yields a `run_started` with no terminal, ever.
5. **A single failed poll fabricates `run_failed`** for a still-running run; the 503 "not signed in" path makes this reachable in normal use, not just on old servers.

Lower severity inline: escaped quotes defeat `[^"]*` in `drop_name`; pane-level `object` blanks every button inside `DeadSessionPane`/`DirectoryPicker`; the heatmap claim needs one real boot to confirm.

Two structural changes close 1, 2, 3, and the quote bug together, and turn the rest into ordinary bugs:

- **Invert element-attribute handling to an allowlist** — keep `class`/`id`, drop every other `attr__*` unconditionally. A missing `object` tag becomes safe-by-default instead of leaky-by-default.
- **Add one gate** — render the rail, session bar, terminal and canvas header with a fixture path, assert the payload contains no `/Users/` and no fixture name. One test, catches the whole class.

Minor, non-blocking: `announcedRuns` grows unbounded for the page's life; `harness_kind?: string` vs `harness_host?: HarnessHost` is inconsistent for two adjacent super-properties; and `TrackingContext.journey`'s doc says to set it explicitly for "a global modal, a command palette" while `CommandPalette` now comments that it deliberately doesn't — one of the two is wrong.

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -67,26 +140,62 @@ export function sanitizeUrl(rawUrl: unknown): unknown {
...
     delete element.$el_text;
     for (const key of Object.keys(element)) {
       if (key.startsWith("attr__") && !KEPT_ELEMENT_ATTRIBUTES.has(key)) delete element[key];
     }
     return;
   }
-  if (typeof element.$el_text === "string") {
+  if (mode === "drop_name") {
+    delete element.$el_text;
+    for (const key of Object.keys(element)) {
```
gwitwer: **`drop_name` keeps `attr__data-testid`, but this codebase's testids interpolate the exact names being redacted.**

The comment above says `class`/`id`/`data-testid` survive "because they are ours." They aren't — they're interpolated with user content:

- `WorkflowRow.tsx:45` — `data-testid={`workflow-${workflow.name}`}` on the *same element* that carries `object: "agent"`
- `WorkflowRow.tsx:64` — `workflow-status-${workflow.name}`
- `DirectoryPicker.tsx:311` — `dir-picker-item-${entry.name}` under `object: "directory"`

posthog-js captures every attribute as `attr__*` (verified in `posthog-js@1.409.5/dist/module.no-external.js`: `r.attributes["attr__"+i.name]`, no default ignore list). So after `$el_text` and `attr__aria-label` are dropped, a click on an agent row still ships `attr__data-testid="workflow-fetch-recent-weather"` in **both** `$elements` and `$elements_chain`.

The new test asserts `attr__class` survives and never checks testid — so it passes green while the name still leaves the machine.

Suggest inverting this to a strict allowlist (keep `class`/`id`, drop every other `attr__*` unconditionally in every mode). That makes a missing `object` tag safe-by-default rather than leaky-by-default, and also closes the `attr__title` / `attr__data-tooltip` holes flagged elsewhere in this review. The cost is interpolated testids as PostHog Actions selectors — worth deciding explicitly, though names-in-testids are poor selectors regardless.

packages/harness/web/src/components/WorkflowRow.tsx
```diff
@@ -42,6 +43,10 @@ export function WorkflowRow({
     <div
       className={"workflow-item" + (isFocused ? " is-focused" : "")}
       data-testid={`workflow-${workflow.name}`}
+      // `object: "agent"` is load-bearing for privacy, not just for breakdowns:
+      // before-send reads it to drop $el_text on this subtree, because an
+      // agent's label is a name its owner wrote. See USER_NAMED_OBJECTS.
+      {...trackingAttrs({ object: "agent" })}
```
gwitwer, line 49: The concrete instance of the `data-testid` leak (see the comment on `before-send.ts` `scrubElement`).

This comment says `object: "agent"` is "load-bearing for privacy" — but line 45 on this very element is `data-testid={`workflow-${workflow.name}`}`, and `drop_name` explicitly preserves `attr__data-testid`. So the agent name still ships on every click here, in both element carriers.

`title={workflow.name}` at :57 *is* correctly dropped (`title` is in `NAME_BEARING_ATTRIBUTE_NAMES`) — it's only the testid that escapes.

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -128,37 +244,101 @@ function scrubElementCarriers(properties: Record<string, unknown>, isSecret: boo
...
+ * So: when there is no text, promote the nearest ancestor's accessible name
+ * into `$el_text`. One rule fixes every such control, including ones nobody
+ * has written yet, and it reuses labels that are already maintained because
+ * screen readers depend on them — strictly better than a hand-kept list of
+ * per-button tracking attributes that silently rots.
+ *
+ * Only runs in `truncate` mode: on a user-named or secret surface the whole
+ * point is that the label goes away.
+ */
+function promoteAccessibleName(properties: Record<string, unknown>): void {
```
gwitwer, line 314: **This copies user-authored names into `$el_text` — the property this PR just cleaned.**

Promotion runs whenever mode is `truncate`, i.e. everywhere no `object` marker exists. That includes every icon button whose `aria-label` interpolates user data:

- `WorkflowsRail.tsx:166/227` — `Copy path for ${label}` (label = user's folder basename)
- `WorkflowsRail.tsx:208/219/264` — `Focus ${label}`, `Scaffold an agent in ${label}`
- `WorkflowActionsHeader.tsx:107` — `More actions for ${detailStep.label}` (step name from the user's own agent code)
- `CanvasStepDetail.tsx:575/1049` — `${f.name}, ${f.type}` (schema field names)

Concretely: user clicks the copy-path icon on their `acme-client-secret` workspace row, and `$el_text` becomes `"Copy path for acme-client-secret"` — user content in the primary breakdown dimension, shredding cardinality exactly the way the PR body says agent names did. `el_text_source: "aria_label"` doesn't mitigate it.

The rule itself is the right call; it just needs the surfaces below to be tagged first (see the `USER_NAMED_OBJECTS` comment).

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -53,11 +56,81 @@ function isSecretSurface(properties: Record<string, unknown>): boolean {
...
+ * `fetch-recent-weather`, `newsletter-autopilot`, `twitter-run` and
+ * `Ewan's Organization` this way. That is wrong twice: it puts user-authored
+ * strings in analytics, and it shreds the numbers, because "clicked an agent in
+ * the rail" splits into one row per private agent name and never aggregates.
+ *
+ * Templates are deliberately absent — a gallery template's name is OUR string
+ * from OUR registry, low-cardinality and safe, and it's the label that makes
+ * the on-ramp funnel readable.
+ */
+const USER_NAMED_OBJECTS: ReadonlySet<string> = new Set(["agent", "workspace", "session", "run", "directory"]);
```
gwitwer, line 100: **`workspace` and `run` are listed here but no component ever sets either.**

Grepping the head commit, only `agent`, `directory`, `session`, `snippet` and `template` are ever passed to `trackingAttrs`. So the rail's workspace rows — the second-largest source of user-authored strings in the rail — get no redaction at all and stay in `truncate` mode:

- `WorkflowsRail.tsx:158` — `$el_text` = folder basename
- `WorkflowsRail.tsx:154` — `title={isDirectory ? cwd : label}` → `attr__title` = the **full absolute path, including the OS username**
- `WorkflowsRail.tsx:325` — `title={cwd}` on session dropdown items
- `WorkflowsRail.tsx:771` — `title={row.session.title}` (prompt text)
- `SessionBar.tsx:213` — `data-tooltip={… · ${activeSession.cwd}}` (`data-tooltip` is *in* `NAME_BEARING_ATTRIBUTE_NAMES`, so it's known name-bearing — it just never fires without an `object`)
- `TerminalBrand.tsx:46` — `title={cwd}` under `surface: "terminal"`
- `WorkflowActionsHeader.tsx:164` — `title={workflow.name}`, rendered inside CanvasPane/CodePanel, neither of which sets `object: "agent"` — so agent names still ship from the canvas header, which is the leak the PR body says it closes

The machinery all exists; it just never fires on these surfaces. This is the finding that most argues for the allowlist flip — tagging seven more components by hand leaves the eighth to whoever writes it next.

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -53,11 +56,81 @@ function isSecretSurface(properties: Record<string, unknown>): boolean {
...
+
+/** Whether this click's `object` marks it as a user-named entity. */
+function isUserNamedObject(properties: Record<string, unknown>): boolean {
+  const object = typeof properties.object === "string" ? properties.object.toLowerCase() : "";
+  return USER_NAMED_OBJECTS.has(object);
+}
+
+/** `text="…"` segments inside the serialized `$elements_chain` string. */
+const CHAIN_TEXT_PATTERN = /\btext="[^"]*"/gi;
```
gwitwer: `[^"]*` doesn't survive posthog's quote escaping.

The chain escapes embedded quotes as `\"` — confirmed in the vendored bundle: `function Ws(e){return e.replace(/"|\\"/g,'\\"')}`. For an agent or folder named `my "cool" agent`, `text="my \"cool\" agent"` matches only up to the first escaped quote, leaving `cool\" agent"` behind in `$elements_chain` and the chain malformed.

Low severity (needs a quote in the name), but it's a partial leak in `drop_name` — the one mode whose entire job is leaving nothing. `NAME_BEARING_ATTRIBUTES` has the same shape, and `CHAIN_HREF_PATTERN` above does too (pre-existing).

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -67,26 +140,62 @@ export function sanitizeUrl(rawUrl: unknown): unknown {
   } catch {
     return rawUrl;
   }
+  const origin = LOOPBACK_HOSTNAMES.has(url.hostname)
```
gwitwer, line 150: Worth verifying against a real boot before treating this as the fix for the heatmaps bullet.

Rewriting `$current_url` to `http://localhost/` stores clickmap data under a URL that no browser session ever actually has — the real page is `http://127.0.0.1:57070/`. The PostHog toolbar queries heatmaps by the URL of the page it's loaded on, so the aggregated data may now accumulate correctly and still be unreachable from the Studio itself, i.e. `enable_heatmaps` could still appear to produce nothing.

The port collapse is clearly right for `$current_url` breakdowns either way — this is only about the heatmap claim specifically.

packages/harness/web/src/lib/use-harness-state.ts
```diff
@@ -540,9 +557,49 @@ export function useHarnessState(): HarnessStateHook {
...
+      const runSlug = workflowPath ? slugFromPath(workflowPath) : undefined;
+      const runBase = { workflow_slug: runSlug, session_id: sessionId, target };
+      // A repeat announcement re-polls but must not re-count. See announcedRuns.
+      const alreadyAnnounced = announcedRuns.current.has(executionId);
+      announcedRuns.current.add(executionId);
+      // Exactly one terminal event per run. Polls are async and the immediate
+      // `void poll()` overlaps the interval, so two in-flight polls can both
+      // observe the terminal status before either clears the timer; without
+      // this latch that run counts twice.
+      let outcomeEmitted = alreadyAnnounced;
```
gwitwer: **This suppresses the terminal event, not the duplicate start — inverting the invariant it documents.**

On a second `execution.started` for the same `executionId`, line 537 has just cleared the previous interval, so the newly-created poller is the only live one — and it starts latched shut, so it can never call `agent.run_succeeded` / `run_failed`.

Net result for the guard's own scenario (a replayed bus message, or `ExecutionDetector` matching twice): one `agent.run_started`, zero terminal events, for a run that completed normally. That's the opposite of the "every `run_started` should be followed by exactly one terminal event" rule stated at `lifecycle.ts:1214`, and it's a different case from the superseded-poller exception documented at :532 (that one is a *new* run on the same session).

Needs two latches: dedupe `run_started` per `executionId`, but start `outcomeEmitted` at `false` unconditionally.

packages/harness/web/src/lib/use-harness-state.ts
```diff
@@ -552,13 +609,15 @@ export function useHarnessState(): HarnessStateHook {
...
           if (run.status !== "running") {
+            emitRunOutcome(run);
             const timer = runPollers.current.get(sessionId);
             if (timer) clearInterval(timer);
             runPollers.current.delete(sessionId);
           }
         } catch {
           // Endpoint absent (server predates runtime analytics) or transient
           // failure: stop polling quietly - the UI simply shows no run state.
+          emitRunOutcome(null);
```
gwitwer: **A single failed poll records a fabricated `agent.run_failed` for a run that's still executing.**

This `catch` fires on *any* error and stops polling immediately. The comment treats that as "endpoint absent," but transient failures reach it too — and one specific case isn't transient or rare: `core/run-state.ts:142` returns `503 "harness is not signed in to Sapiom"`, reachable whenever the coding agent launches a run from the pty with its own key while the harness itself holds none, and `ExecutionDetector` still announces it on the bus. `:110` returns `502 "gateway unreachable"` similarly.

The run completes fine server-side, but the funnel books `run_failed` with `error_kind: "unobservable"` or `"exception"`. Because polling gives up after the *first* error, a single 2-second blip mid-run is enough to log a false failure.

Suggest tolerating N consecutive failures before declaring an outcome — the `unobservable` bucket only means "old server" if we actually retried.

packages/harness/web/src/components/DeadSessionPane.tsx
```diff
@@ -92,7 +93,12 @@ export function DeadSessionPane({
   const canContinue = !canResume && record.status === "ready";
 
   return (
-    <div className="dead-session-pane" data-testid="dead-session-pane" data-has-record={showRecord}>
+    <div
+      className="dead-session-pane"
+      data-testid="dead-session-pane"
+      data-has-record={showRecord}
+      {...trackingAttrs({ surface: "session_history", object: "session" })}
```
gwitwer: Putting `object: "session"` on the whole pane blanks the labels of every control inside it.

`drop_name` deletes `$el_text` and skips aria promotion for the entire subtree, so clicks on `Continue`, `Resume`, `Close` and `Start` (:135/139/149/237) all arrive with no `$el_text` — reintroducing, on this pane, exactly the blank-row problem this PR set out to fix. `DirectoryPicker.tsx:214` has the same shape for its Browse / up-a-level controls.

The name-bearing element is the session row/title, not the pane. Moving `object` down to it keeps the redaction and keeps the buttons legible.

gwitwer:
All five blockers reproduced before I changed anything, and all five are fixed in `91e9c90e`. Point by point:

**1. `attr__data-testid` leaked the names `drop_name` exists to redact.** Confirmed — `data-testid={`workflow-${workflow.name}`}` on the same element carrying `object: "agent"`. Taken your suggested fix: **element attributes are now a strict allowlist** (`class`/`id` only), applied in *every* mode including `truncate`, in both carriers. `data-testid` is dropped rather than special-cased, on your reasoning that a name-bearing testid is a poor Actions selector anyway. That flips the default the right way — a surface nobody tagged is now safe, not leaky, which also closes the `attr__title` / `attr__data-tooltip` holes and makes the quote bug unreachable for attributes.

**2. Promotion copied user names into `$el_text`.** Confirmed, on all four sites you listed. Tagged the surfaces: `WorkflowsRail` (workspace rows), `WorkflowActionsHeader`, `CanvasStepDetail` field lists, `TerminalBrand`, `SessionBar`, session dropdown rows.

**3. `workspace` and `run` declared but never applied.** Confirmed — `object:` only ever appeared as `agent`/`directory`/`session`/`snippet`/`template`. Both are applied now. `WorkflowsRail.tsx:154`'s `title={isDirectory ? cwd : label}` was the worst of it: full absolute path, OS username included.

**4. Redelivery guard suppressed the terminal, not the duplicate start.** Confirmed and it's exactly the inversion you describe. Two independent latches now: dedupe `run_started` per `executionId`, `outcomeEmitted` starts `false` unconditionally.

**5. Single failed poll fabricated `run_failed`.** Confirmed, and the 503 path makes it reachable in normal use. Now tolerates 5 consecutive failures (~10s at the 2s interval). That also makes `unobservable` honest — it only means "old server" if we actually retried.

**Lower severity:** escaped quotes — fixed, both `CHAIN_TEXT_PATTERN` and the new attribute pattern use `(?:\\.|[^"\\])*`. Pane-level `object` — moved down to the summary block in `DeadSessionPane` and to the listing rows in `DirectoryPicker`; you're right that it was reintroducing the blank-row problem on those panes. Heatmaps — I've corrected the claim rather than defended it; the toolbar queries by the page's real URL, so the port collapse fixes breakdowns for certain and heatmaps only maybe. Now says so at the call site.

**Minor:** `announcedRuns` is bounded (500, oldest-out). `harness_kind` is `HarnessKind`, matching `harness_host`. The `journey` doc/`CommandPalette` contradiction — the doc was the wrong one: a navigator has no journey of its own, so the ambient value (which journey you reached for it *from*) is the more useful fact. Doc now says that and names the palette as the counter-example.

---

**Two things back:**

**Your gate suggestion caught a component in its first run** — `TemplateCard` interpolates `template.name` into an aria-label and I'd missed the file entirely. Safe (gallery names are ours, not the user's), now in the allowlist with the reason written down rather than silently passing.

**Your file-level framing has a hole, and I hit it.** `WorkflowActionsHeader` has two render branches. I tagged the steps branch; the static check went green because the *file* contained an `object:` — while the detail branch was still shipping a step name and `More actions for ${detailStep.label}`. Tagged now, but the check as specified would have waved it through, so I've documented that limit in the test instead of letting it read as coverage. The honest version needs a rendered DOM and belongs in `web/e2e` — this runner has none by design (`vitest.config.ts` is explicit that React is the Playwright tier's job). Filed as a follow-up in the PR body rather than bolted on here.

What did land: `redaction-gate.test.ts` asserts on **whole payloads with a needle** per surface, rather than on fields a test author remembered to check — which was the actual failure mode. Includes an untagged-surface case that proves the allowlist holds with no tag at all.

2032 tests / 137 files green, typecheck and lint clean, web build fine.

packages/harness/web/src/lib/use-harness-state.ts
```diff
@@ -540,9 +568,77 @@ export function useHarnessState(): HarnessStateHook {
...
+      if (announcedRuns.current.size >= ANNOUNCED_RUNS_CAP) {
+        const oldest = announcedRuns.current.values().next().value;
+        if (oldest !== undefined) announcedRuns.current.delete(oldest);
+      }
+      announcedRuns.current.add(executionId);
+      // Exactly one terminal event per run. Polls are async and the immediate
+      // `void poll()` overlaps the interval, so two in-flight polls can both
+      // observe the terminal status before either clears the timer; without
+      // this latch that run counts twice.
+      let outcomeEmitted = false;
```
gwitwer: **This fix is correct, and nothing tests it.**

I mutation-tested it: reverting this line to `let outcomeEmitted = alreadyAnnounced;` reintroduces the exact bug from the last round (one `run_started`, zero terminals, forever) and **all 617 tests still pass**.

That is the same shape as the leak this PR just fixed: a real invariant, documented in a careful comment, with nothing holding it in place. The comment even states the rule ("Dedupe the start; leave the outcome free to fire exactly once") — which is precisely the kind of claim that rots silently.

The redaction half now has a gate that genuinely bites. The run-funnel half has none, and it is the half with two logic bugs found in one review round.

`startRunPolling` is not trivially testable as-is, but the classifier boundary is: extracting the latch/failure-count decision into a pure helper next to `runErrorKind` in `lifecycle.ts` — something like `runFunnelStep(state, event) -> "start" | "terminal" | "none"` — would make both this and the retry threshold unit-testable without a React harness.

packages/harness/web/src/lib/use-harness-state.ts
```diff
@@ -552,16 +648,19 @@ export function useHarnessState(): HarnessStateHook {
...
-          // failure: stop polling quietly - the UI simply shows no run state.
-          const timer = runPollers.current.get(sessionId);
-          if (timer) clearInterval(timer);
-          runPollers.current.delete(sessionId);
+          // failure. Retry a few times before concluding anything: a single
+          // 2-second blip is not a failed run, and on a server that genuinely
+          // lacks the endpoint every attempt fails anyway, so the only cost of
+          // retrying is a few seconds before `unobservable` is recorded.
+          consecutiveFailures += 1;
+          if (consecutiveFailures < MAX_RUN_POLL_FAILURES) return;
```
gwitwer: Same as the latch above: mutation-tested, and **deleting this line entirely leaves all 617 tests green.**

The retry threshold is a judgement call with a real number attached (`MAX_RUN_POLL_FAILURES = 5`, ~10s at the 2s interval), and the comment explains exactly why it exists. Nothing checks that a 4th consecutive failure stays silent, that the 5th emits, or that `consecutiveFailures = 0` on the success path actually resets — that last one is the easiest to break later and the hardest to notice, since it only shows up as runs being declared `unobservable` slightly too eagerly.

Grep confirms no test anywhere references `startRunPolling`, `run_started`, `announcedRuns`, or `MAX_RUN_POLL_FAILURES`, and `web/e2e` has no `agent.run_*` coverage either — consistent with the PR body's own "e2e coverage for the run funnel — follow-up" checkbox, which is still unticked.

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -100,23 +218,31 @@ function scrubElement(element: Record<string, unknown>, isSecret: boolean): void
...
     delete properties.$elements_chain;
     return;
   }
-  properties.$elements_chain = properties.$elements_chain.replace(CHAIN_HREF_PATTERN, (match, href: string) => {
-    const sanitized = sanitizeUrl(href);
-    return typeof sanitized === "string" ? `attr__href="${sanitized}"` : match;
-  });
+  // Same allowlist as the object carrier, applied to the serialized form. A rule
+  // that covers only one of the two is a rule a remote-config flip disables.
+  let chain = properties.$elements_chain.replace(CHAIN_ANY_ATTRIBUTE, (match, name: string) =>
```
gwitwer, line 236: Behavior change worth calling out explicitly in the PR body: **`attr__href` is now dropped entirely, not sanitized.**

The previous code ran `sanitizeUrl` over `element.attr__href` and rewrote `attr__href="…"` in the chain. Both paths are gone — `href` simply isn't in `KEPT_ELEMENT_ATTRIBUTES`, so the allowlist removes it in every mode, and the corresponding test was removed too.

That's strictly safer and I'm not arguing against it. But it means autocapture no longer records *where* any link click went, in any mode — any existing PostHog Action or insight matching on `href` silently stops matching. `$el_text` still carries the visible label on `truncate` surfaces so most link analysis survives, but this is a deliberate capability loss that the commit message doesn't mention alongside the other decisions it does document.

Either note it in the "Breaking-ish" section next to the `$current_url`/`$host` shape change, or add `attr__href` back to the allowlist (it already flows through `sanitizeUrl`-equivalent stripping at the URL level via `redactUrls`).

packages/harness/web/src/lib/analytics/before-send.ts
```diff
@@ -128,37 +254,118 @@ function scrubElementCarriers(properties: Record<string, unknown>, isSecret: boo
...
     for (const key of Object.keys(personProperties)) {
       if (URL_VALUED_PERSON_PROPERTIES.test(key)) personProperties[key] = sanitizeUrl(personProperties[key]);
+      else if (HOST_VALUED_PERSON_PROPERTIES.test(key)) personProperties[key] = sanitizeHost(personProperties[key]);
     }
   }
 }
 
+/** An `attr__aria-label="…"` in the serialized chain. Anchored on the full
+ *  attribute name so it cannot match `aria-labelledby` or `aria-hidden`. */
+const CHAIN_ARIA_LABEL_PATTERN = /attr__aria-label="([^"]*)"/i;
```
gwitwer: Leftover from the escape-awareness fix: this one is still `[^"]*`.

`CHAIN_ANY_ATTRIBUTE` and `CHAIN_TEXT_PATTERN` both became `(?:\\.|[^"\\])*`, but the promotion path's pattern didn't. Consequence is much milder — promotion only runs in `truncate` mode, so an `aria-label` containing an escaped quote yields a truncated label rather than a leak — but it's the same latent bug in the same file, and someone reading the two escape-aware regexes above will reasonably assume this one matches.

Cosmetic; fold it in whenever this file is next touched.

packages/harness/web/src/lib/analytics/redaction-gate.test.ts
```diff
@@ -0,0 +1,285 @@
...
+  "SessionStepsBar.tsx": "`Preview :${preview.port}` — a port number the server chose.",
+  "TemplateUseDialog.tsx": "`Use ${template.name}` — a gallery template name from OUR registry.",
+  "TemplateCard.tsx":
+    "`What ${template.name} runs and costs` — GalleryTemplate, served by GET /api/templates. " +
+    "Ours, low-cardinality, and the label that makes the on-ramp funnel readable.",
+  "CanvasOverviewPanel.tsx": "Static labels only; interpolation is over our own counts.",
+};
+
+describe("components that build an aria-label by interpolation must tag their surface", () => {
+  const componentsDir = fileURLToPath(new URL("../../components", import.meta.url));
```
gwitwer, line 294: `readdirSync` here is non-recursive, so the tripwire only scans `components/*.tsx`.

Today that's harmless — the only subdirectory is `components/analytics/`, holding `TrackScope.tsx`, which has no `aria-label`. But the whole point of this check is to catch the component nobody thought about, and "someone adds `components/rail/WorkspaceRow.tsx`" is exactly that scenario. `readdirSync(dir, { recursive: true })` (Node 20+, which is the CI floor per the test matrix) closes it in one word.

Separately, and worth a line in the doc comment above alongside the two-render-branch caveat you already documented: the detector regex only matches template-literal labels (``aria-label={`…${x}`}``). `aria-label={"Focus " + label}` or `aria-label={computeLabel(x)}` pass silently. Neither form exists in the tree today — I checked — so this is about what the comment promises, not a live gap.

gwitwer (commented):
Re-reviewed at 91e9c90e. **All seven findings are genuinely fixed**, and the new gate is real rather than decorative — I mutation-tested it in a scratch worktree rather than trusting that it passes:

| Mutation | Result |
|---|---|
| Re-admit `attr__data-testid` to the allowlist (the original leak) | **5 gate tests fail** |
| Strip every `object` tag from `WorkflowsRail` | **tripwire fails**, with an error message that names the fix |
| Revert both chain regexes to `[^"]*` | **quote test fails** |

Full analytics suite at this commit: **85 passed / 5 files**, including the 17 new gate tests. (Note that `Test` and `Harness` haven't actually run on 91e9c90e — only CodeQL/Analyze/Labeler did, so the green checks on this commit don't cover the suite. I ran it locally against the PR head; worth a re-push or a manual dispatch before merging so CI records it.)

Specific credit where the response went past what I flagged:

- The **allowlist** is the right call over patching the denylist, and dropping `data-testid` rather than special-casing it is the harder, more correct half — a name-bearing testid is a bad Actions selector anyway.
- **`DirectoryPicker`** moving `object` down to the listing rows fixes a leak I flagged only as a blanking bug; the root-level tag was also mis-scoped.
- The **heatmap claim was corrected rather than defended**. That's the response I'd least expected and the one that matters most for whether the next person trusts these comments.
- `SAFE_INTERPOLATED_LABELS` requiring a written reason, and the gate doc naming its own two-render-branch blind spot (`WorkflowActionsHeader`) and where the DOM-level version belongs, are both better than what I asked for.

**One thing to resolve before merge**, inline on both sites: the two run-funnel fixes are correct and completely untested. Restoring either bug — the collapsed latch, or give-up-on-first-failure — leaves all 617 tests green. That's the same failure mode this PR just spent a round fixing on the redaction side: a real invariant, a careful comment explaining it, and nothing holding it in place. The redaction half now has a gate that bites; the half with two logic bugs found in one review round has none.

Extracting the latch/failure-count decision into a pure helper alongside `runErrorKind` would make it unit-testable without a React harness. If you'd rather ship now, that's defensible — the funnel is new, so a regression costs a metric rather than user data — but it should be a tracked follow-up with the e2e checkbox, not left implicit.

Remaining inline notes are minor: `attr__href` is now dropped entirely rather than sanitized (a real capability change that isn't in "Breaking-ish"), one escape-awareness fix didn't reach `CHAIN_ARIA_LABEL_PATTERN`, and the tripwire's `readdirSync` is non-recursive.

Redaction-wise I'd merge this as it stands.

gwitwer:
Took the blocking note. You were right that it was the same shape as round 1's leak — a real invariant, a careful comment, and nothing holding it in place — and on the half that had already produced two bugs in one review.

**Extracted `runFunnelStep` into `lifecycle.ts`**, beside `runErrorKind`, roughly the shape you sketched. The rules weren't untestable because they were complex; they were untestable because they lived in a closure over React refs, timers and `api.getRunState`. `startRunPolling` is now a thin driver that sends whatever the reducer returns.

I mutation-tested it the way you did rather than trusting that it passes:

| Mutation | Result |
|---|---|
| Collapse the two latches (the round-1 bug) | **1 test fails** |
| Give up on the first failed poll (the round-1 bug) | **2 tests fail** |
| Drop the failure-count reset on success | **1 test fails** |
| Allow a second terminal event | **1 test fails** |

That third one is the case you singled out, and you were right that it's the sneaky one — it only surfaces as runs called `unobservable` slightly too eagerly, so it gets a test of its own driving blips-then-success-then-blips and asserting silence.

**On the three minor notes, all taken:**

- `CHAIN_ARIA_LABEL_PATTERN` — correct, that one regex was missed. Milder as you say (promotion only runs in `truncate`, so the cost is a truncated label, not a leak), but fixed and now covered by a test with an escaped quote in the label.
- `readdirSync` — now `{ recursive: true }`. Also documented the *other* blind spot you named: the detector only recognises template-literal labels, so `"Focus " + label` or `computeLabel(x)` would pass silently. Neither exists today, but the doc now says so rather than leaving it for the next reader.
- **`attr__href`** — kept dropped rather than re-added, and documented in Breaking-ish. I checked the project before deciding: the only saved insight mentioning `href` is a web-app one on `/settings/plans` whose own description already reads *"Always 0 today: clicks on /settings are privacy-redacted"*. So nothing Studio-side regresses, and re-admitting an attribute to buy back an analysis nobody is running seemed like the wrong trade against the allowlist's whole point. Now pinned by a test asserting a signed URL doesn't survive.

Also worth reporting: **your prediction about the eighth component came true within hours.** #632 (artifact-first run workspace) landed on main mid-review and the gate went red on three of its components — `ArtifactRenderer` (artifact titles are agent-produced), `RunWorkspace` (`${step.name} attempt N`), and `SchemaInputFields` (`Remove item ${index+1}` — a position, recorded as known-safe). Two real leaks in code neither of us wrote, caught automatically. That's the best argument I have that the allowlist-plus-gate shape beat hand-tagging.

Separately: #632 also left this PR `CONFLICTING`, which is why `Test`/`Harness` were missing on `91e9c90e` — GitHub can't build a merge commit for a conflicted PR, so `pull_request` workflows are skipped silently while `pull_request_target` and CodeQL still run. Merged main in (not rebased, so your inline anchors survive). CI is 9/9 on the merged head.

Now: **2067 passed / 140 files**, 100 of them analytics. Typecheck and lint clean.

PR #633, opened 2026-08-13
feat(content-generation): route video verbs through the capability router (SAP-2575)

FornesLeo:
## E1 — Video transport consolidation (SAP-2575) · SDK half

Repoint `contentGeneration.video.*` from **provider-direct** (`POST ${falGateway}/run/<model>` + self-poll) onto the **capability router** — the same `capabilityCall` seam `contentGeneration.images` already uses. A clean swap: the backend `content.generation.video` capability is already registered and built for it (SAP-1666).

### What changed — `packages/tools/src/content-generation/`
- `createVideo` / `launchVideo` submit via `capabilityCall("content.generation.video", …)` on the single Core base URL (`resolveCoreBaseUrl()`) instead of building a `/run/<model>` URL against the fal gateway.
- `model` is now a request-body field the router's adapter resolves → **semantic aliases (`veo3-fast`, …) resolve from the SDK, no raw provider id required** (the literal fix). `params` ride nested (the adapter re-spreads them), matching the image path.
- Submit handle is camelCase (`{ requestId, statusUrl, responseUrl? }`); the polled result stays fal's raw shape (`mapVideo`). Public handle `{ requestId, dispatch, wait }` is **byte-identical**.
- `VIDEO_MODELS` `@deprecated`; dead code removed (`DEFAULT_BASE_URL`, `modelToPath`, `DEFAULT_VIDEO_MODEL`); the stale "aliases 404" docs corrected.
- `launchVideo` no longer sends `dispatch: 'async'` — `VideoCreateRequest` has no such field (video is async-only, unlike images).

### Acceptance
- [x] `video.launch({ model: 'veo3-fast' })` succeeds with no provider id in caller code — **verified live** against a local backend (handle returned, alias resolved, no 404).
- [x] Public handle shape byte-identical; `servedBy` stripped at the backend boundary (unchanged).
- [x] 522 unit tests green (content-generation 79/79), typecheck + lint clean.

### Follow-ups (not here)
- Backend: workflow pause/resume regression test through the routed path (the riskiest surface).
- SDK release + migration note; submit→asset-ready p50/p95 per model on the shared media-generation board.

### DIFF
--- a/.changeset/route-video-through-capability.md
+++ b/.changeset/route-video-through-capability.md
@@ -5,3 +5,8 @@
 content-generation: route `video.create` / `video.launch` through the `/v1/capabilities/content.generation.video` router (SAP-2575)
 
 Video verbs now submit through the shared capability router (like images), so the SDK no longer builds the gateway-direct `/run/<model>` URL: `model` is a request-body field the router's adapter resolves server-side (a semantic alias like `"veo3-fast"`, or a raw provider id, defaulted when omitted), and authentication rides the `/v1` guard's `x-api-key` header. The public `video.create` / `video.launch` surface and the submit-then-poll-to-completion behavior are unchanged.
+
+**Breaking — video's base URL moves to Core.** Video now resolves its base URL from Core like every other routed capability (`resolveCoreBaseUrl()` → `SAPIOM_BASE_URL` ?? `SAPIOM_API_URL` ?? `https://api.sapiom.ai`):
+
+- The content-generation-specific `SAPIOM_CONTENT_GENERATION_URL` override is **no longer read** for video — set **`SAPIOM_BASE_URL` / `SAPIOM_API_URL`** instead.
+- Any explicit `baseUrl` passed to `video.create` / `video.launch` must now point to **Core** (e.g. `https://api.sapiom.ai`), **not the Fal gateway**. A Fal-gateway URL here will now hit the wrong host.
--- a/.changeset/route-video-through-capability.md
+++ b/.changeset/route-video-through-capability.md
@@ -1,0 +1,7 @@
+---
+"@sapiom/tools": minor
+---
+
+content-generation: route `video.create` / `video.launch` through the `/v1/capabilities/content.generation.video` router (SAP-2575)
+
+Video verbs now submit through the shared capability router (like images), so the SDK no longer builds the gateway-direct `/run/<model>` URL: `model` is a request-body field the router's adapter resolves server-side (a semantic alias like `"veo3-fast"`, or a raw provider id, defaulted when omitted), and authentication rides the `/v1` guard's `x-api-key` header. The public `video.create` / `video.launch` surface and the submit-then-poll-to-completion behavior are unchanged.
--- a/packages/tools/src/content-generation/index.spec.ts
+++ b/packages/tools/src/content-generation/index.spec.ts
@@ -282,6 +282,14 @@ describe("contentGeneration.video.create()", () => {
       (c) =>
         c.init.method === "POST"
           ? jsonResponse({
+              requestId: "req-merge",
+              statusUrl: `${BASE}/queue/req-merge/status`,
+              responseUrl: `${BASE}/queue/req-merge`,
+            })
+          : null,
+      (c) =>
+        c.init.method === "GET"
+          ? jsonResponse({
               video: {
                 url: "https://media/merged.mp4",
                 content_type: "video/mp4",
@@ -290,17 +298,22 @@ describe("contentGeneration.video.create()", () => {
           : null,
     ]);
 
-    const out = await createVideo({ prompt: "merge" }, transport, BASE);
+    const out = await createVideo(
+      { prompt: "merge", pollIntervalMs: 1 },
+      transport,
+      BASE,
+    );
 
-    // The router returns the sync-completion `video` object (some fal ops, notably
-    // ffmpeg merge, complete inline) — mapped snake → camelCase.
+    // The submit always returns a queue handle (the video adapter's
+    // `fromGatewayResponse` throws without one) — the poll target still returns
+    // fal's raw snake_case result, mapped to camelCase here.
     expect(out).toEqual({
       video: {
         url: "https://media/merged.mp4",
         contentType: "video/mp4",
       },
     });
-    expect(calls).toHaveLength(1);
+    expect(calls).toHaveLength(2);
     expect(calls[0]!.url).toBe(
       `${BASE}/v1/capabilities/content.generation.video`,
     );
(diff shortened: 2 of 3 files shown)


ratataque:
The transport change itself looks sound: the request now goes through the shared Core capability seam, aliases remain server-resolved, the queue poll contract is preserved, and both `Test` and `Harness` CI runs are green. The known inline `ffmpeg-merge` incompatibility is also addressed by merged backend PR sapiom/Sapiom#4331.

Before approving this as the SAP-2575 SDK deliverable, I think two release/acceptance gaps still need to be closed.

## 1. Add a changeset and an explicit migration note

### Why this is required

This PR changes the behavior of the published `@sapiom/tools` package:

- video submit moves from the Fal gateway to `/v1/capabilities/content.generation.video`;
- the authentication header changes internally from `x-sapiom-api-key` to `x-api-key`;
- `SAPIOM_CONTENT_GENERATION_URL` is no longer read;
- the third `baseUrl` argument, when used directly, now means the Core base URL rather than the Fal gateway;
- semantic aliases become the preferred model input and `VIDEO_MODELS` is deprecated.

The repository's contribution policy requires a changeset whenever a published package's behavior or API changes. Without one, this PR is not independently releasable and none of this appears in the package release notes.

The environment-variable change deserves particular care. A deployment that currently sets only:

```sh
SAPIOM_CONTENT_GENERATION_URL=http://local-fal-gateway:3000
```

will stop using that value. `resolveCoreBaseUrl()` falls back through `SAPIOM_BASE_URL`, `SAPIOM_API_URL`, then `https://api.sapiom.ai`. Without a migration note, a local or staging process can unexpectedly call the production Core endpoint.

### How to address it

Add a changeset along these lines:

```md
---
"@sapiom/tools": minor
---

content-generation: route video create/launch through the Core capability router.

Video generation now accepts semantic aliases such as `veo3-fast`; raw provider IDs
remain supported for compatibility and `VIDEO_MODELS` is deprecated.

Migration: video generation now uses `SAPIOM_BASE_URL` / `SAPIOM_API_URL`.
Replace any `SAPIOM_CONTENT_GENERATION_URL` configuration, and ensure any explicit
third `baseUrl` argument points to Core rather than the Fal gateway.
```

Also update the content-generation README with an alias-first example:

```ts
const result = await sapiom.contentGeneration.video.create({
  prompt: "a calm ocean wave at sunset",
  model: "veo3-fast",
  params: { duration: 8 },
});
```

A concise migration table would make the host change unambiguous:

```md
| Before | After |
| --- | --- |
| `SAPIOM_CONTENT_GENERATION_URL` | `SAPIOM_BASE_URL` or `SAPIOM_API_URL` |
| Fal gateway URL passed as `baseUrl` | Core API URL passed as `baseUrl` |
| `VIDEO_MODELS.veo3Fast` | `"veo3-fast"` |
```

## 2. Verify the full workflow pause/resume lifecycle

### Why the current tests are not enough

SAP-2575 explicitly calls the completion path the riskiest surface and requires:

> A workflow that pauses on a video handle and resumes on webhook completion passes end to end through the new path.

This PR proves the SDK sends `x-sapiom-workflow-token`. Backend PR sapiom/Sapiom#4324 is useful because it proves the routed Fal call preserves that token through collapsed routing. However, it is still open and deliberately tests only that routing seam.

Neither test proves the complete lifecycle:

```text
SDK launch
  → Core capability router
  → Fal async submit
  → persisted workflow token
  → terminal Fal webhook
  → FalWorkflowResumeService
  → workflows-engine capability-resume
  → paused execution resumes with VideoResultPayload
```

A submit can succeed and return a perfectly valid handle while a defect later in this chain leaves the workflow paused forever. That is why the acceptance criterion asks for end-to-end verification rather than another submit unit test.

### How to test it

The integration test should mock only the external Fal boundary and exercise the real internal resume path. Conceptually:

```ts
it("resumes a workflow paused on a routed video launch", async () => {
  const execution = await startWorkflow({
    run: async (ctx) => {
      const handle = await ctx.sapiom.contentGeneration.video.launch({
        prompt: "a wave",
        model: "veo3-fast",
        storage: { visibility: "private" },
      });

      return pauseUntilSignal(handle, { resumeStep: "finalize" });
    },
  });

  expect(execution.status).toBe("paused");
  expect(execution.signal).toEqual({
    name: VIDEO_RESULT_SIGNAL,
    correlationId: "fal-request-123",
  });

  // Deliver the same terminal webhook the Fal gateway receives in production.
  await deliverFalTerminalWebhook({
    requestId: "fal-request-123",
    video: { url: "https://media/video.mp4" },
  });

  await processFalStorageAndResumeQueues();

  const resumed = await waitForExecution(execution.id, "finalize");
  expect(resumed.input).toEqual({
    outputs: [
      {
        fileId: expect.any(String),
        downloadUrl: expect.any(String),
        downloadUrlExpiresAt: expect.any(String),
      },
    ],
  });
});
```

The exact harness can differ, but the test should not shortcut the behavior by directly calling the workflow signal endpoint; it must exercise the webhook/resume producer that this transport change relies on.

If the end-to-end test belongs in the backend repository, that is fine, but it should be linked here and merged—or the dependency should be explicitly recorded—before SAP-2575 is considered complete.

With the changeset/migration documentation and the full resume regression in place, I would be comfortable approving this PR.

ratataque (changes requested):
Requesting changes for the two release/acceptance blockers detailed in my review comment: (1) add the required @sapiom/tools changeset and migration guidance for the Core URL/environment-variable change, and (2) complete or explicitly gate on the end-to-end pause → webhook → resume regression required by SAP-2575. Detailed rationale and implementation examples: https://github.com/sapiom/sapiom-js/pull/633#issuecomment-5298833764

FornesLeo:
Added the missing changeset in 0303c59 (`@sapiom/tools` minor — SAP-2575 video capability routing), so the version-packages bot will bump/publish. Re-requesting review 🙏

ratataque:
I only need one remaining change before approval: update the changeset or migration documentation to state that:

- SAPIOM_CONTENT_GENERATION_URL is replaced by SAPIOM_BASE_URL / SAPIOM_API_URL.
- Any explicit baseUrl passed to the video functions must now point to Core, not the Fal gateway.

FornesLeo:
Documented in 0303c59's changeset (updated): the base-URL move is now called out as breaking — `SAPIOM_CONTENT_GENERATION_URL` is no longer read for video (use `SAPIOM_BASE_URL` / `SAPIOM_API_URL`), and an explicit `baseUrl` on `video.create` / `video.launch` must now point to Core, not the Fal gateway. Thanks for catching it.

ratataque (approved):
LGTM

PR #634, opened 2026-08-14
fix(agent): skeletonFromJsonSchema handles nullable type-union shape

### LINKED ISSUE #614: [Bug]: agent: skeletonFromJsonSchema ignores array form type (nullable types always return null)
### Affected package or area

@sapiom/agent packages/agent/src/introspection.ts, skeletonFromJsonSchema

latest main (3cbe957)

### Environment

Any pure TypeScript logic, no platform dependency

1. Define an agent with a nullable input field, e.g. z.string().nullable()
2. Zod generates JSON Schema: { type: ["string", "null"] }
3. skeletonFromJsonSchema receives { type: ["string", "null"] }
4. switch(schema.type) receives an array no case matches
5. Falls through to default: return null

### Expected behavior

skeletonFromJsonSchema({ type: ["string", "null"] }) returns "" (the non-null type's skeleton)
skeletonFromJsonSchema({ type: ["number", "null"] }) returns 0

### Actual behavior

Returns null for any nullable type the example seed value is always null instead of a type appropriate placeholder.

Fix: normalize schema.type before the switch 
const type = Array.isArray(schema.type)
  ? schema.type.find(t => t !== 'null') ?? 'null'
  : schema.type;
Affected: introspection.ts line 108

### Acknowledgements

- [x] This report does not describe a suspected security vulnerability; I will use the [Security Policy](https://github.com/sapiom/sapiom-js/security/policy) for private reporting.

ygd58:
Fixes #614

`exampleFromJsonSchema` (via its internal `skeletonFromJsonSchema` helper) returned `null` for any field whose JSON Schema `type` is an array containing `"null"` (e.g. `{ type: ["string", "null"] }`), instead of a type-appropriate placeholder. The `switch(schema.type)` statement never matched an array, so it always fell through to the `null` default.

The `type` field can validly be an array of alternatives per the JSON Schema spec, and some Zod versions (and non-Zod JSON Schema producers) represent a nullable field this way rather than as `anyOf`. `exampleFromJsonSchema` is a public entry point for schemas generated by whatever tooling the caller uses (per its own doc comment), so this shape must be handled regardless of which zod version/tool produced it - not just the `anyOf` branch already handled above.

**Fix:** normalize an array `type` to its non-null member before the switch.

**Tests:** this repo's bundled zod version actually produces `anyOf` for `.nullable()` (not a type array), so the zod-based tests pass via the pre-existing `anyOf` handling and mainly guard against regressions there. The raw type-array tests (constructing `{ type: [...] }` directly) exercise the actual fix and fail without it - verified by temporarily reverting the fix and confirming that test failed, then restoring it.

**Changeset:** added (`@sapiom/agent`, patch).

Verified locally: `vitest run src/introspection.test.ts` (5/5 passing), `eslint` clean.


gwitwer:
Closing in favour of #615, the same one-line normalization with green CI. This branch's test imports vitest but packages/agent runs jest, so typecheck fails. Thanks.

PR #635, opened 2026-08-14
feat(content-generation): surface SAP-2576 cost envelope on SDK media results

FornesLeo:
## Summary

The E2 backend (SAP-2576) already puts a per-generation cost envelope on every media capability response, but its **SDK half was deferred** (#4289) — so a typed `@sapiom/tools` consumer (e.g. Polsia ads re-billing its customers' wallets) could not read `cost.estimateUsd` / `cost.reference`. The settled-cost track was blocked at the SDK, not at the caller. This surfaces that envelope on the SDK results.

> **Stacked on #633** (SAP-2575 / E1 — video verbs → capability router). Base is that branch, so this diff is E2-SDK only. Rebase onto `main` once #633 merges.

## Changes

- Add `MediaCostEnvelope` — the SDK mirror of the backend envelope (`estimateUsd`, `currency`, `isEstimate`, `source`, `reference`).
- `cost?` + `resolvedModel?` on `ImageGenerationResult` / `VideoGenerationResult` and the `ImageDispatchResponse` / `VideoDispatchResponse` submit handles.
- Expose `cost` / `resolvedModel` on the `images.launch` / `video.launch` handles (readable without awaiting `wait()`).
- Thread the submit handle's cost onto the **polled** result for video + async image — the gateway's queue passthrough carries neither cost nor `resolvedModel`. Sync image passes it through the existing response-body spread.
- 5 new tests; `.changeset` (minor → `0.27.0`).

## Motivation

Polsia's ads video re-bills its own customers' ad wallets by the per-generation cost. It calls the media capability but can't read the cost back without this. Design **D1 option C**: `estimateUsd` inline, settled amount out-of-band via `cost.reference` → `GET /v1/transactions/:id/costs`. Envelope shape is stable from E2 onward (E6 only flips `source` → `'metered'`).

## Testing

- [x] `pnpm --filter @sapiom/tools typecheck` — clean
- [x] `pnpm --filter @sapiom/tools test` (content-generation) — 85/85 (5 new)
- [x] prettier + eslint clean

## Related

- Refs: SAP-2576 — fulfills the epic's "SDK types updated and released" deliverable.
- Stacked on #633 (SAP-2575).

## Checklist

- [x] All tests passing
- [x] No hardcoded secrets
- [x] Self-reviewed

### DIFF
--- a/.changeset/content-generation-cost-envelope.md
+++ b/.changeset/content-generation-cost-envelope.md
@@ -2,11 +2,13 @@
 "@sapiom/tools": minor
 ---
 
-content-generation: surface the SAP-2576 per-generation cost envelope on image & video results
+content-generation: surface the SAP-2576 per-generation cost envelope + `resolvedModel` across every media-result path
 
-`ImageGenerationResult` / `VideoGenerationResult` — and the `images.launch` / `video.launch`
-handles — now carry `cost?: MediaCostEnvelope` (`estimateUsd` inline plus the settled charge
-out-of-band via `cost.reference`) and `resolvedModel?`, mirroring the backend capability
-contract (SAP-2576). For video the envelope resolves at submit, so it is threaded from the
-dispatch handle onto the polled result (the gateway's queue passthrough carries neither). A
-reseller can now price a generation without a second API call.
+Consumers (e.g. Polsia ads re-billing its customers' wallets) can now price a generation without a second API call, consistently across synchronous image calls, polled results, launch handles, AND durable workflow resumes:
+
+- New `MediaCostEnvelope` (`estimateUsd` inline + the settled charge out-of-band via `cost.reference`) and `resolvedModel` on `ImageGenerationResult` / `VideoGenerationResult` and the `images.launch` / `video.launch` handles.
+- `resolvedModel` is **required** (the routed backend always echoes it — a cataloged raw id reverse-maps to its alias, an uncataloged one is echoed verbatim), matching the backend SAP-2576 contract. `cost` stays optional (quote/reference are best-effort).
+- The durable resume payload (`VideoResultPayload` / `ImageResultPayload`, via `toVideoResumePayload` / `toImageResumePayload`) now carries `resolvedModel` + `cost` (new shared `MediaResumeFields`), so a workflow step that bills in the **resumed** step — after generation — can still read `cost.reference`.
+- For video the envelope resolves at submit and is threaded from the dispatch handle onto the polled result (the gateway's queue passthrough carries neither).
+
+> Companion (backend, chengdu): the Fal workflow-resume producer must emit `resolvedModel` + `cost` in the resume payload JSON for real webhook-driven resumes to carry them; the SDK mapper covers local stubs/tests only.
--- a/.changeset/content-generation-cost-envelope.md
+++ b/.changeset/content-generation-cost-envelope.md
@@ -1,0 +1,12 @@
+---
+"@sapiom/tools": minor
+---
+
+content-generation: surface the SAP-2576 per-generation cost envelope on image & video results
+
+`ImageGenerationResult` / `VideoGenerationResult` — and the `images.launch` / `video.launch`
+handles — now carry `cost?: MediaCostEnvelope` (`estimateUsd` inline plus the settled charge
+out-of-band via `cost.reference`) and `resolvedModel?`, mirroring the backend capability
+contract (SAP-2576). For video the envelope resolves at submit, so it is threaded from the
+dispatch handle onto the polled result (the gateway's queue passthrough carries neither). A
+reseller can now price a generation without a second API call.
--- a/packages/tools/src/content-generation/index.spec.ts
+++ b/packages/tools/src/content-generation/index.spec.ts
@@ -1217,11 +1217,16 @@ describe("cost envelope (SAP-2576)", () => {
     expect(out.video?.url).toBe("https://media/v.mp4");
   });
 
-  it("createVideo leaves the result untouched when the submit handle carries no cost", async () => {
+  it("createVideo surfaces resolvedModel with no cost when the quote was unavailable", async () => {
     const { transport } = makeTransport([
       (c) =>
         c.init.method === "POST"
-          ? jsonResponse({ requestId: "r", responseUrl: `${BASE}/queue/r` })
+          ? jsonResponse({
+              requestId: "r",
+              responseUrl: `${BASE}/queue/r`,
+              // Contract: the router always echoes resolvedModel, even when the price join is down.
+              resolvedModel: "veo3-fast",
+            })
           : null,
       (c) =>
         c.init.method === "GET"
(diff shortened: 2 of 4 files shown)


ratataque:
Thanks for wiring the SAP-2576 envelope through the normal `create()`, `launch()`, and `wait()` paths. The submit-handle → polled-result merge is the right approach for the current queue passthrough.

I think there are two contract gaps to close before this can fully satisfy the SAP-2576 SDK deliverable.

## 1. The workflow pause/resume path drops `cost` and `resolvedModel`

### Why this matters

The new fields are available on the launch handle, but a durable workflow normally does this:

```ts
const handle = await ctx.sapiom.contentGeneration.video.launch(input);
return pauseUntilSignal(handle, { resumeStep: "bill-generation" });
```

`pauseUntilSignal(handle, ...)` reduces the handle to its `resultSignal` and `correlationId`. The resumed step does not receive the original handle. Today, `ImageResultPayload` and `VideoResultPayload` contain only `outputs`, and both `toImageResumePayload()` and `toVideoResumePayload()` omit the new metadata.

That means the most important durable-workflow consumer cannot read:

```ts
result.cost?.estimateUsd
result.cost?.reference
result.resolvedModel
```

This is especially problematic for the stated Polsia rebilling use case: billing will commonly happen after the generation finishes, in the resumed step, but `cost.reference` has been lost by that point.

### How to fix it

The SDK resume contract should include the same generation metadata:

```ts
interface MediaResumeFields {
  resolvedModel: string;
  cost?: MediaCostEnvelope;
}

export interface VideoResultPayload extends MediaResumeFields {
  outputs: Array<{
    fileId?: string;
    downloadUrl?: string;
    downloadUrlExpiresAt?: string;
    downloadUrlUnavailable?: boolean;
    storageError?: string;
  }>;
}

export interface ImageResultPayload extends MediaResumeFields {
  outputs: Array<{
    fileId?: string;
    downloadUrl?: string;
    downloadUrlExpiresAt?: string;
    downloadUrlUnavailable?: boolean;
    storageError?: string;
  }>;
}
```

The local/stub mappers should retain those fields:

```ts
export function toVideoResumePayload(
  result: VideoGenerationResult,
): VideoResultPayload {
  const metadata = {
    resolvedModel: result.resolvedModel,
    ...(result.cost !== undefined && { cost: result.cost }),
  };

  if (!result.video) return { outputs: [], ...metadata };

  return {
    outputs: [
      {
        ...(result.video.fileId !== undefined && {
          fileId: result.video.fileId,
        }),
        ...(result.video.downloadUrl !== undefined && {
          downloadUrl: result.video.downloadUrl,
        }),
        ...(result.video.downloadUrlExpiresAt !== undefined && {
          downloadUrlExpiresAt: result.video.downloadUrlExpiresAt,
        }),
        ...(result.video.storageError !== undefined && {
          storageError: result.video.storageError,
        }),
      },
    ],
    ...metadata,
  };
}
```

The corresponding backend Fal workflow-resume producer must emit the same JSON shape:

```ts
{
  outputs,
  resolvedModel,
  ...(cost !== undefined && { cost }),
}
```

Updating only the SDK mapper would fix local stubs/tests but not real webhook-driven resumes, so this needs a cross-boundary regression test covering launch → pause → terminal webhook → resumed-step payload. Please include the reference-only case too:

```ts
cost: { reference: "txn_abc123" }
```

That is a valid and important envelope when the price probe is unavailable but the transaction was still created.

## 2. `resolvedModel` should not be optional

The backend SAP-2576 contract makes `resolvedModel` required on both finished image responses and image/video dispatch handles. It also guarantees a value for legacy raw IDs: cataloged raw IDs are reverse-mapped to their public alias, while an uncataloged raw ID is echoed verbatim.

The SDK currently declares:

```ts
resolvedModel?: string;
```

and documents it as absent for a legacy uncataloged ID. That contradicts the backend behavior and the acceptance criterion that `resolvedModel` is present on every response. It also prevents a typed consumer from safely using the field as the model/cost grouping key without unnecessary fallback logic.

Please keep `cost` optional—quote/reference availability is intentionally best-effort—but make `resolvedModel` required across the routed public contract:

```ts
export interface ImageGenerationResult {
  images?: GeneratedImage[];
  resolvedModel: string;
  cost?: MediaCostEnvelope;
  [key: string]: unknown;
}

export interface VideoGenerationResult {
  video?: GeneratedVideo;
  resolvedModel: string;
  cost?: MediaCostEnvelope;
  [key: string]: unknown;
}

interface ImageDispatchResponse {
  requestId: string;
  statusUrl: string;
  responseUrl?: string;
  resolvedModel: string;
  cost?: MediaCostEnvelope;
}

interface VideoDispatchResponse {
  requestId: string;
  statusUrl: string;
  responseUrl?: string;
  resolvedModel: string;
  cost?: MediaCostEnvelope;
}
```

The public `ImageLaunchHandle` / `VideoLaunchHandle` and the `withDispatchCost` handle parameter should follow the same requiredness. The current test that expects `resolvedModel` to be undefined when the mocked submit handle omits it is testing a response the routed backend no longer produces; replace it with coverage for defaulted, pinned-alias, and legacy-raw-ID values.

If compatibility with an older Core deployment is the reason for keeping it optional, that should be an explicit compatibility decision in the PR/changeset rather than described as current SAP-2576 behavior.

Once these two paths are covered, the SDK contract will be consistent across synchronous image calls, polled results, launch handles, and durable workflow resumes—the full set of ways callers consume media-generation results.

ratataque (changes requested):
Requesting changes for the two SAP-2576 contract gaps detailed in my review comment: (1) preserve cost and resolvedModel across the durable workflow pause/resume payload, including the real backend webhook path, and (2) make resolvedModel required consistently with the backend contract, or explicitly document a compatibility exception. Detailed rationale, code examples, and test recommendations: https://github.com/sapiom/sapiom-js/pull/635#issuecomment-5298827362

FornesLeo:
Thanks @ratataque — both gaps addressed in b8e7c29.

**1. Resume payload now carries the metadata.** New shared `MediaResumeFields`; `VideoResultPayload` / `ImageResultPayload` extend it, and `toVideoResumePayload` / `toImageResumePayload` propagate `resolvedModel` + `cost` (including the reference-only envelope). A step that bills in the *resumed* step can now read `cost.reference` / `resolvedModel`.

**2. `resolvedModel` is now required** on `ImageGenerationResult` / `VideoGenerationResult`, `Image`/`VideoDispatchResponse`, `Image`/`VideoLaunchHandle`, and the `withDispatchCost` handle param. `cost` stays optional (quote/reference best-effort). Replaced the "resolvedModel undefined" test with pinned-alias / legacy-raw-id / reference-only / cost-absent coverage; the stub echoes a `resolvedModel` too.

**Companion backend change (follow-up in chengdu):** you're right that updating only the SDK mapper covers local stubs/tests, not real webhook-driven resumes — the chengdu Fal workflow-resume producer must emit `resolvedModel` + `cost` in the resume payload JSON. Flagged in the changeset; I'll prep the backend PR + the cross-boundary regression test as the server half. This PR is the client half.

Green locally: `typecheck` clean, **533/533** tests, `eslint` + `prettier` clean. Re-requesting review 🙏

ratataque:
Before approving, require:

- Merge #633, then rebase #635 onto main.
- Run CI on the rebased commit; Test and Harness must pass. No checks are currently visible for b8e7c29.
- Update the stale PR description: resolvedModel is now required, not optional, and the test summary has changed.
- Link the concrete backend follow-up that emits resolvedModel and cost during real workflow resume.
- Explicitly make that backend PR plus its cross-boundary E2E test a production-rollout/SAP-2576 completion gate.

PR #638, opened 2026-08-14
feat(content-generation): surface SAP-2576 cost envelope on SDK media results

FornesLeo:
> **Re-opens #635**, which GitHub auto-closed when its stacked base branch (`leofornes/sap-2575-…`) was deleted on the merge of #633. Same branch + same commits, now **rebased onto `main`** (E1 / #633 has landed). Full prior review thread: #635.

## Summary
Surface the SAP-2576 per-generation cost envelope + `resolvedModel` across every media-result path in `@sapiom/tools`, so a consumer can price a generation without a second API call — consistent across sync image calls, polled results, launch handles, and durable workflow resumes.

## Changes
- `MediaCostEnvelope` (`estimateUsd` inline + settled charge out-of-band via `cost.reference`) + `resolvedModel` on `ImageGenerationResult` / `VideoGenerationResult` and the `images.launch` / `video.launch` handles.
- For video, the envelope resolves at submit → threaded from the dispatch handle onto the polled result (the gateway's queue passthrough carries neither).
- `resolvedModel` is **required** (backend always echoes it: alias, or raw id verbatim); `cost` stays optional.
- Durable resume payload (`VideoResultPayload` / `ImageResultPayload`) carries `resolvedModel` + `cost` (`MediaResumeFields`) — so a step billing in the *resumed* step reads `cost.reference`.
- Changeset (minor → 0.27.0).

## Review already addressed (from #635, @ratataque)
Both CHANGES_REQUESTED gaps are fixed in this branch (commit `address SAP-2576 SDK review`):
1. Resume payload now carries `cost` + `resolvedModel` (incl. reference-only case).
2. `resolvedModel` made required across results / dispatch / launch handles.
Reply: https://github.com/sapiom/sapiom-js/pull/635#issuecomment-5298971515 · Companion backend (Fal workflow-resume producer emitting the fields) tracked separately.

## Testing
- [x] `typecheck` clean · **533/533** tests · `eslint` + `prettier` clean.

### DIFF
--- a/.changeset/content-generation-cost-envelope.md
+++ b/.changeset/content-generation-cost-envelope.md
@@ -2,11 +2,13 @@
 "@sapiom/tools": minor
 ---
 
-content-generation: surface the SAP-2576 per-generation cost envelope on image & video results
+content-generation: surface the SAP-2576 per-generation cost envelope + `resolvedModel` across every media-result path
 
-`ImageGenerationResult` / `VideoGenerationResult` — and the `images.launch` / `video.launch`
-handles — now carry `cost?: MediaCostEnvelope` (`estimateUsd` inline plus the settled charge
-out-of-band via `cost.reference`) and `resolvedModel?`, mirroring the backend capability
-contract (SAP-2576). For video the envelope resolves at submit, so it is threaded from the
-dispatch handle onto the polled result (the gateway's queue passthrough carries neither). A
-reseller can now price a generation without a second API call.
+Consumers (e.g. Polsia ads re-billing its customers' wallets) can now price a generation without a second API call, consistently across synchronous image calls, polled results, launch handles, AND durable workflow resumes:
+
+- New `MediaCostEnvelope` (`estimateUsd` inline + the settled charge out-of-band via `cost.reference`) and `resolvedModel` on `ImageGenerationResult` / `VideoGenerationResult` and the `images.launch` / `video.launch` handles.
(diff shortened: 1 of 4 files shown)


ratataque (approved):
please link the concrete backend PR/ticket covering:

- Emitting resolvedModel and cost from the real Fal workflow-resume producer.
- A cross-boundary E2E test covering webhook completion → workflow resume.

This does not need to block merging #638, but it must explicitly gate the SAP-2576 production rollout. Without it, the SDK paths are covered while real durable workflow resumes could still omit the new fields.

FornesLeo:
Thanks @ratataque. Backend companion tracked as **SAP-2650** (child of SAP-2576, flagged **High / gates the SAP-2576 E2 prod rollout**): https://linear.app/sapiom/issue/SAP-2650

It covers exactly your two points:
- `FalWorkflowResumeService.toResumeResult` emitting `resolvedModel` + `cost: { reference }` from the real Fal workflow-resume producer (persisted on the fal async intent at submit — `transaction_reference` from the gateway's own `request.sapiomTransactionId`, `resolved_model` threaded via `x-sapiom-resolved-model` like `x-sapiom-workflow-token`).
- A cross-boundary E2E: webhook completion → capability-resume bridge → resumed-step payload carries both (incl. reference-only).

Backend PR is in progress; it'll attach to SAP-2650 and won't flip E2 to prod until green. As you noted, this doesn't block the #638 merge (done).

PR #654, opened 2026-08-17
fix(mcp): capText's truncation marker could push output past budget

### LINKED ISSUE #653: capText's truncation marker can push output past the documented hard char budget
Summary
`execution-projection.ts` documents `capText` (in `packages/mcp/src/tools/shared.ts`) as a "hard char budget" primitive: "the result is bounded for ANY argument combination." The implementation didn't honor that. It sliced text to exactly `budget` characters, then appended a truncation marker (e.g. `…[truncated 48000 chars — open https://... for the full value]`) AFTER the slice — so the returned string was `budget + marker.length` characters, not `budget`.

Reproduction
```js
capText("x".repeat(50000), 2000, "https://app.sapiom.ai/runs/exec-123/steps/4").length
// -> 2094, not 2000
```
Verified with a 240-case brute-force sweep across budgets (including the small ones actually used in production: `PREVIEW_BUDGET = 2000`, `DEFAULT_FIELD_BUDGET = 32000`) and a range of text lengths — the old implementation violated the bound in every case where truncation kicked in.

Impact
Low in this specific call site today: `CEILING` in the test suite is 250,000 chars, and the marker overrun is on the order of tens to ~100 chars per field, so it's very unlikely to blow the overall response ceiling in practice. But it directly contradicts the module's own documented "hard budget... bounded for ANY argument combination" guarantee, and any future caller (or a lowered `CEILING`/per-field budget) relying on that documented guarantee would be silently wrong.

Suggested fix
Since the marker's own length depends on the dropped-char count, which depends on where the text is sliced, which depends on the marker's length, this needs a small converging loop (shrink the slice point until slice + marker fits within budget) rather than a single subtraction, plus a hard `.slice(0, budget)` backstop for degenerate tiny budgets. I have a fix ready with a regression test suite (`shared.test.ts`, which didn't exist for this module before) that fails 4/6 cases against the current implementation and passes 6/6 against the fix. Happy to open a PR.

teyrebaz33:
## Primary change type

- [x] Bug fix
- [ ] Documentation
- [ ] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

`execution-projection.ts` documents `capText` (`packages/mcp/src/tools/shared.ts`) as a "hard char budget" primitive: "the result is bounded for ANY argument combination." The implementation didn't honor that — it sliced text to exactly `budget` characters, then appended a truncation marker AFTER the slice, so the returned string was `budget + marker.length` characters, not `budget`. Concretely, `capText(text, 2000, someUrl)` could return a 2094-char string.

## Summary and scope

Since the marker's own length depends on the dropped-char count, which depends on where we slice, which depends on the marker's length, this resolves it with a small converging loop that shrinks the slice point until slice + marker fits within `budget`, with a final `.slice(0, budget)` backstop for degenerate tiny budgets. This also fixes the reported dropped-char count, which the old version could get slightly wrong for the same reason.

Out of scope: no changes to `execution-projection.ts` itself or to the field budgets it configures; no dependency changes.

## Related work

Related issue or discussion: #653

## Validation

```text
# packages/mcp: vitest run / tsc --noEmit / eslint src --ext .ts
15/15 test files, 107/107 tests passing (including the new suite)
tsc --noEmit: clean
eslint: 0 errors, 0 warnings
```

### Tests and documentation

Added `packages/mcp/src/tools/shared.test.ts` (no test file existed for this module before): a 240-case brute-force sweep across budgets (including 0/1/5/10, and the real `PREVIEW_BUDGET`/`DEFAULT_FIELD_BUDGET` values) confirming the fix never exceeds `budget`, plus a regression test verified via `git stash` to fail 4/6 cases against the pre-fix implementation and pass 6/6 against the fix. No user-facing documentation changes needed.

- Breaking or externally visible changes: None. `capText`'s signature and return type are unchanged; only truncated outputs near a field's budget boundary get slightly shorter (by the marker's length) to actually respect the documented bound.
- Changeset: Added (`.changeset/fix-captext-budget-overrun.md`), patch bump.

## Security

      will follow the
      private reporting.

## AI assistance

I used Claude (Anthropic) as a coding assistant throughout: it helped find the bug (including writing a small script that proved the overrun with real numbers), draft the fix and the brute-force test sweep, and run the verification commands quoted above. I reviewed and ran every command myself, read and understood the resulting diff line by line, and can explain and maintain every change in this PR.

## Checklist

      any N/A checks above.

### DIFF
--- a/.changeset/fix-captext-budget-overrun.md
+++ b/.changeset/fix-captext-budget-overrun.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/mcp": patch
+---
+
+Fix a truncation helper that could return output slightly longer than its configured character budget.
--- a/packages/mcp/src/tools/shared.test.ts
+++ b/packages/mcp/src/tools/shared.test.ts
@@ -1,0 +1,66 @@
+/**
+ * `capText` — the hard char-budget primitive execution-projection.ts relies on
+ * to keep every field bounded "for ANY argument combination" (see that
+ * module's doc comment). Regression coverage for a bug where the truncation
+ * marker appended after slicing pushed the result past `budget`, since the
+ * marker's own length was never subtracted from the content slice.
+ */
+import { describe, it, expect } from "vitest";
+import { capText } from "./shared.js";
+
+describe("capText", () => {
+  it("returns text unchanged when it already fits the budget", () => {
+    expect(capText("hello", 10)).toBe("hello");
+    expect(capText("hello", 5)).toBe("hello");
+  });
+
+  it("never returns a string longer than budget, with a webappUrl", () => {
+    const text = "x".repeat(50_000);
+    const budget = 2_000;
+    const result = capText(text, budget, "https://app.sapiom.ai/runs/exec-1");
+    expect(result.length).toBeLessThanOrEqual(budget);
+    expect(result).toContain("truncated");
+    expect(result).toContain("app.sapiom.ai");
+  });
+
+  it("never returns a string longer than budget, without a webappUrl", () => {
+    const text = "x".repeat(50_000);
+    const budget = 2_000;
+    const result = capText(text, budget);
+    expect(result.length).toBeLessThanOrEqual(budget);
+    expect(result).toContain("truncated");
+  });
+
+  it("reports an accurate dropped-char count consistent with the actual slice", () => {
+    const text = "x".repeat(50_000);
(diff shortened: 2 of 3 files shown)


gwitwer:
Closing in favour of #597, which fixes the same capText budget overrun and was opened first. Thanks.

PR #672, opened 2026-08-21
docs(tools): models.run model doc matches warn-and-honor behavior + surface result warnings (SAP-2765)

gwitwer:
## Summary

Companion to a server-side fix (internal tracker): the server now honors a supplied `model` as a routing label (known label → routed; unknown → platform default + a warning on the result — never a silent drop).

`ModelRunSpec.model` promised "Override the model / routing alias" while the platform silently ignored it — a doc-vs-behavior mismatch. This PR fixes the doc and surfaces the platform's warning on the result.

## What changed

- **Doc fix:** `ModelRunSpec.model` doc-comment now states the actual contract — honored as a routing label when known, warned-and-defaulted when not, omit to let the platform choose.
- **Surface the warning:** additive `ModelRunOutcome.warnings?: string[]`, mapped from the wire with a runtime guard. ONE encoding of "no warnings" reaches consumers: the key is **absent** unless at least one string warning survives the guard (missing key, wire `[]`, non-array, and all-junk arrays all map to absent) — treat `undefined` as none, per the JSDoc and changeset. Without the field the whitelist mapper would drop the platform's warning, defeating "warning surfaced on the result" for the SDK audience.
- Changeset: `@sapiom/tools` minor.

## Testing

- `packages/tools`: models suites 32 passed (new: warning passthrough; single no-warnings encoding across missing key / wire `[]` / non-array / mixed arrays); `tsc --noEmit` and `eslint` clean; `pnpm build` clean.

https://claude.ai/code/session_01B6TdePdocPc7sfYxKegs3Y

### DIFF
--- a/.changeset/honest-model-label-warnings.md
+++ b/.changeset/honest-model-label-warnings.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": minor
+---
+
+`models.run`: `ModelRunOutcome` gains an optional `warnings` array surfacing routing/honesty warnings when the platform reports them on the run result — e.g. a supplied `model` value the platform didn't recognize, which (with the SAP-2765 platform-side change) routes via the platform default instead of being silently dropped. Treat absent as no warnings.
--- a/packages/tools/src/models/index.ts
+++ b/packages/tools/src/models/index.ts
@@ -622,9 +622,15 @@ function mapModelResult(r: ModelWireResult | null | undefined): ModelRunOutcome 
     lane: r.lane ?? null,
     durationMs: r.duration_ms,
     costUsd: r.cost_usd,
-    // Passthrough with a runtime guard (the wire is untyped at runtime); the
-    // key stays absent when there are no warnings — `undefined` means none.
-    ...(Array.isArray(r.warnings) ? { warnings: r.warnings.filter((w): w is string => typeof w === "string") } : {}),
+    // Passthrough with a runtime guard (the wire is untyped at runtime). ONE
+    // encoding of "no warnings" reaches consumers: the key is absent unless at
+    // least one string warning survives the guard — a wire `[]` (or a
+    // non-array, or an all-junk array) maps to absent, matching the documented
+    // "treat undefined as none".
+    ...(() => {
+      const warnings = Array.isArray(r.warnings) ? r.warnings.filter((w): w is string => typeof w === "string") : [];
+      return warnings.length ? { warnings } : {};
+    })(),
     usage: {
       inputTokens: r.usage?.input_tokens ?? 0,
       outputTokens: r.usage?.output_tokens ?? 0,
--- a/packages/tools/src/models/index.ts
+++ b/packages/tools/src/models/index.ts
@@ -481,8 +481,10 @@ export interface ModelRunSpec {
   /**
    * Routing label for the run's LLM calls (e.g. `"smart"`). The platform
    * resolves it against its configured label set — a raw provider model id
-   * is never honored. Omit to let the platform choose (the recommended
-   * default); pass `"smart"` if you need to pin explicitly.
+   * is never honored. An unrecognized value is never silently dropped: the
+   * run routes via the platform default and the platform reports it in the
+   * result's `warnings` (SAP-2765). Omit to let the platform choose (the
+   * recommended default); pass `"smart"` if you need to pin explicitly.
    */
   model?: ModelLabel;
   /** Max output tokens per turn. */
@@ -509,6 +511,13 @@ export interface ModelRunOutcome {
    * `undefined`/`null` means not disclosed — same caveats as `servedClass`.
    */
   lane?: string | null;
+  /**
+   * Routing/honesty warnings reported by the platform (SAP-2765) — e.g. a
+   * supplied `model` the platform didn't recognize (the run then routed via
+   * the platform default). Treat `undefined` as none on any path: the wire
+   * omits the key on a clean run and the stub never sets it.
+   */
+  warnings?: string[];
   durationMs: number;
   costUsd: number;
   usage: CodingRunUsage;
@@ -576,6 +585,8 @@ interface ModelWireResult {
   model_used: string | null;
   served_class?: string | null;
   lane?: string | null;
+  /** Present only when the run has warnings (e.g. an unhonored `model` pin); guarded at map time. */
+  warnings?: string[];
   duration_ms: number;
   cost_usd: number;
   usage?: {
@@ -611,6 +622,9 @@ function mapModelResult(r: ModelWireResult | null | undefined): ModelRunOutcome 
     lane: r.lane ?? null,
     durationMs: r.duration_ms,
     costUsd: r.cost_usd,
+    // Passthrough with a runtime guard (the wire is untyped at runtime); the
+    // key stays absent when there are no warnings — `undefined` means none.
+    ...(Array.isArray(r.warnings) ? { warnings: r.warnings.filter((w): w is string => typeof w === "string") } : {}),
     usage: {
       inputTokens: r.usage?.input_tokens ?? 0,
       outputTokens: r.usage?.output_tokens ?? 0,
--- a/packages/tools/src/models/index.ts
+++ b/packages/tools/src/models/index.ts
@@ -558,6 +558,18 @@ export type ModelRunResultPayload = ModelRunResult;
 /** Thrown by {@link modelRunResultSchema}.parse on a malformed resume payload. */
 export class ModelRunResultSchemaError extends Error {}
 
+/**
+ * Guarded `warnings` normalization — enforces the ONE encoding of "no
+ * warnings" (`undefined`) on every path that hands a {@link ModelRunOutcome}
+ * to a consumer: only string elements survive; an empty/absent/malformed value
+ * becomes `undefined`, never a present-but-empty array a consumer's
+ * `if (outcome.warnings)` would misread.
+ */
+function normalizeWarnings(value: unknown): string[] | undefined {
+  const warnings = Array.isArray(value) ? value.filter((w): w is string => typeof w === "string") : [];
+  return warnings.length ? warnings : undefined;
+}
+
 /** Runtime validator for {@link ModelRunResultPayload}. */
 export const modelRunResultSchema = {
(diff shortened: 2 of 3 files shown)


gwitwer:
All four findings addressed in f3c1c8de — thanks, #1 and #2 were exactly right for a published surface.

**#1 (invariant not upheld) — fixed by making the contract honest instead of propping up the invariant.** `warnings` is now genuinely optional: absent = none. The HTTP mapper passes it through only when the wire actually sends an array, so the type, the JSDoc, and all three producer paths (HTTP, stub, resume payload) agree without touching the stub or the resume schema. The doc now says explicitly to treat `undefined` as no warnings and names the paths that omit the key. This also means no consumer can be misled into `outcome.warnings.length` by documentation — TypeScript forces the narrow.

**#2 (unattested labels) — dropped.** The JSDoc no longer names any label; it says labels are platform-resolved against the platform's configured set and that omission is the recommended default. Nothing internal or provider-branded ships to npm from this file.

**#3 (changeset asserts unmerged backend behavior) — reworded.** The changeset and JSDoc now describe what the SDK does (the field is a routing label the platform resolves; the SDK surfaces `warnings` when the platform reports them) rather than asserting the companion change's runtime behavior. Sequencing note for the human merging: the server-side companion change should still deploy before or with the next `@sapiom/tools` release so the doc-comment describes live behavior.

**#4 (test gaps) — covered by the new contract + a guard test.** With absent-means-none there is no stub/resume invariant left to test; instead there's a new test that a malformed wire value (`warnings: "oops"`) is dropped entirely and never leaks into the typed outcome (the second nit), and the fixtures are now neutral (`not-a-known-label`, `warn-1`/`warn-2` — first nit).

gwitwer:
Round-2 addressed in da5b030d.

**Resume-payload claim — dropped.** You're right that nothing enforces it and the platform likely serializes the same result object into the resume signal. The doc now states only what holds on every path: treat `undefined` as none; the wire omits the key on clean runs and the stub never sets it. No claim about resumed-step payloads either way — a resumed step that checks `warnings` will see them if the platform sends them.

**Changeset — made self-qualifying rather than SDK-only.** It now attributes the routing behavior to "the platform-side change" in the text itself, so the published changelog carries its own sequencing qualifier instead of relying on a PR comment the merger won't see. I kept the behavior description in (rather than scoping to pure SDK mechanics) because the acceptance criterion is that this doc-comment matches the platform contract — stripping it would re-create a `model` field whose semantics are documented nowhere. Agreed the final deploy-before-release ordering is the merger's call; the server-side companion change is green and ready.

gwitwer:
Heads-up + one doc alignment in cf21aa1c: a contract change was posted on the server-side companion change (now merged, with a pivot follow-up in progress): the customer-selectable vocabulary is model CLASSES (`smart`, and size classes as they land) — a provider model id is never selectable. The `ModelRunSpec.model` JSDoc now says exactly that, with `smart` as the example (attested public vocabulary per the platform's SKU contract, addressing the earlier concern about unattested labels — this one is the contract's own word). No shape or behavior change; `warnings` semantics unchanged.

gwitwer:
Applied the finding from the review above (20dfe127): reverted the `ModelRunSpec.model` JSDoc from "Model class" back to "routing label" — matches the changeset (which already said "label"), the already-shipped `llm.run` CHANGELOG entry, and the platform's corrected model-execution contract (internal). Kept the accurate "a provider model id is never selectable" sentence.

Re-ran: `cd packages/tools && npm run gen:version && npx jest src/models` → 27 passed, 3 suites. `tsc --noEmit` and `eslint src/models` clean. Changeset file needed no change — it already said "routing label."

Mergeable as-is now.

gwitwer:
**Review verdict: needs changes** (GitHub blocks a formal "Request changes" review from the PR author's own account, so posting as a comment instead.)

Reviewed against what actually shipped server-side (internal tracker) and the current platform contract. Ran the package's tests, `tsc --noEmit`, `eslint`, and `pnpm build` locally (had to build the `@sapiom/analytics-core` workspace dependency first — its `dist/` wasn't present, same class of pre-req the server-side companion changes for this effort have also needed).

## 1. `ModelRunSpec.model` doc-comment — needs changes

The latest commit (`cf21aa1c`, "contract pivot") reframes the field as a **"Model class"** ("e.g. `smart`" — "a recognized class pins routing"), citing the platform's internal SKU-contract spec. That spec is explicitly marked **"DRAFT / proposed authority"** and its own "Current reality" section states plainly that there is no public size/lane input yet — the SDK is still `models.run({prompt})`, no `size`, no `lane`. It's describing a **future, unshipped** contract, not what the server-side companion change actually merged.

What actually shipped server-side is a **known-label allowlist** (the server-side label allowlist, internal) — a routing label, not a size class. Most of the values in that allowlist are model-family/version names, not `small`/`medium`/`large`-style classes. Calling this "a model class" risks exactly the doc-vs-behavior mismatch this effort exists to fix — a caller reading "Model class" could reasonably try passing `medium` (not a valid label today) or be confused about why the other allowlisted values are valid "classes."

This is also now **inconsistent within the PR itself**: the changeset (`.changeset/honest-model-label-warnings.md`) still says *"the field is a routing label the platform resolves against its configured label set"* — "label," not "class." And it's inconsistent with the rest of the package: the existing (already-shipped) `llm.run` CHANGELOG entry says *"`model` names a Sapiom routing label"* for the identical mechanism. And the platform's corrected model-execution contract (internal, recently updated) also settles on **"label"** as the author-facing term ("A requested model string is a **label** naming a class...").

**Recommend:** revert the JSDoc to "routing label" (matching the changeset, `llm.run`'s prior art, and what actually shipped server-side), and drop the "model class" framing until the size/lane SKU contract actually ships to the SDK. The "provider model id is never selectable" claim is accurate and worth keeping regardless of which term wins.

(Nit, not blocking on its own: an earlier review round on this PR deliberately dropped specific label examples as "provider vocabulary [that] shouldn't ship... to npm" — the pivot re-added `smart` as an example. If "label" wins, worth deciding intentionally whether to keep that example rather than as a side effect of the class/label back-and-forth.)

## 2. Warnings plumbing — correct

Verified the wire field name against the server's actual response shape: the `warnings` array is nested inside the result object exactly as `ModelWireResult.warnings` / `mapModelResult` expect. No mismatch. The runtime guard (`Array.isArray(r.warnings)` + a `string`-filter) is good defensive coding against the wire being untyped at the boundary, and the "malformed wire value never reaches the typed outcome" test covers it well. The resume-payload wording fix from round 2 (dropping the unattested "resumed-step payloads omit the key" claim) is also correct — it now only asserts what's true on every path.

## 3. Breaking changes — none

`ModelRunOutcome.warnings?: string[]` is additive-optional; existing consumer-constructed values still type-check. `ModelWireResult`/`ModelRunDoc`/`mapModelResult` are internal (not exported) — no public surface change there. Clean.

## 4. Build/tests

- `cd packages/tools && npm run gen:version && npx jest src/models` → **27 passed**, 3 suites (PR body says 26 — one more now, harmless, likely from the round-2/3 fixup commits).
- Full package: `npx jest` → **566 passed**, 29 suites.
- `npx tsc --noEmit` → clean.
- `npx eslint src/models` → clean.
- `npm run build` → clean.
- Needed `cd packages/analytics-core && npm run build` first — `@sapiom/tools/src/_client/analytics.ts` imports `@sapiom/analytics-core`, whose `dist/` wasn't present in a fresh worktree.

## Verdict

**Needs changes** — functionally and wire-contract-wise this is correct and well-tested (items 2–4 are clean), but the "model class" doc-comment pivot describes a draft/unshipped contract instead of what actually shipped server-side, and it's now internally inconsistent with this PR's own changeset. Recommend reverting that one doc-comment to "label" terminology before merging.

gwitwer:
Both findings addressed in 24b1439e.

**#1 (two encodings of "no warnings") — fixed.** The mapper now emits the key only when at least one string warning survives the guard, so `undefined` is the single no-warnings encoding on every path: missing key, wire `[]`, non-array, and all-junk arrays all map to absent. `if (outcome.warnings)` can no longer render an empty banner based on which empty encoding the server chose. The new test covers all four shapes, including the mixed-type array (`[1, "warn-a", null]` → `["warn-a"]`) that previously had no coverage.

**#2 (stale PR description) — corrected.** The body now matches the code, the changeset, and the tests: absent-means-none, key emitted only for real warnings, 32 tests. The "normalized to []" / "always populates" claims were left over from the pre-round-1 design and are gone.

gwitwer:
Follow-up addressed in bdec9509; the two declines below with reasoning.

**Resume-path normalization — fixed (your first option).** Both paths now share one `normalizeWarnings` guard: `modelRunResultSchema.parse` applies the same encoding as `mapModelResult` (only string elements survive; empty/absent/malformed → the key is deleted from the payload's result), so "treat `undefined` as none on any path" is now enforced rather than asserted. New test mirrors the four-case matrix on the resume payload. Agreed with your correction of the earlier framing — the broader unvalidated cast on `ModelRunOutcome` predates this PR and stays out of scope; only the warnings encoding this PR introduced is normalized.

**`CodingRunSpec.model` — deliberate non-change, not an oversight.** SAP-2765 names `ModelRunSpec.model` specifically; the coding surface is a different execution path (sandboxed coding runs, own routing flags and result shape) whose backend does not emit `warnings` today. Adding the warn-and-default sentence or a `CodingRunOutcome.warnings` field ahead of that backend behavior would recreate the exact doc-vs-behavior mismatch this ticket exists to fix, on the surface it didn't audit. Its `model` doc does deserve the same honesty pass once the coding backend's label handling is settled — flagging that as a follow-up ticket candidate rather than folding it in here.

**`warnings: string[]` vs `{code, message}` — declining for now.** The live wire (backend merged in sapiom/Sapiom#4480) emits strings, so a structured type here would either fabricate codes client-side or mismatch the server. The typed evolution of this vocabulary is SAP-2768's structured `Degradation` annotation (code + reason, emitted from all recovery paths); when that lands, a structured field can ship alongside `warnings` additively. Committing to `{code, message}` now would mean inventing a code taxonomy in the SDK ahead of the platform's.

(CI note: the `test (20.x/22.x)` failure on the previous run is the known `packages/harness generated-retention.test.ts` ENOTEMPTY teardown flake — unrelated to `packages/tools`, 2094/2095 passing; this push retriggers.)

PR #681, opened 2026-08-23
fix(examples): Content Pack renders the quote into its graphics and ships a coherent teaser clip (SAP-2781)

gwitwer:
Fixes [SAP-2781](https://linear.app/sapiom/issue/SAP-2781/templates-content-pack-ships-a-blank-quote-graphic-no-text-ever-added) — a follow-up in the media-reliability epic (SAP-2787).

## What was wrong (evidence on the ticket)

The platform worked end to end; the template was the bug:

1. **The quote text never got added.** The LLM's `imagePrompt` asked for a background with *"no text in the image"*, no compositing step exists, and the blank navy background shipped as the finished "quote graphic".
2. **The clip prompt was the wrong artifact.** The full 45-second narration script (`videoScript`) went to a 5-second image-to-video model with the blank background as keyframe → hallucinated gibberish text panels.
3. **Aspect ratio ignored** — no `aspectRatio` on the image launch, and the raw uncataloged Kling id meant no param normalization ever ran.
4. **Plain-text email** — even a good graphic never rendered in an inbox.

## The fix (`examples/content-repurposing-pipeline`)

- **The quote is now composed into the image launch prompt in code** (`buildGraphicPrompt`), never trusted to the LLM, and graphics render on the typography-capable cataloged alias **`ideogram-v3`** instead of the default fast model. The LLM's `imagePrompt` is demoted to art direction only (system prompt updated to match).
- **The clip is a cataloged semantic alias** (**`kling-standard`**, silent 5s text-to-video) driven by a **purpose-written short visual prompt** (`buildClipPrompt`); `isNarrationScript` guards against narration-shaped LLM output. The teaser is deliberately **decorative and text-free** — general video models render on-screen text illegibly, so the quote lives in the typography-model graphics and the clip stays abstract. This also removes the raw `fal-ai/kling-video/v2.1/pro/image-to-video` pin that allowlist enforcement (**SAP-2582/E8**) will reject with `400 unknown_model` — a prerequisite for closing the allowlist.
- **`aspectRatio: "16:9"`** set on the image launch and shared with the clip (neutral E4 params, validated server-side). `duration` is pinned only on the default alias, since duration vocabularies differ per model and an unsupported value is a 400.
- **HTML email body** (`renderPackHtml`) so the graphics embed inline in the inbox; markdown stays as the text/plain fallback.

Note: the clip is now **text-to-video** rather than image-to-video — at the time of writing the semantic catalog has no alias advertising `referenceImage` (verified against the platform's video model catalog on its latest default branch), so "animate the finished graphic" is not currently expressible without a raw provider id.

## Tests (intent asserted, per the ticket)

17 tests pass (`tsx --test`): the quote must appear in the image launch prompt; a narration-shaped script must never reach the video model; the clip prompt must request no on-screen text; the model must be a cataloged alias (no `fal-ai/` ids, no `passthrough`); neutral params must be set; the email must embed the graphics as ``.

## Sibling media template audit (second commit)

- **personalized-media-at-scale** — added an HTML body so the personalized image renders inline (was a bare URL in plain text). Video path already used a cataloged alias + neutral `aspectRatio`; its no-text prompts are deliberate (decorative marketing media).
- **newsletter-autopilot** — pinned the header banner to `aspectRatio: "16:9"`. Already sends HTML; the no-text header illustration is intentional.
- **scene-to-video** — its keyframe→animate design *requires* image-to-video, which the catalog does not offer yet, so the raw Kling pin stays but is now explicitly flagged at the pin as an SAP-2582/E8 hard break to repoint the moment the catalog grows an i2v alias.

## Verification

- `tsc --noEmit` clean on all three touched examples
- `tsx --test index.test.mjs`: 17/17 pass
- `pnpm examples:check` + `pnpm terminology:check` pass
- Prettier clean on touched files

https://claude.ai/code/session_01W2bwKiedonBzeqRH3Um2o8

### DIFF
--- a/examples/content-repurposing-pipeline/AGENTS.md
+++ b/examples/content-repurposing-pipeline/AGENTS.md
@@ -24,8 +24,9 @@ repurpose ─▶ graphics ⇄ collectGraphic ─▶ clip ⇄ collectClip ─▶ 
   'collectGraphic' }`); `collectGraphic` records it and loops back for the next quote or
   advances once every graphic is in. Sequential, not a concurrent `Promise.all` — see
   "Why sequential" below.
-- **clip ⇄ collectClip** — launches an async image-to-video job from the first
-  graphic and `pauseUntilSignal`s on it. The
+- **clip ⇄ collectClip** — launches an async text-to-video job (a cataloged
+  semantic alias, with a purpose-written short visual prompt — never the LLM's
+  narration-shaped output; see `buildClipPrompt`) and `pauseUntilSignal`s on it. The
   `pause: { signal: VIDEO_RESULT_SIGNAL, resumeStep: 'collectClip' }` declaration is
   the graph edge; the video-generation webhook fires the signal to resume `collectClip`.
 - **package** — renders the whole pack as one markdown doc and uploads it to file
--- a/examples/content-repurposing-pipeline/README.md
+++ b/examples/content-repurposing-pipeline/README.md
@@ -19,11 +19,11 @@ repurpose ──▶ graphics ⇄ collectGraphic ──▶ clip ⇄ collectClip �
 2. **graphics ⇄ collectGraphic** — one quote-graphic image at a time
    (`images.launch`, async): launch the job, pause until the webhook resumes
    `collectGraphic`, record it, then loop back for the next quote or advance.
-3. **clip ⇄ collectClip** — renders a short teaser clip that puts the lead
-   pull-quote on screen: launches an async text-to-video job (`video.launch`,
-   a cataloged semantic alias) with a purpose-written short visual prompt and
-   pauses on it; the video-generation webhook resumes `collectClip` when the
-   clip is ready.
+3. **clip ⇄ collectClip** — renders a short decorative teaser clip: launches an
+   async text-to-video job (`video.launch`, a cataloged semantic alias) with a
+   purpose-written short visual prompt — deliberately no on-screen text, since
+   general video models render text illegibly — and pauses on it; the
+   video-generation webhook resumes `collectClip` when the clip is ready.
 4. **package** — assembles the whole pack as one markdown document and uploads it
    to file storage (`fileStorage.upload`) for a durable `fileId` + download URL.
 5. **deliver** — fans the pack out to every `deliverTo` recipient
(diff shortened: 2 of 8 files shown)


gwitwer:
Review round 1 addressed in 03c122ae. Point by point:

**Confidentiality** — ✅ fixed. The PR body no longer carries employee names (rewritten to roles) or the production execution id. The execution id is also dropped from the code comment and both test titles; SAP-#### ticket ids stay, per the precedent this review notes.

**1. `ideogram-v3` attestation** — verified, no change. Both `ideogram-v3` and `gpt-image-2` are cataloged aliases in the platform's image model catalog (on its latest default branch — `ideogram-v3` is documented there as "text-to-image with strong typography / logo rendering", and its caps include `16:9`). They aren't attested in *this* repo because the SDK's doc comment lists aliases as examples ("e.g."), not exhaustively. On the fallback suggestion: deliberately not added. Silently falling back to the default fast model on an unknown-model error would ship an illegible or blank "quote graphic" while reporting success — the exact failure this ticket fixes, now invisible. A loud launch failure is the honest outcome, and content-level smoke coverage for this template is its own ticket in the same epic (SAP-2783).

**2. Text-to-video rendering text** — ✅ agreed and fixed. This was the strongest catch: the fallback prompt asked a general video model for "bold, legible type", which is the garbled-clip failure through a different door. The teaser is now deliberately decorative and text-free (fallback prompt requests "no text, no watermark", matching `personalized-media-at-scale`'s convention), the LLM's `videoScript` instruction now demands no on-screen text, and the tests assert the fallback contains `no text` and never hands the quote string to the video model. The quote lives in the typography-model graphics, where it's legible.

**3. Catalog-state claims in published docs** — ✅ softened. The model-constant comment and README now say "at the time of writing no cataloged alias advertises `referenceImage`" and state allowlist enforcement in the future tense ("will reject… once it lands" / "is expected to reject").

**Nits:**
- *Fallback >300 chars* — ✅ resolved as a side effect of #2: the fallback is now a fixed-length constant with no quote interpolation.
- *`renderPackHtml` fileId-without-downloadUrl* — defensive-only branch, leaving as is. It's unreachable at deliver time: `collectGraphic` mints `downloadUrl` whenever `fileId` exists (and throws when neither exists), and `package` re-mints for any graphic still missing one before `deliver` runs. The `<em>` fallback exists so the function stays total rather than throwing mid-send.
- *`escapeHtml` duplication* — deliberate, leaving as is. `examples/` templates are self-contained by design (each is a standalone package users copy; there's no shared module between examples, and `newsletter-autopilot` already had its own copy before this PR). Extracting a shared helper would couple the templates.

gwitwer:
Round 2 addressed in 9c12b052:

**1. No-text guarantee on the LLM path** — ✅ agreed and fixed. `buildClipPrompt` no longer returns an accepted `videoScript` verbatim: a script that doesn't already forbid text gets `". No text, no watermark."` appended, so the JSDoc claim now holds on both paths in code, not just via the system prompt. Chose append over reject: rejecting on keywords like quote/text/type would throw away an otherwise-good visual prompt (and "type" false-positives easily), while the appended directive is the same mitigation the fallback and `personalized-media-at-scale` already rely on. Tests updated — the stale acceptance fixture no longer models a text ask, and a new test asserts the pull-quote-asking script gets the directive appended (18/18 pass).

**2.** Acknowledged — residual risk stands as you stated it; the content-level template smoke (SAP-2783, same epic) is the backstop that would catch an alias rename breaking the sample run.

**3. Internal path in my reply** — ✅ fixed; the earlier comment now says "the platform's image model catalog" with no repo layout.

PR #673, opened 2026-08-21
feat(tools): expose servedModel / costUsd disclosure on execution results (SAP-2764)

gwitwer:
## Summary

SDK half of a corrected server-side contract (internal tracker): expose the server's serving disclosure — **the billing class + lane the request resolved to**, in the SKU vocabulary the platform bills in. Never a model or provider id, never a provider price (an earlier draft's `servedModel`/provider `costUsd` are gone per the corrected contract, shipped as a server-side companion change).

## What changed

- **`models.run`** — `ModelRunOutcome.servedClass` + `lane` (wire `served_class`/`lane`): the billing class (size) the run's label resolved to and the lane it executed in; `modelUsed` keeps meaning the requested label. **`costUsd` is deprecated and now `number | null`**: servers report null (the customer dollar amount is the metering pipeline's, summed on run views); older servers' numbers still map through.
- **`models.coding.run`** — `CodingRunOutcome.servedClass`/`lane`/deprecated `costUsd` (all null from coding servers today; null = unknown, never fabricated).
- **`llm.run` / `llm.redeem` / `llm.callSession`** — `LlmDisclosure` now describes `served_class`/`lane` (injected top-level into raw /v2 non-streaming bodies; streams carry the same data as `x-sapiom-served-class`/`x-sapiom-lane` headers); `readDisclosure()` returns the camelCase `LlmDisclosureResult { servedClass, lane }`. The response `model` field keeps echoing the label.
- Reserved `degradation` (typed `unknown`) unchanged.
- Minor changeset rewritten, calling out the `costUsd` `number` → `number | null` widening.

## Compatibility

- All new fields optional/nullable → existing consumers compile/run unchanged; old servers map missing fields to `null` — covered on both mappers + `readDisclosure`.

## Testing

- `packages/tools`: jest green (mapping incl. class/lane, old-server absent-fields, null-from-coding-servers, `readDisclosure` happy/absent/malformed); `pnpm build` clean; provider-neutral copy check passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SZmJ79CfTQ4M1EeY5RD9s9

PR #682, opened 2026-08-23
feat(tools): served class/lane disclosure, structured output, label ergonomics (SAP-2764)

gwitwer:
## Summary

Reissues #673 (closed — the disclosure vocabulary changed under it, twice: first `servedModel`→`servedClass`/`lane` mid-review, then the whole PR was closed to land as part of this broader ergonomics PR) under the corrected disclosure contract, plus the strong-defaults ergonomics items from the same plan. Contract authority: the platform's model-execution contract (internal) — server side landed via a companion server-side change.

**Hard constraint honored: additive only.** No published type or behavior changes for existing consumers — verified with `tsc --noEmit`, full `jest`, and `pnpm build`, all clean.

## What changed (packages/tools)

1. **Disclosure types.** `models.run` (`ModelRunOutcome`) and `models.coding.run` (`CodingRunOutcome`) gain optional `servedClass`/`lane` (wire `served_class`/`lane`) — the billing class the run's label resolved to and the lane it executed in. `llm.run`/`redeem`/`callSession` gain a new `LlmDisclosure` wire-shape type + `readDisclosure()` helper returning the camelCased `LlmDisclosureResult`. All `undefined`/`null` on older servers or where the platform doesn't disclose — never fabricated.
2. **Structured-output convenience for `llm.run`.** New optional `output: { name, schema }` on `LlmRunSpec` — automates the blessed tool-calling pattern (appends a forced tool + `tool_choice`). `run()`'s return type is unchanged either way. New `structuredOf()` reads the parsed value back out; new `textOf()` reads the plain-text reply, correctly skipping a `thinking` block that may precede it.
3. **Label ergonomics.** `model`/`label` fields across `llm.run`, `llm.submit`, `llm.createSession`, `models.run`, `models.coding.run` are now a soft union (`"smart" | (string & Record<never, never>)` — the lint-safe spelling of `string & {}`, matching `content-generation/index.ts`'s existing `LiteralUnion`) for autocomplete, with JSDoc settled on "routing label" terminology: omit to let the platform choose, `"smart"` if you must pin, a raw provider model id is never honored.

**Explicitly out of scope** (per the plan): no `deadlineMinutes` on `models.*`/`agents.*` (server side, hasn't shipped) and no `inspectStep` (separate PR). Also dropped `#673`'s `degradation` reservation on all three result types — its round-2 review left that unresolved (untyped reserved surface for an unshipped feature); out of scope here, can land with a follow-up.

## Design choice — structured output (deliverable #2)

**Chosen:** additive `output?: { name, schema }` on `LlmRunSpec`, which only builds the request (injects the tool + forces `tool_choice`); `run()`'s return type is **never** touched — still `Promise<T>`, the verbatim response, for every caller regardless of whether `output` is set. Reading the parsed value is a second, explicit step via the new `structuredOf(response, name?)` helper (paired with `textOf()` for the plain-text case).

**Rejected: TS overloads that return the parsed object directly when `output` is set.** Better single-call ergonomics, but overload resolution keyed on an optional property inside an object-literal parameter is fragile in TypeScript — it would need a conditional generic (`S extends { output: infer O } ? Parsed<O> : T`) to make `run()`'s return type branch on the shape of `spec`, and that's exactly the kind of change most likely to regress an existing caller's inference in some edge case I can't fully enumerate. Given the hard "zero breaking changes" constraint, I chose the design with no way to touch the existing contract at all over the one with better ergonomics but non-zero regression risk.

## Wire-shape verification (deliverable #1)

Verified against a fresh pull of the current server-side implementation rather than trusting an older checkout:

- `models.run`'s wire response: `served_class`/`lane` always present as keys (null when unresolved), plus `cost_usd` — confirmed real, non-null size×lane estimate in the common case (deprecated as an *authoritative* customer figure, not nulled server-side). **I left `ModelRunOutcome.costUsd` completely untouched** (still `number`, non-nullable) — widening it to `number | null` is exactly the kind of type change the additive-only mandate rules out, even though the server does occasionally send `null` for a narrow legacy-row case (rows persisted during a brief prior live window). That's a pre-existing type/runtime mismatch in the current published type (not introduced by this PR) — flagging it rather than fixing it, since fixing it properly needs the exact widening this constraint forbids.
- `models.coding.run`'s wire response: `served_class`/`lane` always present (null today — coding can't observe them), and confirmed **no `cost_usd` key at all** — coding never had a real cost signal to preserve, unlike `models.run`.
- The gateway backing `llm.run`/`redeem`/`callSession`: injects `served_class`+`lane` top-level into buffered 2xx JSON bodies (confirmed no `cost_usd`, no `usage` injection — `usage` is already native to the Anthropic Messages response, nothing to add there).
- **Caveat, can't verify from here:** server-side rollout of the disclosure fields may lag this SDK change; fields are optional and absent until served, so `served_class`/`lane` may not appear on live responses immediately even though both the server and this SDK now support them.

## Testing

- `packages/tools`: `npm run test` → **32 suites, 583 passed** (564 baseline + 19 new: disclosure mapping for both old- and new-server shapes on `models.run`/`models.coding.run`, `readDisclosure` happy/absent/malformed, `textOf`/`structuredOf` including the thinking-block-skip case, `output`'s request-building — non-mutation of the caller's object, merging with existing tools, unchanged behavior when omitted, end-to-end round trip).
- `npm run typecheck` clean, `npm run lint` clean, `npm run build` clean, `node scripts/provider-neutral-copy-check.mjs` passes (74 audited files) — had to reword one JSDoc line to reuse an already-allowlisted phrase (`"Anthropic Messages"` exact match) rather than a new "Anthropic's ___" construction the checker doesn't recognize.

Refs: internal tracker

### DIFF
--- a/.changeset/served-class-lane-disclosure.md
+++ b/.changeset/served-class-lane-disclosure.md
@@ -1,0 +1,12 @@
+---
+"@sapiom/tools": minor
+---
+
+Execution results now expose the server's serving disclosure, in SKU vocabulary — plus structured-output and routing-label ergonomics for `llm.run`. Reissues #673 under the corrected SAP-2764 contract (`servedClass`/`lane`, not the earlier draft's `servedModel`/provider `costUsd`).
+
+- `models.run` (`ModelRunOutcome`) and `models.coding.run` (`CodingRunOutcome`): new optional `servedClass` + `lane` (wire `served_class`, `lane`) — the billing class (size) the run's label resolved to and the lane it executed in. Never a model or provider id. `undefined`/`null` on older servers or when coding cannot observe it — never fabricated.
+- `llm.run` / `llm.redeem` / `llm.callSession`: new `LlmDisclosure` type describing the `served_class` / `lane` fields the server injects top-level into raw `/v2` non-streaming response bodies, plus a `readDisclosure()` helper returning the camel-cased `LlmDisclosureResult`. The response `model` field is unchanged and keeps echoing the requested label.
+- `llm.run` gains an optional `output: { name, schema }` field — the blessed tool-calling pattern for structured output, automated: it appends a forced tool call to the request and forces `tool_choice` onto it. `run`'s return type is unchanged either way (still the verbatim response); read the parsed value with the new `structuredOf()` helper. A new `textOf()` helper reads the plain-text reply, correctly skipping a `thinking` block that may precede it.
+- `model`/`label` fields across `llm.run`, `llm.submit`, `llm.createSession`, `models.run`, and `models.coding.run` are now typed as a soft union (`"smart" | (string & {})`, lint-safely spelled) for autocomplete, with JSDoc settled on "routing label" terminology: omit to let the platform choose, `"smart"` if you must pin, a raw provider model id is never honored.
+
+All additions are optional: existing consumers compile and run unchanged. On results from older servers the mappers and `readDisclosure` return `servedClass`/`lane` as `null` (unknown).
--- a/packages/tools/src/llm/disclosure.spec.ts
+++ b/packages/tools/src/llm/disclosure.spec.ts
@@ -1,0 +1,36 @@
+import { readDisclosure } from "./index.js";
+
+// Serving disclosure on raw /v2 non-streaming bodies: `served_class` + `lane`
+// are injected top-level by the gateway (SKU vocabulary — never a model or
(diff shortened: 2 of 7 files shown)


gwitwer:
Consolidated into #685 — content unchanged (merged cleanly with #683's changes to the same file, both preserved). Closing this in favor of that.

PR #503, opened 2026-08-04
[SAP-2365] Clarify cloud operation contracts

ynadge:
## Summary

- define deploy as a bundle of current reachable working-tree source, including uncommitted changes, pushed through a synthesized build tree
- expose exact `buildRunId` language throughout Agent Core and CLI guidance instead of treating Git HEAD as the release identity
- distinguish metered cloud builds and production runs from the account-free local scaffold/check/Local Run boundary
- make execution inspection explicitly cost-agnostic and name the returned attempts, evidence, and lineage
- synchronize the canonical authoring skill across Agent Core, bundled starters, and the Sapiom plugin
- add a patch changeset and an MCP identity regression

## Why

SAP-2365 dogfooding proved that deploy includes uncommitted reachable source, that a production run pins the resulting build, and that execution audit does not include the dashboard's separate run-grain charge. Existing product guidance contradicted each of those implemented boundaries and described the whole developer MCP as unmetered.

## Dogfood evidence

- deployed a dirty default-starter checkout through the locally built MCP
- observed the uncommitted marker in the completed production output
- correlated the returned ready build ID with the execution audit and dashboard version list
- verified separate run usage and overage reads without exposing account credentials or organization identity

## Verification

- Agent Core: 158 tests passed across 16 suites
- CLI: 73 tests passed across 6 suites
- MCP local-contract regression: 1 test passed
- Agent Core, MCP, CLI, and Harness typechecks passed
- affected-package lint: 0 errors; one pre-existing Harness warning
- terminology check passed across 373 files with no stale allowlist entries
- `git diff --check` passed

Linear: [SAP-2365](https://linear.app/sapiom/issue/SAP-2365/docs-publish-deploy-production-run-and-inspection-guides)

PR #163, opened 2026-07-07
[SAP-1365] Agent-authoring skill v2 + scaffold templates ship it + instructions fallback sync

ynadge:
## ⚠️ Held draft — merge only in the agents/models rename train

Written against the **locked naming map** (`defineAgent`, `@sapiom/agent`, `sapiom_dev_agents_*`, `ctx.sapiom.models.coding`, executions). This branch is intentionally inconsistent with pre-rename `main`; it rebases onto the rename branch and merges in the same release window.

## What

1. **Canonical skill** `packages/orchestration-core/skills/sapiom-agent-authoring/SKILL.md` — task-shape trigger ("automate a multi-step / scheduled / deployable task…"), full authoring guide in the new vocabulary, MCP bootstrap step, capability *pointer* instead of a catalog enumeration (`ctx.sapiom.*` grew 8→11 namespaces in 10 days — enumerations rot).
2. **Both scaffold templates ship it** at `.claude/skills/sapiom-agent-authoring/` (auto-loads as a Claude Code project skill) + one `AGENTS.md` pointer line each. npm-pack dot-dir inclusion verified empirically; zero scaffold-code change (plain `cpSync`).
3. **`@sapiom/mcp` instructions fallback** rewritten to the canonical v2 text (identical to the backend constant PR for SAP-1367) — sapiom-dev identity, two-MCP frame with `tool_discover`, thinned to lifecycle + canonical rules + pointers.
4. **Drift guard**: `skill-sync.test.ts` asserts template copies === canonical.

## Rebase checks (verify against the rename branch at merge time)

- [ ] Export name/generics: `export const agent = defineAgent<...>` matches the renamed manifest loader
- [ ] Context type name `AgentExecutionContext` + `ctx.agentName` field
- [ ] Dev-tool suffixes (`_run_local`, `_run`, `_inspect`) as David ships them
- [ ] Remote cloud-tool names/verbs in the instructions' two-MCP paragraph (execute vs run — SAP-1363 ask #3)
- [ ] `CODING_RESULT_SIGNAL` / `CodingResultPayload` import home after the `agent`→`models` namespace move in @sapiom/tools
- [ ] Docs URLs `/agents/*` live (SAP-1368 ships redirects)

## Verification

- `pnpm --filter @sapiom/mcp test` → **68/68** (9 files, incl. updated instructions assertions)
- `pnpm --filter @sapiom/orchestration-core test` → all pass incl. new `skill-sync.test.ts`
[...]

PR #683, opened 2026-08-23
docs(skills): teach the LLM call-surface rule + naming conventions

gwitwer:
## Summary
Companion to a server-side companion change — this is the FULL version of the LLM call-surface rule + naming conventions, landing in the canonical `sapiom-agent-authoring` skill (`packages/agent-core/skills/sapiom-agent-authoring/SKILL.md`) rather than the Sapiom-repo monorepo pointer. This is the **highest-reach teaching surface of the whole effort**: it ships into every scaffolded agent project via `.claude/skills/`, and into the Claude Code plugin.

## What landed

**New "## Calling LLMs from Steps" section:**
- Table: `ctx.sapiom.llm.run` (one-shot) vs `ctx.sapiom.models.run` (platform-driven multi-turn loop, `models.coding.run` for sandboxed coding — never for a one-shot, it overthinks) vs `ctx.sapiom.agents.run` (dispatch a deployed agent by slug).
- **Explicit warning** against sending single-shot intent through a multi-turn surface, with a worked "wrong vs right" example: WRONG = `models.run` + "reply with ONLY JSON" + `JSON.parse`; RIGHT = `llm.run` with a forced tool call for the fixed shape, and an explicit `type`-filtered content read (never assume `content[0]` — a `thinking` block can precede it). Notes the incoming `output`/`structuredOf`/`textOf` convenience (#682, still open) that will automate this once it ships, without depending on it now.
- The label rule: never pick a model; omit `model` (recommended) or pin `"smart"`; raw provider ids never honored; results disclose `servedClass`/`lane`.
- A debugging pointer (dashboard run detail → step row id → **Run Inspector**) and the public guide link.

**New "## Naming Conventions" section** — a table settling agent / run / task / session / dispatch, plus "label" as the author-facing term for a `model:` value, and the rule that new capabilities must not add a sixth meaning to an already-overloaded word.

**Synced across all four copies** `skill-sync.test.ts` guards: the canonical source, both scaffold templates (`default`, `coding-pause`), and the Claude Code plugin copy (`plugins/sapiom/skills/`). Verified with checksums (all four MD5-identical) and the test itself.

**`packages/tools/src/models/index.ts` stale-comment fix:** the "Default agent … `agent.run` / `agent.launch`" header named a namespace that's actually exported as `models` — fixed, along with two more instances of the same problem I found in the same file while I was in there: the top-of-file module docstring's OWN self-description (`` `agent` capability`` + a code example calling `agent.coding.run`, when the real export is `models.coding.run` — arguably the more prominent occurrence, since it's the file's own introduction) and the `Ambient-bound `agent.coding` namespace` comment right above the block that flagged the issue. **#682 (open) also touches this file** but a different, non-overlapping region (the `ModelRunSpec`/`CodingRunSpec` type + JSDoc changes) — no hunk collision either way merges first.

## Important finding — an existing terminology guard caught real mistakes in my first draft

`scaffold.test.ts`'s `"copies the %s template with exact Agent terminology"` test bans certain internal service naming (case-insensitive) from anything shipped into a scaffold. My first draft tripped it twice:
1. The Naming Conventions "agent" row explained the internal-vs-external naming split by directly quoting the internal module's own docstring, which uses different internal vocabulary for the same concept. Reworded to convey the same fact (internal source comments use different, non-public vocabulary) without repeating the banned terms.
2. The debugging pointer cited a literal internal REST endpoint path — real (it's what `packages/agent-core/src/client.ts`/`run.ts` actually call), but that internal service naming must never reach scaffolded, customer-facing content. Dropped the literal path; the pointer now just says to use the **Run Inspector**.

**This same constraint may affect four other, already-open PRs in this teaching stack** that used the identical debugging-pointer text verbatim, none of which have this specific guard test, so nothing failed there, but the underlying "don't leak internal service naming to customers" policy this test enforces may still apply:
- #679 (MCP instructions fallback)
- #680 (harness system prompt)
- two companion server-side PRs (backend MCP instructions + migration; server-side skills pointer)

Not touched here — flagging for your call rather than unilaterally reopening four already-reviewed PRs.

## Testing
```
cd packages/agent-core
npx jest skill-sync                    # 7 passed
npx jest src/__tests__/scaffold.test.ts # 19 passed (including the terminology guard)
npx jest                                # full package: 18 suites, 171 passed
```
```
cd packages/tools
npx jest src/models                     # 25 passed
npx jest                                # full package: 29 suites, 564 passed
npx eslint src/models/index.ts && npx tsc --noEmit   # both clean
```
Needed a repo-root `pnpm build` first in this fresh worktree (agent-core/tools/mcp/harness all transitively depend on several unbuilt sibling packages).

Added a changeset (`@sapiom/agent-core` + `@sapiom/tools`, both `patch`) matching this repo's existing convention for skill-content-only changes (e.g. #163, #503).

## Related
Refs: internal tracker
#679, #680 (the rest of the teaching stack in this repo)
#682 (open, touches the same file in a non-overlapping region)
Companion server-side PRs (internal, not linked here)

## Checklist
- [x] Full rule + warning box + worked example + naming conventions (the complete version)
- [x] Synced across all 4 copies `skill-sync.test.ts` guards (checksum + test verified)
- [x] `models/index.ts` stale-comment fix folded in, no overlap with #682
- [x] Terminology-guard test respected — flagged the same risk for 4 earlier PRs rather than silently reopening them

### DIFF
--- a/.changeset/teach-llm-call-surface-rule.md
+++ b/.changeset/teach-llm-call-surface-rule.md
@@ -1,0 +1,15 @@
+---
+"@sapiom/agent-core": patch
+"@sapiom/tools": patch
+---
+
+`sapiom-agent-authoring` skill: teaches the LLM call-surface rule from step
+code (`llm.run` one-shot vs `models.run` platform-driven loop vs `agents.run`
+deployed-agent dispatch, with a worked example against the "reply with only
+JSON" + string-parsing mistake) and settles the platform's naming
+conventions (the overloaded "agent"/"run"/"task"/"session"/"dispatch" terms,
+and "label" as the author-facing term for a `model:` value). Synced across
+the canonical source, both scaffold templates, and the Claude Code plugin
+copy. `@sapiom/tools`: corrected stale `agent.run`/`agent.coding` naming in
+`models/index.ts`'s doc comments — the actual exported namespace is
+`models`.
--- a/packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
+++ b/packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
@@ -305,6 +305,114 @@ try {
 }
 ```
 
+## Calling LLMs from Steps
+
+Three DIFFERENT capabilities call an LLM from step code — picking the wrong one for the
+job is the most common mistake in authored agents:
+
+| Capability              | Use for                                                                                            | Never for                                                          |
+| ------------------------ | ---------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------- |
+| `ctx.sapiom.llm.run`    | ONE LLM call — summarize, extract, classify, one-shot generate                                       | A multi-turn task, or anything needing its own tool-calling loop      |
+| `ctx.sapiom.models.run` | A platform-driven multi-turn reasoning + tool-calling loop (minutes, not seconds). `models.coding.run` for sandboxed coding tasks. | A one-shot completion — it will loop and overthink                    |
+| `ctx.sapiom.agents.run` | Dispatching a DEPLOYED agent by slug — composing systems from small deployed agents                  | Anything that isn't itself a deployed agent                           |
+
+**⚠️ The mistake to never repeat:** sending single-shot, fixed-shape intent through a
+multi-turn surface (`models.run`, or the deferred `llm.submit`/`redeem` lane) instead of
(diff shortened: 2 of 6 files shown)


gwitwer:
Consolidated into #685 — content unchanged plus the terminology scrub (already applied here; carried through the merge). Closing this in favor of that.

PR #684, opened 2026-08-23
feat(agent-core): inspectStep for per-attempt full-fidelity step I/O (SAP-2778)

gwitwer:
## Summary

Exposes per-step-attempt full-fidelity I/O to programmatic consumers — the data path a platform step-inspection feature needs, and the SDK counterpart to a corresponding backend MCP tool (companion change, server-side).

Refs: internal tracker

## What changed

- New `inspectStep(opts, client)` hitting `GET /executions/:id/steps/:stepId/io`, mirroring `inspect()`/`inspectBuild()`'s existing `(opts, client) -> Promise<T>` calling convention rather than introducing a new one.
- New exported `StepIoDetail` (mirrors the engine's server-side DTO shape) and `decodeStepIoDetail`, following `decodeExecutionProjection`'s tolerant degradation posture — never throws on a malformed/partial body, missing fields become `null`/empty defaults, never fabricated. `error` decodes through the existing `decodeStepError` so a caller gets the same structured `StepError` shape `inspect()` already returns for the identical underlying wire form.
- `StepProjection` gains an optional `id` (the step-attempt row id). The wire already sent this on every execution-detail read, but `decodeStep()` silently dropped it — so there was no way to get from an `inspect()` read to the `stepExecutionId` `inspectStep` needs. Fixed as part of this change, since the feature is unusable end-to-end without it. Additive: `null` on an older server; existing consumers unaffected.
- Fixture-based parity tests in `inspect.parity.spec.ts`, anchored to a new code-authoritative fixture matching the shape of the real wire response, matching the file's existing "assert against the real wire, not a shape drawn to please the decoder" convention.

## Why this design

Mirrors `inspectBuild`'s pattern (a second, more specific "inspect" verb alongside the general `inspect()`) rather than adding a mode flag to `inspect()` itself — the step-IO read is a genuinely different endpoint with a different cap tier, not a variant of the execution read.

## Testing

- `packages/agent-core`: `npx jest` → **18 suites, 176 passed** (all pre-existing tests untouched + 5 new: endpoint path construction, per-attempt field fidelity, structured-error decode parity, and honest degradation on an empty/garbage body).
- `npm run typecheck` clean, `npm run lint` clean, `pnpm --filter "@sapiom/agent-core..." build` clean (also confirms the new `StepProjection.id` compiles through every dependent).

Companion change: a server-side companion PR (internal), not linked here.

### DIFF
--- a/.changeset/inspect-step-io.md
+++ b/.changeset/inspect-step-io.md
@@ -1,0 +1,8 @@
+---
+"@sapiom/agent-core": minor
+---
+
+Adds `inspectStep(opts, client)` — fetches one step attempt's full-fidelity `input`/`output`/`error`/`logs` (`GET /executions/:id/steps/:stepId/io`), at a higher size cap than `inspect()`'s own `steps[]` bounds its aggregate read to. Reach for it when a step's fields on the execution projection look truncated.
+
+- New exported `inspectStep`, `InspectStepOptions`, `StepIoDetail`, and `decodeStepIoDetail` (the tolerant decoder, mirroring `decodeExecutionProjection`'s degradation posture).
+- `StepProjection` gains an optional `id` field (the step-attempt row id) — previously decoded from the wire but silently dropped; needed to call `inspectStep` from an `inspect()` read. `null` on a read from a server that doesn't report it — existing consumers are unaffected.
--- a/packages/agent-core/src/__tests__/step-io-detail.wire.fixture.json
+++ b/packages/agent-core/src/__tests__/step-io-detail.wire.fixture.json
@@ -1,0 +1,43 @@
+{
+  "executionId": "exec_0001",
+  "id": "step_0002",
+  "stepName": "render",
+  "stepOrder": 1,
+  "attempt": 2,
+  "status": "failed",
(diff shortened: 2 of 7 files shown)


gwitwer:
Fixed a cross-package type break Test CI caught: `StepProjection.id` was documented as additive/null-on-older-servers but was typed as a *required* `string | null` field, so any hand-constructed `StepProjection` literal built before this field existed — including this repo's own `packages/mcp` test fixture — failed to typecheck (`TS2741: Property 'id' is missing`).

Widened it to `id?: string | null` to match the stated design, and made `inspectStep`'s own JSDoc explicit about the absent/null case: fall back to Run Inspector or the projection's own `input`/`output`/`logs` instead of calling `inspectStep` when `id` isn't there.

Re-verified with the exact commands the Test workflow runs (repo-root `pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm test` — each fans out across every package via `pnpm -r --filter='./packages/*'`), not just the touched package in isolation. All clean, including `packages/mcp typecheck` (the one that broke) and the full test suite.

PR #685, opened 2026-08-23
feat(tools,mcp,harness,skills): teach the LLM call-surface rule across all authoring surfaces (SAP-2764/2775/2776)

gwitwer:
## Summary

Consolidates four previously-separate PRs into one. Content unchanged from each; this PR additionally applies a terminology scrub (dropping internal service naming that had leaked into customer-facing copy) and fills in two missing changesets. Closes/supersedes #679, #680, #682, #683 (each closed with a pointer to this PR). #672 and #684 stay separate (a pre-existing fix and a feature with its own review surface, respectively).

Refs: internal tracker

## Fix round (round 3)

Addresses every review-round finding still open:

- **Structured-output teaching was self-contradictory.** Forced tool-use output has no `text` block — the result lives in `tool_use`. Fixed everywhere the wrong instruction appeared: the MCP fallback instructions, the harness system prompt, and the skill's worked example (now built on `output`/`structuredOf` instead of a hand-rolled `tools`/`tool_choice` block, since that's what ships in this same release).
- **`structuredOf`/`textOf`/`readDisclosure` are now reachable from `ctx.sapiom.llm`** — added to `SapiomClient["llm"]`'s type and runtime implementation (and the `run_local` stub, which is separately typed against the same interface), additive only. New tests cover all three through the client, not just the bare module export.
- **Removed the skill's "is landing" blockquote** — the convenience it described ships in this same PR; the worked example now demonstrates it directly.
- **Fixed "`llm.submit`/`redeem` is a multi-turn surface"** — it's a single call with deferred capacity, not multiple turns; the mistake-to-avoid warning now names only `models.run`'s actual multi-turn loop.
- **`LlmSessionCreateSpec.model` reverted to `string`** — it pins an exact alias, mutually exclusive with the real label field (`label`); only true label fields keep the `RoutingLabel` soft union.
- **Two more sentences trimmed per a documentation-disclosure ruling:** the skill's "agent" row no longer references non-public internal vocabulary by name; the "label" row no longer describes an unshipped billing dimension.

All four `SKILL.md` copies stay byte-identical (`skill-sync` guard). Full re-run of every touched package's suite plus the terminology and provider-neutral-copy guards — see the updated test evidence below.

## Fix round (round 4)

Three more review findings addressed:

[...]

PR #688, opened 2026-08-23
fix(mcp): scope served-class disclosure claim, drop no-op pin advice

gwitwer:
## Summary

Sync fix for `packages/mcp/src/instructions.ts`'s "Calling LLMs and running agent loops" section — restores byte-identity with the corresponding fix landing in a companion Sapiom-repo PR (backend `DEFAULT_MCP_INSTRUCTIONS` + its content-release migration snapshot). #685 (merged) only picked up an earlier round's structured-output fix for this section; the findings below surfaced afterward and need the identical wording here too.

## What changed

- **Scoped the served-class/lane disclosure claim.** "reports the served class and lane on the result" was stated flatly across `llm.run`, `models.run`, and `models.coding.run` alike — but `models.coding.run` returns both as honest `null`s today (the in-sandbox coding agent never sees the gateway's response headers), and older servers omit the fields entirely (`packages/tools` types say treat missing as unknown). The line now says both.
- **Dropped "If you must pin, use the `smart` label."** `smart` already is the platform default, so the clause was a no-op; on the sessions surface, `model` pins an exact alias rather than a routing label, so the advice was also wrong-field there.
- **Structured-output line names the one payload location.** The `output` param is sugar that forces a tool call, so the payload is always the `tool_use` block's `input` — the line now says so, instead of listing the param and a forced tool call as if they were two mechanisms that might put the payload in different places (they are one mechanism; `structuredOf` reads `tool_use.input` either way).
- **Drift-guard assertions** in `instructions.test.ts` pin all three: the scoped disclosure wording, the forced-tool structured-output wording, and the absence of the removed pin advice. The changeset covers all three changes.

## Checksum

The shared section is byte-identical with the backend constant/migration copy at head — SHA-256 `3f1cf1f0c92ba9f27c2f5cf0094c422152cdd16d9c98042c43a9b44f7dc39588`.

**Merge together** with the companion backend PR — merging one without the other reintroduces cross-repo drift.

## Test evidence

`cd packages/mcp && npx vitest run src/instructions.test.ts` → 5 passed (including the new drift-guard assertions). `node scripts/provider-neutral-copy-check.mjs` → passed (74 files audited). `npx eslint src/instructions.ts src/instructions.test.ts` clean. (`npx tsc --noEmit` shows pre-existing unrelated errors from unbuilt workspace deps in this fresh worktree — none in `instructions.ts`.)

Added a `patch` changeset for `@sapiom/mcp`.


gwitwer:
All three addressed in `db2b03db`:

1. **Disclosure scope** — the line now reads "(absent on older servers — treat missing as unknown)" alongside the coding-run `null` scope, matching the `packages/tools` type docs. Changed in lockstep with the backend copy (companion PR pushed the identical wording); shared-section SHA-256 across both repos is now `3f1cf1f0c92ba9f27c2f5cf0094c422152cdd16d9c98042c43a9b44f7dc39588`.
2. **Assertions** — `instructions.test.ts` now pins the scoped disclosure ("treat missing as unknown", "reports both as `null` today"), the forced-tool structured-output wording ("it forces a tool"), and the absence of the removed pin advice (`not.toContain("If you must pin")`).
3. **Body/checksum/changeset resynced** — the PR body now describes head (all three wording changes + the new assertions), carries the recomputed checksum, and the changeset covers the structured-output change too.

PR #694, opened 2026-08-24
docs(skills): teach composition for multi-stage systems + drop the pin-smart no-op

gwitwer:
## Why

Two behavioral-eval findings from the six-scenario authoring gate (run against the released stack; details in the companion internal tracker):

1. **Asked to build a 5-stage pipeline, the authored result was one `defineAgent` with five steps** — `agents.run` composition never appeared, even as an option. Traced in-transcript: the scaffold's "Keep exactly one `defineAgent(...)` export" rule reads as a *design* instruction in context, and the skill's one table row about composition is no counterweight when concretely building.
2. **The skill still taught the pin no-op.** Its label rule said "pass `\"smart\"` if you must pin" (and the worked example's comment offered the same) — advice already removed from the MCP-instructions copies in an earlier correction round, surviving here.

## What changed

- **New skill section: "Composing Deployed Agents (System Design)"** — one agent per PROJECT, but a system is several projects; wrong/right worked example (the five-stage pipeline case) with a thin coordinator dispatching children by slug via `ctx.sapiom.agents.run` (typed against `AgentRunSpec`/`AgentRunResult`), `agents.launch` + `pauseUntilSignal` for long-running children, and bottom-up deploy order. Applied to all **4** skill copies (canonical + 2 templates + plugin) — `skill-sync` enforces identity.
- **Scaffold `AGENTS.md` (all 3 templates)**: the one-export rule now continues "— one agent per PROJECT. A multi-stage system is several small projects composed with `ctx.sapiom.agents.run`", so the rule can no longer be read as "inline every stage".
- **Label rule**: "pass `\"smart\"` if you must pin" → passing `"smart"` is a no-op (it already is the default); same fix in the worked example's comment. Matches the MCP-instructions copies.

## Test evidence

`packages/agent-core`: `npx jest skill-sync` → 7/7 (canonical ↔ both template copies ↔ plugin copy identity). `node scripts/provider-neutral-copy-check.mjs` → passed (74 files). Changeset: patch for `@sapiom/agent-core` + `@sapiom/cli`.

After release, eval scenario s6 (and s2, together with the docs guide fix) gets rerun with the same rig; the gate closes on 6/6.

https://claude.ai/code/session_01W6yq48r2bGVo3GbDXNV4Np

### DIFF
--- a/.changeset/skill-compose-and-drop-pin.md
+++ b/.changeset/skill-compose-and-drop-pin.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/agent-core": patch
+"@sapiom/cli": patch
+---
+
+`sapiom-agent-authoring` skill + scaffold `AGENTS.md`: system-design teaching for multi-stage builds. New "Composing Deployed Agents" section — one agent per PROJECT; a multi-stage system is several small projects composed via `ctx.sapiom.agents.run`, with a worked coordinator example — and the scaffold's "keep exactly one `defineAgent` export" rule now says so inline, so it reads as a per-project rule rather than a design instruction to inline every stage. Also drops the "pass `smart` if you must pin" no-op from the label rule (omitting `model` is the recommendation; `smart` already is the default).
--- a/packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
+++ b/packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
@@ -423,7 +423,8 @@ const research = await ctx.sapiom.agents.run({
 // and does NOT throw — a non-completed child is data the coordinator must
 // branch on, or a failed stage silently feeds `null` downstream.
 if (research.status !== "completed") {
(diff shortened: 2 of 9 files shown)


gwitwer:
All three fixed (plus the nit) in the latest push:

1. **Mojibake** — both agent-core template AGENTS.md now carry a real UTF-8 em dash; verified no `c3 a2` byte sequence anywhere under templates/skills/plugins. The three AGENTS.md copies also share the same sentence now (the wording drift the nit flagged).
2. **Failure branching** — the worked example checks `research.status !== "completed"` before chaining, with a comment stating the contract (resolves on any terminal status, never throws, non-completed child is data to branch on).
3. **Content guards** — skill-sync now asserts on the canonical: no "if you must pin" (case-insensitive), the composition section exists, and the example keeps its status branch; plus a per-template AGENTS.md check (mojibake telltale + composition rule present). 11/11.

PR #696, opened 2026-08-24
[SAP-2793] Enforce ctx.shared writes atomically

ynadge:
## Summary

- make `InMemoryContextStore.set()` validate the complete candidate snapshot before committing it
- accept exactly 262,144 compact-JSON UTF-8 bytes and preserve the previous snapshot after quota or serialization failure
- publish the versioned `CTX_SHARED_SERIALIZATION_FAILED` terminal-error contract and register it with SAP-2792's closed retry-classification path
- pass the real step name from the local host and update SDK, authoring-skill, scaffold, and MCP guidance

## Behavior

- quota failures throw `CtxSharedSizeLimitExceededError` synchronously with phase `ctx_shared_set`
- circular references, BigInt, throwing `toJSON`, and root `toJSON` returning `undefined` throw `CtxSharedSerializationError`
- normal `JSON.stringify` omission and coercion semantics remain unchanged
- both structured failures terminalize on attempt zero when the host supports SAP-2792 atomic settlement

## Verification

- `@sapiom/agent`: 135 tests
- `@sapiom/agent-runtime`: 46 tests
- local runner and skill sync: 31 focused tests
- `@sapiom/mcp`: 103 tests
- affected package typechecks, builds, lint, formatting, and changeset validation

## Release sequencing

The changeset requests minor releases for `@sapiom/agent` and `@sapiom/agent-runtime`, plus patch releases for `@sapiom/agent-core` and `@sapiom/mcp`. The backend PR will follow against `dev` after the published agent version is pinned in both runtime hosts and the backend lockfile is regenerated.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/b813283a-c382-4595-b242-6f62f745da3b)


ynadge:
Addressed the required findings in c33b273d:

- Scoped every published enforcement claim to the SDK's `InMemoryContextStore` and the hosts that construct this store version. The Changeset, README, SDK JSDoc, four synchronized skill copies, both scaffold guides, and MCP instructions now call out the execution-boundary rollout fallback.
- Marked the synchronous failure mode as **Breaking**, documented attempt-0 local terminalization and the no-`delete()` recovery path, and changed `@sapiom/agent-core` from a patch to a minor bump.

I kept whole-snapshot measurement intentionally. Per-key accounting is not equivalent to `JSON.stringify(snapshot)`: stored values are mutable references and can change between `set()` calls without an accounting update, while a root `toJSON` property can change serialization of the whole snapshot. The new mutable-reference regression test demonstrates the first case; the existing root-`toJSON` test covers the second. With the snapshot capped at 256 KiB, the O(snapshot-size) validation is the correctness boundary for this change.

The suggested inbound `runner-core` defense-in-depth check remains outside SAP-2793's setter-gate scope; final completion validation is tracked separately.

Validation: agent 136/136, focused local/skill-sync 31/31, MCP 103/103; affected typechecks, lint, builds, Changesets status, skill identity, and `git diff --check` all pass.

ynadge:
Addressed in 6d622833: the Changeset now uses the same hedged rollout language as the published docs (hosts that have not adopted the store version *may* enforce only at execution boundaries), without asserting current platform behavior. I also removed the stale `Internal:` label from the exported `InMemoryContextStore` JSDoc.

Changesets status, agent typecheck, and `git diff --check` pass after the edit.

PR #697, opened 2026-08-24
fix(examples): Content Pack honors renderClip:false + placeholder-free newsletter

gwitwer:
## Why

Two bugs surfaced by a full end-to-end replay of a real user's run (internal tracker):

1. **`renderClip: false` was silently ignored — and billed a clip the caller declined.** The input was documented on the type but **absent from the entry zod schema**, so validation stripped it and `!usedSampleSource` always decided: any run with a real `source` rendered a ~$0.10 video even with `renderClip: false`. (It double-billed the test that found it.)
2. **The delivered newsletter ended "More soon, [Your name]"** — the LLM writes the newsletter and nothing told it not to ship bracketed fill-ins; a real inbox received the placeholder verbatim.

## What changed

- `renderClip` added to the entry schema (+ `RepurposeInput`); new exported `resolveRenderClip(explicit, usedSampleSource)` gives the explicit input precedence — the sample-source heuristic remains the default when the field is omitted, so the zero-setup gallery run still skips the pricey leg.
- The repurpose system prompt is now the exported `buildRepurposeSystem(audience, numQuotes)` and requires the newsletter be send-ready verbatim: neutral sign-off, no bracketed placeholders (explicitly naming `[Your name]` / `[Company]`).

## Test evidence

`node --test` on the template: **22/22** (4 new: explicit-wins ×2 cases, omitted-falls-back ×2 cases, schema-declares-renderClip parse, prompt-rules pin incl. the pre-existing directives). `npx tsc --noEmit` clean. Provider-neutral check green (74 files).

Refs: SAP-2858

https://claude.ai/code/session_01W6yq48r2bGVo3GbDXNV4Np

### DIFF
--- a/examples/content-repurposing-pipeline/AGENTS.md
+++ b/examples/content-repurposing-pipeline/AGENTS.md
@@ -110,7 +110,8 @@ Drive `check` / `run_local` / `link` / `deploy` / `run` via the Sapiom MCP dev t
 ## Cost
 
 A full run bills an LLM call, N quote **images**, one image-to-**video** clip, and one
-email per recipient. Use `dryRun` while iterating on the copy, and start with a small
+email per recipient. Use `dryRun` while iterating on the copy, `renderClip: false` for the cheaper
+middle ground (real graphics, no clip — the clip is the priciest leg), and start with a small
 `numQuotes` for real runs.
 
 ## Determinism
--- a/examples/content-repurposing-pipeline/README.md
+++ b/examples/content-repurposing-pipeline/README.md
@@ -34,7 +34,10 @@ Input: `{ "source": "<your blog post or transcript>", "title": "..." }`. With no
 `source` at all, the run repurposes a built-in sample post and says so in its output.
 Optional: `audience`, `numQuotes` (default 2, max 4), `deliverTo` (one or more
 recipient emails), `schedule` (a cron string), `model` (a video model — a Sapiom
-semantic alias like `"veo3-fast"`), and `dryRun` (copy only).
+semantic alias like `"veo3-fast"`), `renderClip`
+(set `false` to keep the graphics but skip the teaser clip — the priciest leg;
+defaults to `true` with your own `source`, `false` for the built-in sample),
+and `dryRun` (copy only).
 
 ## Delivery: fan-out per recipient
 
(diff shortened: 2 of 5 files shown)


gwitwer:
All three addressed in the latest push:

1. **Enforcement, not just instruction** — new exported `stripPlaceholderLines` runs inside `parsePack` on the newsletter field: lines carrying a bracketed fill-in are stripped in code (markdown links `[text](url)` spared via a lookahead), and a placeholder-only newsletter falls back — the exact reported shape ("More soon,\n[Your name]") can no longer reach an inbox regardless of model behavior. Three behavior tests cover the strip, the markdown-link exemption, and the fall-back-when-empty path; the prompt-wording pin stays as a second layer.
2. **Orphaned JSDoc** — `resolveRenderClip` and `buildRepurposeSystem` moved below `buildGraphicPrompt`; the SAP-2781 doc block sits directly on its function again.
3. **Docs** — `renderClip` added to the README's optional-inputs list and cost note, `template.json`'s notes plus a new "Full pack, no clip (skip the priciest leg)" example input, and the AGENTS.md cost section (the cheaper middle ground between `dryRun` and a full run).

25/25 template tests, `tsc --noEmit` clean, provider-neutral check green (74 files).

gwitwer:
Round 2 made the case against the stripping approach better than I did: a regex that needs exceptions for inline placeholders, reference-style links, footnotes, and checklists — and can silently delete real copy — is the wrong tool for a template. Reverted it (`stripPlaceholderLines` and its tests are gone; `parsePack` is back to its original shape).

The cause of "[Your name]" is that the prompt asked for a newsletter without giving the model any author identity, so it invented a sign-off to fill later. The prompt now states that plainly: no author/company/reader name is known and nothing is filled in later, so no greetings or sign-offs that need one, no bracketed fill-ins anywhere in the pack (covers #2 — LinkedIn and tweets too), end the newsletter on its takeaway line. That removes the reason the placeholder existed rather than policing its shape. Deliberately prompt-level: for a template, that's the proportionate fix; the pinned prompt test keeps it from regressing silently.

`renderClip` fix + docs unchanged. 22/22, tsc clean, provider-neutral green.

PR #700, opened 2026-08-25
fix(harness): recognize Codex 0.143 composer

ynadge:
## Summary

- recognize Codex CLI 0.143's `Use /skills to list available skills` empty-composer frame
- tolerate future placeholder copy changes using the input marker plus model/cwd footer
- reopen the readiness safety latch after a user accepts a new directory's trust prompt
- keep MCP startup authentication warnings non-blocking and preserve Claude Code behavior

## Root cause

The v0.3.3 readiness fallback correctly saw and retained Codex's trust screen, but it only accepted the newer `Ask Codex to do anything` composer copy as proof that onboarding had cleared. Codex 0.143 renders a different empty-composer placeholder, so the session remained `ready: false` after trust was accepted and Agent Studio kept holding the initial prompt.

The failing Mac's live PTY replay confirmed the trust screen was in history, the current frame contained the 0.143 composer copy, and the MCP warnings did not match a blocking-screen signature.

## Validation

- focused Codex adapter/session-manager tests: 144 passed
- harness unit suite excluding the sandbox-only unreadable-directory test: 2,105 passed
- harness performance suite: 4 passed
- harness typecheck: passed
- harness lint: no errors (one pre-existing warning)
- desktop tests: 152 passed
- desktop typecheck and full dependency build: passed

The excluded workspace-watcher test was run separately and reproduces unchanged because this sandbox cannot make the fixture directory unreadable with `chmod`; it is unrelated to this patch.

### DIFF
--- a/.changeset/codex-0143-ready-composer.md
+++ b/.changeset/codex-0143-ready-composer.md
@@ -3,5 +3,5 @@
 ---
 
 Recognize the Codex CLI 0.143 empty-composer screen so accepting a new
-directory's trust prompt releases Agent Studio's held initial prompt. MCP
-startup warnings remain non-blocking and Claude Code readiness is unchanged.
+directory's trust prompt releases Agent Studio's held initial prompt. Also
+recognize future composer copy changes from the input marker and cwd footer.
--- a/.changeset/codex-0143-ready-composer.md
+++ b/.changeset/codex-0143-ready-composer.md
@@ -1,0 +1,7 @@
+---
+"@sapiom/harness": patch
+---
+
+Recognize the Codex CLI 0.143 empty-composer screen so accepting a new
+directory's trust prompt releases Agent Studio's held initial prompt. MCP
+startup warnings remain non-blocking and Claude Code readiness is unchanged.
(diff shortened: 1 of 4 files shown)


ynadge:
Addressed the automated review in 1c662d4:

- Added a copy-independent composer proof using the `›` input marker plus the cwd footer, while keeping recognized blocking modals authoritative.
- Added coverage for an unknown future placeholder, a selection marker without a footer, and a diff frame containing both the trust modal and underlying composer.
- Trimmed the changeset to describe only behavior changed by this patch.
- Replaced the exact 749/750 ms boundary assertion with a generous settle-window crossing.

Focused adapter/session-manager tests now pass 143/143; typecheck passes; lint remains at zero errors (one unrelated pre-existing warning).

ynadge:
Addressed the partial-repaint safety finding in 15318ff:

- Composer readiness is now vetoed by any single known onboarding fragment. Full blocker detection still requires the complete multi-phrase signature, but an already-latched blocker can only clear on a clean composer frame.
- Added the exact partial trust repaint from the review (`Yes, continue` / selected `No, quit` plus the underlying cwd footer) at both adapter and session-manager levels.
- Updated the changeset to call future-copy recognition best-effort.

Focused adapter/session-manager tests pass 144/144; typecheck passes; lint has zero errors.

PR #707, opened 2026-08-26
fix(tools): widen ModelRunOutcome.costUsd to number | null (SAP-2811)

gwitwer:
## Primary change type

- [x] Bug fix
- [ ] Documentation
- [ ] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

`ModelRunOutcome.costUsd` is published as `number`, but the platform already
sends `cost_usd: null` on a bounded set of runs — rows whose stored estimate
came from a source it will not stand behind, where the honest pre-disclosure
figure is unrecoverable and an explicit null beats resurfacing the wrong number
(server-side guard: sapiom/Sapiom#4504, `AgentRunResultView.costUsd` is already
`number | null` there).

So a strict consumer doing `outcome.costUsd.toFixed(2)` hits a runtime `null`
the type says cannot exist. This was found during the SDK ergonomics work
(#682) and deliberately left alone in that PR — widening is exactly the kind of
type change an additive-only pass rules out, so it was flagged in the PR body
for a minor instead. This is that minor.

## Summary and scope

- `ModelRunOutcome.costUsd: number` → `number | null`, with JSDoc stating the
  consumer-facing contract: it's an **estimate**, not the billed amount, and
  `null` means "not reported" rather than "free".
- `ModelWireResult.cost_usd` → `number | null` and optional, matching what the
  server serializes.
- `mapModelResult` lands a wire `null`, a missing key, and a malformed value all
  on `null` — one encoding, mirroring how this file already normalizes
  `warnings`. A fabricated `0` would read to a consumer as "this run was free",
  which is worse than an honest absence. A genuine `0` still passes through.
- `modelRunResultSchema.parse` applies the same coercion when a consumer
  validates a resume payload (added in review). The schema is opt-in boundary
  hardening; it does not claim every unparsed resume payload is normalized.
- Changeset (`@sapiom/tools`, minor) labelled `**Breaking (type-level):**`,
  matching how this package's CHANGELOG already marks compile-breaking changes
  shipped at minor.

**Out of scope:** the deprecation itself. No `@deprecated` tag is added — the
field has no replacement in the SDK surface, and striking it through in every
consumer's editor with nothing to migrate to would be worse than silence.
Retiring it belongs with the release that gives consumers something else to
use. `CodingRunOutcome` is untouched — coding runs never carried a `cost_usd`
key at all.

## Related work

Related issue or discussion: SAP-2811. Follows sapiom-js#682 (which flagged this
mismatch) and sapiom/Sapiom#4504 (the server-side `servedModel` stale-row guard
that produces the null).

## Validation

```text
# command — result
pnpm build                                                  — clean (all packages)
pnpm typecheck                                              — clean, exit 0
pnpm lint                                                   — clean, exit 0 (only pre-existing warnings)
npx jest --maxWorkers=1 src/models/{run-launch,resume-payload,launch}.spec.ts
                                                            — 3 suites, 32 tests passed
pnpm provider-copy:check                                    — passed (74 audited files)
Node 20 rehydrate-session stress loop                        — 10/10 passed; current main's equivalent fix merged
```

### Tests and documentation

Two cases in `run-launch.spec.ts`, one per delivery path (`ONE encoding of 'no
cost estimate'` for the polled path, and the matching resume-payload case next
to its `warnings` sibling) — each covers a wire `null`, a missing key, and a
non-number, and confirms a real estimate (including a genuine `0`) still comes
through. Docs: the JSDoc on `costUsd` is
the user-facing documentation for this change; no README changes needed
(`packages/tools/src/models/README.md` does not enumerate outcome fields).

- Breaking or externally visible changes: **Type-level only, no runtime change
  for callers.** Under `strictNullChecks`, code that does arithmetic on or
  formats `result.costUsd` now fails to compile until it guards
  (`outcome.costUsd ?? 0`, or skip the row). That is the point — the runtime
  `null` was already reaching those call sites. Read/log-only code is
  unaffected. Shipped as a **minor**, per the ticket.
- Changeset: Added — `.changeset/widen-model-run-cost-usd.md` (`@sapiom/tools`,
  minor), labelled `**Breaking (type-level):**`.

## Security

      will follow the
      private reporting.

## AI assistance

Written with Claude Code and OpenAI Codex. Claude Code read the server-side
`toResultView` guard, made the type and mapper edits, and wrote the model tests.
Codex diagnosed the Node 20 failure as an expired retention fixture, confirmed
the equivalent fix had landed on main, merged it, applied follow-up review
wording, and reran the validation above.

## Checklist

      any N/A checks above.

### DIFF
--- a/.changeset/widen-model-run-cost-usd.md
+++ b/.changeset/widen-model-run-cost-usd.md
@@ -13,9 +13,5 @@ Under `strictNullChecks`, code that does arithmetic on or formats
 or skip the row). Read-only and logging code is unaffected, and nothing changes
 at runtime for a run that does report an estimate.
 
-Both delivery paths land the same encoding: a polled result and a step resumed
-via `pauseUntilSignal` both give you `null` — never `undefined`, never a
-fabricated `0` — when no estimate was reported.
-
 `costUsd` is an estimate, not a billed amount; don't reconcile invoices against
 it.
--- a/.changeset/widen-model-run-cost-usd.md
+++ b/.changeset/widen-model-run-cost-usd.md
@@ -1,0 +1,24 @@
+---
+"@sapiom/tools": minor
+---
+
+`ModelRunOutcome.costUsd` is now `number | null` — the type told a lie.
+
+The platform already sends `cost_usd: null` on a bounded set of runs, where the
+stored estimate came from a source it will not stand behind and the honest
+figure is unrecoverable. The published type said `number`, so a strict consumer
+could do `outcome.costUsd.toFixed(2)` on a value that was `null` at runtime.
+The mapper now lands a wire `null`, a missing key, and a malformed value all on
+`null` — one encoding, never a fabricated `0` that would read as "this run was
+free".
+
+**Strictness impact:** under `strictNullChecks`, code that does arithmetic on or
+formats `result.costUsd` will now fail to compile until it guards —
+`outcome.costUsd ?? 0`, or skip the row. That is the point: the runtime `null`
+was already reaching those call sites. Code that only reads or logs the value is
+unaffected.
+
+Worth guarding rather than defaulting: `costUsd` is a size×lane **estimate**,
+not the amount you are billed. Billed cost belongs to the metering rail (plan
+allowance + priced overage), and this field is deprecated as an authoritative
+dollar figure.
--- a/.changeset/widen-model-run-cost-usd.md
+++ b/.changeset/widen-model-run-cost-usd.md
@@ -2,23 +2,20 @@
 "@sapiom/tools": minor
 ---
 
-`ModelRunOutcome.costUsd` is now `number | null` — the type told a lie.
+**Breaking (type-level):** `ModelRunOutcome.costUsd` is now `number | null`.
 
-The platform already sends `cost_usd: null` on a bounded set of runs, where the
-stored estimate came from a source it will not stand behind and the honest
-figure is unrecoverable. The published type said `number`, so a strict consumer
-could do `outcome.costUsd.toFixed(2)` on a value that was `null` at runtime.
-The mapper now lands a wire `null`, a missing key, and a malformed value all on
-`null` — one encoding, never a fabricated `0` that would read as "this run was
-free".
+The platform doesn't report a cost estimate for every run, so `costUsd` could
+already arrive as `null` at runtime — the published type said `number` and left
+no room for it.
 
-**Strictness impact:** under `strictNullChecks`, code that does arithmetic on or
-formats `result.costUsd` will now fail to compile until it guards —
-`outcome.costUsd ?? 0`, or skip the row. That is the point: the runtime `null`
-was already reaching those call sites. Code that only reads or logs the value is
-unaffected.
+Under `strictNullChecks`, code that does arithmetic on or formats
+`result.costUsd` now fails to compile until it guards (`outcome.costUsd ?? 0`,
+or skip the row). Read-only and logging code is unaffected, and nothing changes
+at runtime for a run that does report an estimate.
 
-Worth guarding rather than defaulting: `costUsd` is a size×lane **estimate**,
-not the amount you are billed. Billed cost belongs to the metering rail (plan
-allowance + priced overage), and this field is deprecated as an authoritative
-dollar figure.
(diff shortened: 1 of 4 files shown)


gwitwer:
Thanks — pushed dca7d49c. Took 1, 3 and 4; took the hardening in 2 but the stated bug doesn't reproduce; declining the nit. Detail below.

### 1. Consumer-facing prose — agreed, edited both surfaces

Agreed on the edit. I'd frame the *reason* differently, though: the problem isn't that "metering rail" or "plan allowance + priced overage" is confidential — it's that neither is **actionable for an npm consumer**, and the pricing structure would drift independently of this SDK while sitting permanently in `CHANGELOG.md`. Same outcome, so both are cut.

One correction on the citation: `packages/harness/CHANGELOG.md:778` records removing *cost/pricing UI surfaces from Studio* (wallet card, per-step figures). That's not precedent against documenting a field's semantics — and I've deliberately **kept** "an estimate, NOT the amount you are billed." That sentence changes what a consumer builds: without it, someone reconciles an invoice against `costUsd` and is wrong. Cutting it to protect pricing prose would trade a real user-facing correctness risk for a cosmetic one.

Also worth noting for scope: only the exported interface's JSDoc reaches consumers. `ModelWireResult` isn't exported and mapper body comments don't survive into `.d.ts` — I verified against the built output. Published surface now reads:

```
Rough cost estimate for the run, in USD — an estimate, NOT the amount you
are billed. Don't reconcile invoices or gate spend against it.

`null` when the platform doesn't report an estimate for a run. Guard before
arithmetic (`outcome.costUsd ?? 0`, or skip the row): `null` means "not
reported", never "this run was free".
```

### 2. The resume-path failure doesn't reproduce — but I took the hardening

I checked this against the server rather than reasoning from the SDK, and the repro doesn't hold. Server-side, `toResumePayload` builds the resume payload from **`toResultView(raw)`** — the same projection the polled path uses (`agent-run.service.ts:1155-1165`). `toResultView` sets `costUsd: isStaleDisclosureRow ? null : rawCostUsd` **unconditionally** (`:1290`), so the key is always present with exactly `number | null`. The polled path is that same object run through the serializer's snake_case pass. There is no path today where a resumed step sees `undefined`.

Which means the widening alone already made the resume path honest — the type now states precisely what that path delivers. `warnings` needs normalizing on both paths because it has two encodings the server can genuinely emit (`[]` vs absent); `costUsd` has one.

That said, `parse` exists to defend this boundary, and "a `number | null` field must never yield `undefined`" is worth enforcing regardless of what the server does today. So I added it next to the `warnings` block:

```ts
r.costUsd = typeof r.costUsd === "number" ? r.costUsd : null;
```

Covered in `run-launch.spec.ts` rather than `resume-payload.spec.ts` — the latter tests coding's `codingResultSchema`; every `modelRunResultSchema` test lives in the former, next to its `warnings` sibling. Flagging the framing because "the exact bug class this PR claims to close, left open on the other path" would have gone into the record as a live defect that was shipped; it wasn't one.

### 3. Breaking label — agreed, precedent verified

Confirmed `packages/tools/CHANGELOG.md:114`, `:289`, `:365`, including one under `### Minor Changes`. Now leads `**Breaking (type-level):**` with the `?? 0` migration retained.

### 4. Deprecation claim — agreed, dropped

Right, and dropping it is the correct half of the either/or. Adding `@deprecated` would strike the field through in every consumer's editor while it remains the only cost signal the SDK exposes, with nothing to point at — that's worse than silence. Deprecating it properly belongs with whatever release gives consumers a replacement.

### Nit — declining, with reasoning

`stub/index.ts` returning `costUsd: 0` is honest: a stub run is local and genuinely costs nothing, exactly like the `durationMs: 0` on the line above it. The changeset condemns a fabricated `0` standing in for *an estimate that wasn't reported* — not a real zero for a run that really was free.

Setting it to `null` would make the stub *less* representative of the modal server response (a real number) in exchange for exercising a guard that `strictNullChecks` already forces at compile time. The type is the enforcement mechanism here; the stub's runtime value isn't.

---

Gates re-run after the changes: `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `provider-copy:check` passes (74 files); model specs 3 suites / 32 tests pass.

PR #705, opened 2026-08-26
fix(examples): route one-shot LLM calls through gateway

gwitwer:
## Primary change type

- [x] Bug fix
- [ ] Documentation
- [ ] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

The shipped examples used `ctx.sapiom.models.run` for one-shot LLM work. That surface is the managed multi-turn loop, so these templates could overthink simple calls and contradicted the canonical SAP-2776 gateway guidance.

## Summary and scope

- Migrates all 22 one-shot LLM examples to `ctx.sapiom.llm.run({ request })` and extracts text with `ctx.sapiom.llm.textOf`.
- Updates example SDK dependencies, the news-roundup lockfile/stub, gallery capability declarations, manifests, READMEs, AGENTS files, source comments, and authoring guidance.
- Adds an examples CI guard that keeps the known one-shot templates, copied docs, dependencies, and registry declarations on the gateway surface while explicitly preserving `models.run` for genuine managed multi-turn loops.
- Does not change any published SDK package API or implementation.

## Related work

Related issue or discussion: SAP-2776; follows sapiom/Sapiom#4520, sapiom/sapiom-js#683, and sapiom/docs-internal#133.

## Validation

```text
pnpm examples:check — passed (11 gallery templates and copied assets validated)
pnpm examples:check:test — passed (153 tests)
22 standalone example `pnpm typecheck` runs — passed against @sapiom/agent 0.12 and @sapiom/tools 0.31
pnpm --dir examples/content-repurposing-pipeline test — passed (22 tests)
pnpm --dir examples/news-roundup test — passed (25 tests)
pnpm --dir examples/research-to-microsite test — passed (15 tests)
pnpm --dir examples/scene-to-video test — passed (8 tests)
pnpm --dir examples/the-brain test — passed (2 tests)
node --test --test-name-pattern='detectAnomalies' examples/scheduled-db-insight-report/index.test.mjs — passed (3 tests)
git diff --check — passed
```

### Tests and documentation

Added a static regression suite for the one-shot call surface and updated the existing example doubles for raw gateway responses. Updated every copied/user-facing example surface that named or stubbed the old call.

Root workspace build/lint/test are N/A because no workspace package source changed; the examples live outside the pnpm workspace and are covered by the example validator, standalone typechecks, and affected example suites listed above.

## Compatibility and release impact

[...]

PR #708, opened 2026-08-26
fix(examples): never fabricate a deliverable when a model reply can't be read (SAP-2892)

gwitwer:
Closes SAP-2892.

## The bug

Nineteen templates asked a model for "ONLY minified JSON", then sliced its reply from the first `{` to the last `}` and `JSON.parse`'d it. Any reply containing a stray brace defeats that — a TypeScript import in generated code, the schema quoted back mid-reasoning, a `{COMPANY}` placeholder — and on failure the `catch` **substituted invented content while the run reported `succeeded`**. From the outside a filler run and a real run were indistinguishable: nothing in Sentry, no metric moving.

What shipped that way:

| Template | Fabricated |
| -- | -- |
| `pr-review-bot` | a `verdict` it never formed |
| `autonomous-pr` | a self-review `verdict` + `summary` |
| `dependency-upgrade` | `risk: "medium"` — which **auto-publishes** under the default `medium` bar |
| `wait-for-webhook` | an accept/reject guessed from `payload.status`, reported as the model's |
| `content-repurposing-pipeline` | the whole content pack — `"<title>: a quick thread. 🧵"` plus 240 chars of the customer's own source |
| `proposal-generator` | a priced quote (`"Professional services", qty 1, $0`), rendered to PDF and sent for human approval |
| `newsletter-autopilot` | an issue over a bare link list, emailed to subscribers; and a judge score of `0` that drove a real revision loop |
| `scheduled-compliance-audit` | `needs_review` with an empty check list, signed and archived as an attestation |
| `cold-outreach-engine` | "I've been following what your team is building" — the generic flattery the prompt forbids — emailed to a real prospect |
| `error-triage-digest` | an `untriaged-batch` cluster, **persisted**, so every later failed batch matched it as *recurring* |
| `research-to-microsite` | an empty site, published at a real URL |
| `scene-to-video` | a bible + shot list, then billed for the images and video |
| `the-brain` | a fleet plan that **launched child agents** and escalated to people |
| `meeting-notes-crm` | `"Unknown contact"` with no action items, written to the CRM and emailed as the recap |
| `logged-in-screenshots` | `[]` — a clean bill of health, which is this template's entire output |
| `human-in-the-loop` | empty ranking criteria, then input order presented to the approver as a ranked shortlist |
| `scheduled-db-insight-report` | a computed top-value presented as the analyst model's finding |
| `news-roundup`, `eval-gate` | already failed loudly, but still hand-parsed; converted so nothing teaches the old shape |

## The fix

Every site now uses the blessed structured-output pattern: `output: { name, schema }` on the `llm.run` spec — which appends that tool and forces `tool_choice` onto it — read back with `ctx.sapiom.llm.structuredOf`. **There is nothing left to slice**, and the "reply with ONLY JSON" prompt text is gone everywhere the schema replaced it. #705 adopted `textOf` at 30 sites and `structuredOf` at zero, leaving the templates behind their own documentation; that gap is closed.

A reply this code cannot read is now a failed step. The readers `throw`, and each says what it refused to invent.

**Two deliberate exceptions, both non-fabricating.** `autonomous-pr`'s self-review is advisory and the branch is already pushed, so it reports `review: null` plus a `reviewUnavailable` reason — the same discipline the file's own preview step already used, and a reader can tell "the model didn't answer" from "the model approved it". `cold-outreach-engine` marks a prospect with no opener **undeliverable** rather than failing a whole batch or emailing a canned line; `verify` and `send` both honor that.

**Where defaulting survives, and why.** Only where the field is not the judgment: snapping a clip duration to one the video model accepts, defaulting a transition to `"cut"`, reading a missing cluster severity as `medium`. Defaulting a *verdict* is inventing one. An empty list stays a real answer where the prompt says so ("no issues found is a valid answer") — which is exactly why a **missing** list has to throw instead of reading as an empty one.

## Guarding against drift

The call surface already drifted back once, so:

- **`pnpm examples:check` rejects the slice.** Repo-wide over `examples/` rather than scoped to a template list, and over every `.ts` file — the parse lived in `lib/` helpers (`news-roundup/lib/select.ts`) and sibling modules (`research-to-microsite/critique.ts`) as well as `index.ts`. Four fixture tests cover it, including that JSON arriving *as* JSON (an HTTP body, a stub payload) is left alone.
- **`pnpm examples:test` runs each template's own suite** (new CI step). `examples/` sits outside the pnpm workspace, so these can't ride the root `pnpm test`; the runner installs each template from the published ranges its `package.json` pins, exactly as a customer's project would. **21 suites, ~130 tests**, each asserting that an unreadable reply throws rather than producing a plausible-looking deliverable.
- **`examples/AUTHORING.md`** — the file an author actually reads — gains a "Reading a model reply" section (with the pattern, the throw rule, and the empty-vs-missing distinction) plus a checklist line.

## Two supporting changes

**The local `llm.run` stub now honors `output`** (`@sapiom/tools`, patch changeset). `output` forces a tool call on the real surface, so a stub that always answered with a text block made `structuredOf` read `undefined` under `run_local` — meaning every template that rightly refuses to invent a value would have failed locally for the wrong reason. It now returns a `tool_use` block carrying a minimal placeholder built from the declared schema (required properties only, one element per array, the first `enum` member). A step that branches on the actual value should still override `llm.run` in its stub file.

**One pre-existing red test fixed in passing.** `scheduled-db-insight-report`'s chart test PUT the rendered SVG at `https://upload.invalid`, and the step's degrade-to-chart-less path swallowed the network error, so its `chartUrl` assertion could never pass. It fails on `main` too — `fetch` is stubbed now, because the new CI step would otherwise start red.

## Verification

```
pnpm build                ✓
pnpm typecheck            ✓  (0 errors)
pnpm lint                 ✓  (0 errors)
pnpm examples:check       ✓
pnpm examples:check:test  ✓  159 pass / 0 fail
pnpm examples:test        ✓  21 suites pass / 0 fail
pnpm terminology:check    ✓
pnpm provider-copy:check  ✓
npx tsc --noEmit          ✓  in all 19 touched template dirs
```

The guard was also confirmed to actually catch a regression: reintroducing one `indexOf("{")` fails `examples:check` with the file and line named.

## Token budgets: raised to 8000 at every forced-tool-call site

Verified by deploying a template to prod and running it — the first time these
templates have been exercised on the real gateway rather than under `run_local`.

**The bug.** Thinking tokens count against `max_tokens` and the thinking budget is
unbounded, so a small cap starves the forced tool call it was meant to leave room
for. Deployed `pr-review-bot` with the previous `max_tokens: 600` and a 5.4k-char
findings blob: the model spent the entire budget thinking (`output_tokens` exactly
equal to the cap) and emitted a `tool_use` block whose `input` was `{}`. The reader
this branch added correctly refused it — *"the model returned no usable verdict"* —
three attempts, run failed (run `374011`).

The same run at 8000 completes all five steps on the first attempt with a real
`request_changes` verdict (run `375422`).

So the readers were right and the budgets were wrong. Every `output`-bearing
`llm.run` site now budgets **8000** — about 10x observed usage, since the passing
`assess` call spends 822 output tokens. `max_tokens` is a cap, not a reservation,
so the headroom costs nothing unless the tokens are produced. 22 sites, 19 files;
the smallest was `eval-gate` at 256, with four more at 300.

## Known limits at merge

Neither is a regression from this branch — both are pre-existing conditions this
branch made visible by refusing to fabricate over them.

**`tool_choice` is not enforced by the gateway ([SAP-2918](https://linear.app/sapiom/issue/SAP-2918), draft).** Measured over 36
calls, **~25% return prose with no `tool_use` block at all** — `stop_reason:
end_turn`, output tokens far below the cap. Not truncation, and independent of
`max_tokens`: the rate is flat at 600, 1000, 4000, 8000 and 200000. Caller-supplied
`thinking: { type: "disabled" }` is silently ignored, which is plausibly the cause,
since a named `tool_choice` cannot be forced while extended thinking is on. The
engine's default three attempts per step absorb most of it (~1.6% residual per
step), but any step without retries is exposed.

Related, seen in the passing run: the model's structured values carry stray markup —
a trailing `</summary>`, and `comments` array elements holding several findings
separated by `</item>\n<item>`. Readers accept it (non-empty strings), so it lands
in the deliverable rather than failing. Noted on SAP-2918.

**`database.get` under `run_local` ([SAP-2909](https://linear.app/sapiom/issue/SAP-2909)).** Four templates — `the-brain`,
`scheduled-db-insight-report`, `error-triage-digest`, `meeting-notes-crm` — cannot
complete a local trace, because the stub returns a DSN the template's own `pg`
client dials with `ssl: "require"`. Reproduces identically on `main`. The other 15
complete cleanly.

## Verification of the budget change

```
pnpm examples:check       ✓
pnpm examples:check:test  ✓
pnpm lint                 ✓
npx tsc --noEmit          ✓  all 19 touched template dirs
run_local sweep           ✓  15/19 (the 4 above are SAP-2909)
deployed prod run         ✓  pr-review-bot 375422 completed, real verdict
```

## Relationship to SAP-2768

Independent of it. Because these templates now throw rather than degrade, the engine-side `Degradation` annotation has much less to cover here.

https://claude.ai/code/session_018SbU3aN7Dq8RtW3mJyniMB


gwitwer:
Thanks — five of these were real. Fixed in 0e9adc7d.

**Confidentiality.** You were right, and there were seven sites, not four: `content-repurposing-pipeline/index.ts:305` and `newsletter-autopilot/index.ts` (the `readIssue` docstring) made the same claim, and `content-repurposing-pipeline/index.test.mjs` also carried an internal run id. All seven now describe the mechanism and drop the claim about what happened. Grep for the phrasings is clean.

**`Number()` coercion.** Real bug, and the worst one here — `Number(null)`, `Number("")`, `Number([])` and `Number(false)` are all a finite `0`, so it reintroduced precisely the substituted `0` this PR is about. All four sites now use `typeof x === "number" && Number.isFinite(x)`, with a comment saying why, and tests over `[null, "", [], false]` on each of the three judges plus the null opener index.

**`applyRanking`.** Also right, and it contradicted its own new docstring. It now throws when nothing in the list matched a candidate — absent key, empty list, or ids from nowhere; "a container with nothing usable in it is not a ranking". A partial ranking still appends the dropped candidates as `(unranked)`, which the docstring already justified. And `readParsed` now requires the `criteria` array to be present — the same empty-vs-missing rule, which I'd applied everywhere except there.

**`cold-outreach-engine` blaming the leads.** Right, and reachable: `readOpeners` only throws when *nothing* is usable, so openers with out-of-range indices returned a non-empty map, matched no contact, and reached the `unmet: ["leads"]` terminal. `personalize` now fails when no prospect got an opener, and the terminal's contact projection carries `verifyStatus` so a partial miss reads as a model miss rather than a bad address.

**Both nits taken.** The guard walks `.mjs`/`.cjs`/`.js` too (verified no false positives), and `autonomous-pr`'s `llm.run` moved inside the `try` its comment already promised.

Two I'm not changing, with reasoning:

**The npm-registry dependency in `examples:test` — keeping it.** On (b), never exercising `packages/`: that's the intent, not a gap. These templates are what a customer installs, so the value is testing them against the ranges their `package.json` actually pins; the new stub `output` path is covered by `packages/tools/src/stub/llm-structured-output.test.ts`, which does run against this PR's code. On (a): an unrelated `@sapiom/tools@0.31.x` publish that breaks 21 gallery templates is something I'd rather CI catch loudly than have a customer find — that's the check working. Pinning exact versions in the runner would test a resolution no customer gets, and would drift from the committed `package.json` silently. A registry outage is a real but generic risk this job shares with every other job's `pnpm install --frozen-lockfile`. The job ran 1m13s on this PR, so it's cheap to keep blocking; if it does flake in practice, making it non-blocking is a one-line change and I'd rather make it on evidence.

**`n > 1 && n <= 100 ? n / 100 : n` mapping 1.5 → 0.015 — leaving it.** Pre-existing in all three judges, and now out-of-contract either way: the schemas declare `minimum: 0, maximum: 1`, so `1.5` violates them. Both readings of it are inventions — `0.015` invents a rejection, clamping to `1` invents a pass — so the existing behavior isn't worse than the alternative, and picking one would be me guessing at a number the judge didn't give under a schema that already forbids it. The genuinely wrong case in that expression was `score: null`, which is fixed above. Worth a follow-up to decide whether the 0–100 tolerance should survive the schema at all, but that's a behavior change to three templates and not this ticket.

gwitwer:
Good catch — that was a regression I introduced, and I'd verified the throw but not what the stub feeds it. Fixed in 0b612d76.

**Reproduced first**, to be sure of the mechanism: `createStubClient().llm.run({ output: { schema: RANK_SCHEMA } })` → `{"ranking":[{"id":"(stub) id","score":0,"rationale":"(stub) rationale"}]}`. Matches no candidate, `ranked` empty, guard throws, run dies at `rank`.

**The throw stays; the schema was too loose.** `buildRankSchema(candidates)` now declares `id` as an `enum` of this run's candidate ids — the same bound-at-the-wire discipline as `the-brain`'s play allow-list, and built per call like `buildRepurposeSchema`/`buildPlanSchema` already are. The stub then picks a real id and the graph completes; more to the point, the model can no longer name a candidate outside the pool at all, which is a failure `applyRanking` would otherwise only catch after paying for the call. I preferred this to a committed stub override because it fixes the production path too, and a stub file would have to be kept in sync with the sample candidates by hand.

**Did the pass you asked for over the other new throws** — empirically, running each schema through the stub rather than reasoning about it. All survive: the count-constrained ones get exactly what they require (`content-repurposing-pipeline`'s `minItems == maxItems == numQuotes` → 3 quotes; `research-to-microsite`'s `minItems: 3` → 3 sections; `scene-to-video` → 1 shot, `duration: 5` off the enum); `readAnomaly` reads the unmatched `table` as `null`, which its schema already permits and `followUp` already skips on; `proposal-generator`'s `quantity: 0` (from `exclusiveMinimum`, which the stub doesn't read) is normalized to 1 by `coerceLine`. You were right that `cold-outreach-engine` survives by accident, so `buildOpenersSchema(count)` now bounds `i` with `maximum: count - 1` — which also closes the shape behind last round's mislabeling finding, since an out-of-range index is exactly what left every contact without an opener.

**One thing worth stating plainly, because your premise depends on it.** "The stub builds structured output from the declared schema" is only true after this PR's `@sapiom/tools` patch publishes. The templates install the published range, and 0.31.0's stub ignores `output` entirely — it answers with a text block, so `structuredOf` reads `undefined` and *every* converted template's reader throws under `run_local` today. The changeset here ships that as a patch, which `^0.31.0` picks up on merge. So the `run_local` story needs both: the stub patch (all 19 templates) and the `id` enum (`human-in-the-loop` specifically, still broken after the patch without it). I verified the published behavior directly rather than assuming it.

That's also why the tests are split rather than one integration test: the template pins the schema property it depends on (the enum equals the candidate ids, and a ranking built from the first allowed id is accepted end-to-end through `applyRanking`), and `packages/tools/src/stub/llm-structured-output.test.ts` pins the mechanism — the stub picks an enum member for a *nested array item*, not a placeholder — because that one runs against the workspace build instead of the published package. An integration test in the template would have tested 0.31.0's stub, which is the thing being replaced.

All gates green: build, typecheck (0), lint (0), `examples:check`, 159 guard fixtures, 21 template suites, terminology, provider-copy.

gwitwer:
Worth flagging for whoever reviews this, because the check is misleading: **`claude-review` on `0b612d76` did not review it — it skipped.** The job log says `Skipping Claude review (review_limit_reached_2_of_2)`, and the workflow reports that skip as a green check. So the two bot comments above cover `a7581b51` and `0e9adc7d`; the `run_local` fix that the second review asked for has had no automated review.

Self-review of that commit, since nothing else will:

- **Empty-pool safety.** `buildRankSchema` and `buildOpenersSchema` now derive constraints from a runtime length, so an empty pool would produce `enum: []` / `maxItems: 0` alongside `minItems: 1` — a contradictory schema. Both call sites early-return before that: `rankCandidates` at `if (candidates.length === 0) return []`, and `personalize` at `if (contacts.length === 0) return goto("verify", ...)`. `Math.max(0, count - 1)` on the opener index is belt-and-braces for the same thing. Neither schema can be built from an empty pool.
- **Duplicate ids.** With `id` now an enum, a model could still name the same candidate twice; `seen` dedupes it and the rest are appended as `(unranked)`, unchanged.
- **The `parseRequest` early return** (`if (!request) return { summary: "(empty request)", criteria: [] }`) does not go through `readParsed`, so the new criteria-presence throw doesn't affect it.
- **Verified rather than reasoned:** I ran each of the new schemas through the actual stub to confirm the values the readers receive, and confirmed the published `@sapiom/tools@0.31.0` stub returns a text block for an `output` spec (which is why the patch in this PR is load-bearing for `run_local`, and why the stub-interaction test lives in `packages/tools` rather than in the template).

Gates on this commit: build, typecheck (0 errors), lint (0 errors), `examples:check`, 159 guard fixtures, 21 template suites, terminology, provider-copy — all green, plus `tsc --noEmit` in the two templates it touches.

Happy to split the `run_local` fix into its own PR if you'd rather it get a fresh automated review pass.

gwitwer:
## `e26d6190` — a regression the automated reviews couldn't have caught, and didn't

`claude-review` skipped this commit and the previous one (`Skipping Claude review (review_limit_reached_2_of_2)`, reported as a green check). So the two bot comments cover `a7581b51` and `0e9adc7d`; nothing has reviewed the two commits written in response to them. Flagging that rather than letting six green checks imply otherwise.

Separately, a human review drove all 19 rewritten templates through `runLocalFromDir` — the bar AUTHORING.md itself sets, and something **none of this PR's tests do**: the template suites exercise the pure reader functions and never execute a graph, so green CI said nothing about runnability. That found a regression I introduced.

**`news-roundup` — fixed.** It is the only template with a tracked `.sapiom-dev/stubs.json`, and this PR converted its `select` step to forced-tool structured output without touching it. The override still returned a text block containing a bare JSON array, so `structuredOf` read nothing and the reader threw — 3 attempts, then the run died at a step that passes on `main`. Rewritten as a `tool_use` block named `emit_selection` with the `{ articles: [...] }` shape, keeping the deliberate 3-of-4 selection (dropping the bakery result is what that stub exists to assert). Verified: `search → select → illustrate → publish`, all succeeded.

**The guard gap was the real finding.** `examples:check` reads source files, so it structurally cannot see stub drift — and the acceptance criterion asks for a guard that stops this class coming back. `checkStubStructuredOutput` now pairs the two files: for every step a committed stub overrides `llm.run` on, if that step forces a tool call, the override must carry a `tool_use` block with the matching name. Resolved **per step**, by slicing `index.ts` on `defineStep(` boundaries — `eval-gate` legitimately reads text in `draft` and a tool call in `judge`, so a per-template answer would reject its text stub. My first cut only resolved tool-name consts in `index.ts`, which silently missed `news-roundup`'s `SELECTION_TOOL` in `lib/select.ts`; it now resolves across sibling modules. Verified by reintroducing both drift shapes (text block; `tool_use` with the wrong name).

**One more stub gap, newly reachable.** `research-to-microsite` died at `critique` on `main`; with this PR it clears critique, runs 11 more steps, then failed at `publish` because `@sapiom/tools/stub` never implemented `sandbox.deployPreview` — an unoverridden call resolved to `undefined` and the caller dereferenced it. Added defaults for the handle methods `examples/` actually calls: `deployPreview`, `createPublicUrl`, `uploadFile`, `uploadDir`. That template now completes 13 steps.

**`run_local` sweep is 15/19**, up from 13. The remaining four — `the-brain`, `scheduled-db-insight-report`, `error-triage-digest`, `meeting-notes-crm` — fail identically on `main`: `database.get` hands back a connection string the template's own pg client really dials (`Client network socket disconnected before secure TLS connection was established`). Pre-existing, unrelated to this ticket, and worth its own — along with AUTHORING.md's claim that `database.get` "never fails", which is untrue one layer down.

### Two things a reviewer should weigh

1. **Merge order.** The templates install `^0.31.0`, whose stub ignores `output` entirely. Until this PR's changeset publishes as `0.31.1`, `run_local` fails for 18 of the 19 templates (`news-roundup` is the exception — its committed override short-circuits before the default). Publishing needs a second merge, since `release-pr.yml` opens a version PR. Landing the `@sapiom/tools` patch as its own PR first avoids that window.
2. **The deployed path is still untested.** `run_local` cannot cover real resource handles or pauses with real resumers, and `meeting-notes-crm`'s `type: ["string", "null"]` schema has never been validated against the real API.

PR #711, opened 2026-08-26
fix(harness): teach omit-model in the Studio prompt, and guard the right direction

gwitwer:
## What was wrong

The Studio coding-agent system prompt said:

> Omit `model` (recommended) or pin the `smart` label; raw provider ids are never honored.

The label pin is a **no-op** — it is already the default — so the prompt taught authors to write a line of code that changes nothing. Every other surface that states this rule had already dropped that advice:

| Surface | pin-a-label advice |
|---|---|
| `agent-core` authoring skill | dropped |
| `@sapiom/mcp` instructions (and the served copy) | dropped |
| call-surface guide | dropped |
| **harness system prompt** | **still present** — fixed here |
| **`@sapiom/tools` JSDoc (4 sites, ships in `.d.ts`)** | **still present** — fixed here (review round 1) |

Corrected in review round 1: the original table omitted `@sapiom/tools`, because the inventory grep searched for `pin the \`smart\`` / `if you must pin` and the tools phrasing is `pass \`"smart"\` if you need to pin explicitly`. The highest-read-volume surface was invisible to the search that produced the table — the same class of failure as the drift itself.

The prompt is the highest-authority surface — it is injected into every Studio session, ahead of the skill and the MCP instructions.

## Why it survived several fix rounds

`packages/harness/src/core/inject/system-prompt.test.ts` asserted the wrong direction:

```ts
expect(content).toContain("pin the `smart` label");
```

So correcting the prompt would have failed CI, while `agent-core`'s `skill-sync.test.ts` asserts the opposite for the canonical skill:

```ts
expect(canonical.toLowerCase()).not.toContain("if you must pin");
```

Two guards in one repo, pinning contradictory versions of the same rule. That is the actual defect — the stale prompt text is just its symptom. The test's own comment claimed it was "kept in sync with the MCP instructions", a sync no mechanism enforced.

## How it surfaced

A clean-room harness eval (six scenarios against the released teaching stack, judged per-criterion). The reproducibility scenario asked the subject to pin a raw provider model id. It correctly refused the raw id and correctly explained that pinning does not buy reproducibility — then recommended pinning the label anyway, because the system prompt told it to. That scenario had already failed a previous run for the same advice coming from the guide; the guide was fixed, the prompt was not.

## The change

- **`packages/harness/src/profiles/default.ts`** — say **omit `model` entirely**, and state why a pin buys nothing, rather than presenting it as a second valid option.
- **`packages/harness/src/core/inject/system-prompt.test.ts`** — assert `Omit \`model\` entirely` is present and that pin-a-label phrasings are absent, so the guard now points the same way as `agent-core`'s. A future correction to this rule will no longer be blocked by its own test.
- Changeset: `@sapiom/harness` patch.

No behaviour change beyond prompt copy; no exported API touched.

## Testing

```
npx vitest run src/core/inject/system-prompt.test.ts   →  4 passed
```

`packages/harness` full suite is unchanged against the base commit — 25 files / 22 tests failing both before and after this change (all pre-existing, in `canvas-render` and friends). `pnpm typecheck` in this package reports 52 pre-existing errors from unbuilt workspace deps; none in either file touched here. `pnpm lint` reports one pre-existing unused-import warning in an unrelated test.

## Follow-up worth doing separately

The rule now lives in ~11 authored places across three repos, and only one group of them (the three MCP instruction copies) has a drift guard. A cross-surface content check — one test that asserts the rule's key claims agree across the prompt, the skill, and the instructions — would have caught this in CI instead of in an eval three fix-rounds later. Related: serving the Studio prompt from the backend rather than shipping it npm-upgrade-bound (SAP-2810), which is what let this drift persist unnoticed.

https://claude.ai/code/session_01FB4ZTXuSgMJcUeLQ8LMQLL

### DIFF
--- a/.changeset/harness-prompt-omit-model.md
+++ b/.changeset/harness-prompt-omit-model.md
@@ -2,4 +2,4 @@
 "@sapiom/harness": patch
 ---
 
-Studio's coding-agent system prompt now says to omit `model` outright instead of offering a label pin, matching the authoring skill, the MCP instructions, and the call-surface guide.
+Studio's coding-agent system prompt now says to omit `model`, states that `smart` is a no-op, and keeps size labels as the deliberate choice — matching the authoring skill, the MCP instructions, and the call-surface guide.
--- a/.changeset/harness-prompt-omit-model.md
+++ b/.changeset/harness-prompt-omit-model.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+Studio's coding-agent system prompt now says to omit `model` outright instead of offering a label pin, matching the authoring skill, the MCP instructions, and the call-surface guide.
--- a/.changeset/tools-jsdoc-omit-model.md
+++ b/.changeset/tools-jsdoc-omit-model.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": patch
+---
+
+`model` JSDoc on the llm and models specs no longer suggests pinning `smart` — it is the default, so pinning it does nothing. The hover text now says to omit `model`, and to pass a size label only when picking a class deliberately.
--- a/packages/harness/src/core/inject/system-prompt.test.ts
+++ b/packages/harness/src/core/inject/system-prompt.test.ts
@@ -34,7 +34,14 @@ describe("generateSystemPromptFile", () => {
     expect(content).toContain("ctx.sapiom.llm.run");
     expect(content).toContain("ctx.sapiom.models.run");
     expect(content).toContain("ctx.sapiom.agents.run");
-    expect(content).toContain("pin the `smart` label");
+    // Omit-only is the advice. This assertion used to require the OPPOSITE
+    // ("pin the `smart` label"), which pinned a no-op into the highest-authority
+    // teaching surface and made correcting it fail CI — see the eval finding in
+    // the PR body. Guard the direction the rule actually points, matching
+    // agent-core's skill-sync content guard.
+    expect(content).toContain("Omit `model` entirely");
+    expect(content.toLowerCase()).not.toContain("pin the `smart`");
+    expect(content.toLowerCase()).not.toContain("if you must pin");
     // Structured/forced-tool output has no `text` block — the reply lives in the
     // `tool_use` block's input. Reading only `type === 'text'` there returns
     // `undefined` and invites exactly the string-parsing fallback this rule bans.
--- a/packages/harness/src/core/inject/system-prompt.test.ts
+++ b/packages/harness/src/core/inject/system-prompt.test.ts
@@ -34,14 +34,23 @@ describe("generateSystemPromptFile", () => {
     expect(content).toContain("ctx.sapiom.llm.run");
     expect(content).toContain("ctx.sapiom.models.run");
     expect(content).toContain("ctx.sapiom.agents.run");
-    // Omit-only is the advice. This assertion used to require the OPPOSITE
-    // ("pin the `smart` label"), which pinned a no-op into the highest-authority
-    // teaching surface and made correcting it fail CI — see the eval finding in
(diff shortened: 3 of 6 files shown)


gwitwer:
All three findings accepted and fixed in `514e56b3`. Finding 1 was the serious one — thank you, that would have shipped a new wrong rule in the same sentence slot as the old one.

## 1. Scoped to `smart` — you were right, and my copy was worse than the bug

Verified the label set is `smart | small | medium | large` (`llm/index.ts:15`, and `model?: RoutingLabel` documents all four), and that a label resolves to a billing size reported back as `servedClass`. So "pinning a label buys nothing" told authors deliberate class selection is pointless, which is false for three of the four labels.

New copy:

> **Omit `model` entirely** — the platform routes it, and `smart` is already the default, so naming it changes nothing. Reach for `small`/`medium`/`large` only to choose a class deliberately. Raw provider ids are never honored.

I went with "is already the default, so naming it changes nothing" rather than your suggested "is a no-op" for one reason worth flagging: it lets the guard in finding 3 forbid *any* juxtaposition of pin-language with `smart`, which "pinning `smart` is a no-op" would itself trip. Same claim, guardable phrasing.

## 2. `@sapiom/tools` JSDoc — fixed, and the PR body claim was wrong

You're right that the completeness table was inaccurate, and the reason is worth recording: my inventory grep searched for `pin the \`smart\``, `if you must pin`, and `or pin the`. The tools phrasing is `pass \`"smart"\` if you need to pin explicitly` — none of those patterns match it. So the surface with the highest read volume was invisible to the search that produced the table. That is the same class of failure as the drift itself.

Fixed all four sites (`llm/index.ts:109`, `:154`; `models/index.ts:69`, `:487`) to:

```
recommended default). `"smart"` IS that default, so pinning it is a no-op;
pass `"small"`/`"medium"`/`"large"` only to pick a billing class deliberately.
```

Added a `@sapiom/tools` patch changeset. `grep -rIn "if you need to pin" packages/*/src` now returns only the two test guards. PR body table corrected.

## 3. Guard hardened — and your suggestion caught a bug in my first attempt

I first wrote `not.toMatch(/\b(?:or|pass|use|set|specify)\s+(?:the\s+)?\`?"?smart/i)` and tested it against the phrasings it needs to reject. It **missed the exact defect this PR fixes** — in "or pin the \`smart\` label" the word after `or` is `pin`, not `smart` — and it missed "set model to smart". A guard that passes the current text while failing to catch the thing it exists to catch is worse than no guard.

Replaced with two shape-based assertions:

```ts
expect(content).toMatch(/`smart` is already the default/i);
// (a) pin-language and `smart` in the same sentence, either order
expect(content).not.toMatch(/\bpin\w*\b[^.]*smart|smart[^.]*\bpin\w*\b/i);
// (b) `smart` offered as a value to supply
expect(content).not.toMatch(/\b(?:pass|use|set|specify|choose|select)\b[^.]{0,30}`?"?smart/i);
```

Checked against six phrasings — all rejected, and the new copy passes all three assertions:

```
CAUGHT | the defect this PR fixes ("or pin the `smart` label")
CAUGHT | old tools JSDoc ("pass `smart` if you need to pin explicitly")
CAUGHT | "use `smart` to pin the tier"
CAUGHT | "set model to smart if you want determinism"
CAUGHT | "or specify smart explicitly"
CAUGHT | "choose `smart` when you want the reasoning tier"
```

## Testing

- `system-prompt.test.ts` → 4 passed.
- `packages/harness` full suite unchanged vs base: 25 files / 22 tests failing before and after (pre-existing).
- `packages/tools`: 31 suites failing and 5 typecheck errors both before and after (pre-existing, unbuilt workspace deps — `@sapiom/analytics-core` unresolved). No errors in either file touched.
- Tools change is comment-only; no runtime or type surface altered.

PR #719, opened 2026-08-27
fix(run-local): give steps a ctx.isLocalTrace signal, stop the stub DSN pointing at a real host, and fix the four templates (SAP-2909)

gwitwer:
## Summary

Four gallery templates (`the-brain`, `scheduled-db-insight-report`, `error-triage-digest`, `meeting-notes-crm`) cannot complete a `run_local` trace. Each holds its own `pg` client and dials what `database.get` returns, and the stub returned a `localhost:5432` DSN — so on any machine with Postgres running, the trace **opened a socket to the author's own database server**, getting as far as authentication and stopping only because `stub_user` did not happen to exist there:

```
ssl:"require" => ECONNRESET | Client network socket disconnected before secure TLS connection was established
ssl:false     => 28000     | role "stub_user" does not exist
```

That environment dependence is why the symptom was an opaque TLS error locally and would be `ECONNREFUSED` on CI.

**The DSN is the smaller of two defects.** The real one: `run_local` gave step code no way to know it was a local trace. Templates therefore hand-roll a `dryRun` *input* flag whose default is `false`, so a zero-input `run_local` takes the live branch by construction. Every template holding a raw socket has this hole — the four that touch a database just fail loudly.

## Changes

- **`ctx.local`** on `AgentExecutionContext`, set by `LocalStubDispatcher`. Deliberately optional and **absent on a deployed run**, so `input.dryRun ?? ctx.local` reads as live in production without the production step runner setting anything — this ships with no coordinated engine-side change. The explicit input flag stays the override; `{ "dryRun": false }` still forces the live path.
- **Stub DSN host → the RFC 6761 `.invalid` TLD**, which is guaranteed never to resolve. The dial fails at DNS in ~20ms, **before any socket is opened**, with the stub named in the error. `database.get` itself still succeeds — only dialing what it returns fails, which is the honest shape.
- **`AUTHORING.md`** claimed the stubbed `database.get` never fails. The *call* doesn't; the *dial* does. Corrected, with the `ctx.local` gate documented alongside it.
- Regression tests for both, plus a changeset.

### Considered and rejected

- *Stub at the `sql` / query layer* — templates call the third-party `postgres` package directly, so this would mean intercepting the module in the bundler.
- *A DSN backed by something that accepts and no-ops* — a fake Postgres wire-protocol server answering DDL and queries with plausible rows; far more machinery than the problem warrants.

## Not in this PR

Templates resolve `@sapiom/agent` from npm, so they cannot reference `ctx.local` until this ships (verified: `error TS2339: Property 'local' does not exist on type 'Ctx'`). **A follow-up PR** bumps the four templates and applies the one-line change, at which point all four complete a zero-input `run_local` with no network — confirmed against a local build:

```
the-brain                    completed  scan → assess → actuate → report
scheduled-db-insight-report  completed  snapshot → detectAnomalies → narrate → followUp → chart → deliver
error-triage-digest          completed  collect → triage → dedupe → digest
meeting-notes-crm            completed  intake → extract → upsert → summary
```

## Test plan

- `packages/tools` **609/609**, `packages/agent-core` **186/186**, full `pnpm test` exits 0.
- `pnpm typecheck` clean; `pnpm lint` shows only pre-existing warnings in untouched packages.
- `pnpm examples:check` and `pnpm provider-copy:check` pass.
- End-to-end on this branch, zero input: `getaddrinfo ENOTFOUND sapiom-run-local-stub.invalid`, DSN `…@sapiom-run-local-stub.invalid:5432/…` — no socket opened, stub named. The template still fails here, which is the expected pre-follow-up state.

Closes SAP-2909 (PR 1 of 2).


gwitwer:
Worked all four findings. Verified each against the repo before acting — one holds more strongly than stated, one is narrower, and the nit doesn't reproduce.

**1. The authoring skill never learned the field — confirmed, and worse than reported.** It ships in **four** byte-identical copies, not three: the canonical, both scaffold templates, and `plugins/sapiom/`. `skill-sync.test.ts` caught the plugin copy after I'd synced the other three. Added the `ctx` reference row and a gate section under "Testing with `run_local`" to all four; md5s match.

**2. Renamed `ctx.local` → `ctx.isLocalTrace`.** Agreed the absence being load-bearing under a generic name is the risk, and the failure mode is silent and in the wrong direction. Worth noting this repo *already* uses a bare `local: true` for an unrelated analytics property (`run-local.ts:121`) — the collision in miniature. The changeset now also states plainly that the deployed runner sets nothing, rather than leaving it implied.

**3. Softened the `.invalid` claims — narrower than stated, though.** The JSDoc doesn't reach consumers: declaration emit strips all comments (no field in the published `context.d.ts` carries one). So the only *published* overclaim was the changeset → `CHANGELOG.md`. Fixed there and in the source anyway: "does not resolve on any conforming resolver", framed as a backstop rather than the guard, timing figure dropped.

**4. Tests now pin the contract.** Rewrote the block to read the step's own `input` argument instead of `ctx.input`, so it exercises the documented expression, and added `{ dryRun: false }` forcing the live path plus `{ dryRun: true }`. Four cases, all passing.

The related doc gap was the best catch here — `input.dryRun ?? ctx.isLocalTrace` only reads correctly in the **entry** step, since `input` downstream is the previous step's output. Both `AUTHORING.md` and `SKILL.md` now show resolving it once at entry and carrying it in `ctx.shared`, which is what the four target templates already do.

**Not acted on — the nit doesn't reproduce.** `@sapiom/agent` ships `files: ["dist", "README.md", "LICENSE", "CHANGELOG.md"]`, and the codename appears nowhere in `dist/`. Confirmed with `npm pack --dry-run`: no source files in the tarball, and declaration emit drops the comment. Nothing leaks.

Green after the changes: `tools` 609/609, `agent-core` 188/188, full `pnpm test` exit 0, `pnpm typecheck` clean, `examples:check` / `terminology:check` / `provider-copy:check` all pass.

gwitwer:
**Folded the follow-up PR into this one — the split was self-inflicted.**

I'd claimed a second PR was unavoidable because templates resolve `@sapiom/agent` from npm. That was true only because I marked the changeset `minor`, which moved the version to 0.13.0 — outside the `^0.12.0` the templates pin — forcing a dep bump that couldn't be written before the version existed.

The addition is an optional field: purely additive, nothing breaks. And this repo reserves `minor` for breaking type changes — `ModelRunOutcome.costUsd` widening to `number | null` shipped as `minor` and said "**Breaking (type-level)**" in the changeset. Corrected to `patch`, so 0.12.2 lands inside the range the templates already pin. No dep bump, no second PR.

Checked the one thing that could still have forced ordering: none of the four commit a lockfile, so all resolve the range fresh. (`examples-test.mjs` uses `npm ci` only where a lockfile exists.)

**Zero-input `run_local`, no network:**

```
the-brain                    completed  scan → assess → actuate → report
scheduled-db-insight-report  completed  snapshot → … → deliver
error-triage-digest          completed  collect → triage → dedupe → digest
meeting-notes-crm            completed  intake → extract → upsert → summary
```

`{ "dryRun": false }` still forces the live path in all four. All four template suites pass under `examples-test.mjs`, the same entry point CI uses.

**One extra fix while in there:** `error-triage-digest` gated a raw HTTP pull on a *second, separate* read of `input.dryRun`. That's exactly the I/O `run_local` cannot stub, so a zero-input local trace would have fetched the batch for real. It now reads the resolved gate.

Each template resolves the gate once in its **entry** step and carries it in `ctx.shared`, per the downstream-`input` point raised in review.

**Two things a reviewer should weigh:**

1. Until the bot publishes 0.12.2, the four templates don't pass their own `npm run typecheck` — they reference a field the published SDK lacks. CI doesn't typecheck templates (`examples-test.mjs` runs install + `test`), so this is invisible to CI and closes on publish. #708 shipped the analogous thing: 19 templates changed alongside a `@sapiom/tools` stub fix they depended on.
2. No lockfiles are committed here, deliberately — one would pin 0.12.1 and silently defeat the fix under `npm ci`.

PR #733, opened 2026-08-28
fix(harness): an agent is never a project — rail model, create affordance, and green rows

gwitwer:
Fixes the issues from the live review of SAP-2926 on a real 75-agent install.

## Why the original shipped broken

Every mock fixture is **authored**, so every agent sits where an agent should sit
and every project row is a project. A lived-in install does not fail that way.
429 green e2e specs, five user-visible defects within a minute of real use, twice.

To fix that I captured a real `~/.sapiom/harness` into the design prototype as a
first-class fixture (`design-eng/agent-studio-v2`, `?org=dogfood`). It paid for
itself immediately, and then the *clean* fixture caught a bug in the first fix.

## A — rail rows painted full brand green

`--accent` is `var(--brand)`, not a wash. `.rail-add-row:hover` → `--bg-hover`;
`.rail-reset-row.is-armed` → a red wash, matching its own "danger tone" comment
instead of contradicting it with the success colour. Guarded by
`css-brand-surface.test.ts`, which allowlists the ~20 legitimate brand
backgrounds and fails on a new one. Mutation-verified.

## B / C — an agent is never a project

    A PROJECT IS A DIRECTORY YOU CHOSE THAT HOLDS AGENTS.

1. **An agent's own directory is not a project.** Dropped when another project
   already shows that agent; otherwise promoted to its nearest non-agent
   ancestor. Not a new rule: `agentNeedsOwnProject` has enforced it on every
   *new* registration since the accumulation was diagnosed. It was never applied
   to entries already in the list, so the guard stopped the bleeding and left
   the wound.
2. **A session cwd is not a choice.** It earns a row only if it holds an agent
   no other project already shows, **or a session is live in it** — "a session
   ran here once" and "something is running here now" are different claims, and
   collapsing them cost a bare scaffold session its row.

A promotion may not create a project that **contains another project**: without
that guard the clean demo fixture promoted three agent folders to `/Users/demo`
and produced one project holding every other project. The known limit is stated
in the test rather than papered over.

Nothing is deleted. Both clauses are derivational, `recentDirs` on disk is
untouched, and any folder is one "Add a project" away — so no migration, no
first-run flow, no undo.

**Measured on the real state: 42 project rows → 3, all 89 registered agents
still visible.** C falls out of this: the one empty project left is the one that
was actually chosen, which is the case "Create the first agent here" was for.

## B4 — the one-node map

`workspaceKey` won the row's onClick unconditionally, so a root-agent row
*always* opened a dependency graph with exactly one node and never the agent.
"I have to click that in order to see my agent" was one operator ordering; both
behaviours were already built and only one was reachable. The more specific
identity now wins and the graph moves to its own control on that row.

The e2e helper for this was named `focusRfqAgentThroughProjectGraph` — eleven
specs took the detour. That is how the bug sat behind a green suite.

## B3 — no way to create an agent inside a project

`onScaffoldSession` was wired at exactly one place, inside the **empty**-project
branch, so a project that already held an agent offered no create affordance at
all. Every project row now has a hover `+`, reusing the existing
`.workspace-row-action` recipe.

## E / F — copy

Header is "Projects" on both axes. The CTA is "Create new agent", deliberately
not "Create new project": it opens the composer, which scaffolds an agent, and
adding a project is already the header's folder-plus.

## Deliberately NOT here

**B2 (long nested labels).** The rail derives four testids from the project
label, so eliding it silently renames them and broke six specs. That wants a
stable row identity first; bundling it here would hide it inside a rename.

## Verification

- **2593 unit tests pass.** `system-graph-freshness` fails identically on clean
  `main` on macOS and is green on Linux CI.
- **440 e2e pass.** The one failure, `session-tabs:268`, also fails on clean
  `main` in a full run and passes in isolation; clean `main` fails one *more*.
- Both rule clauses **mutation-tested**: each fails the suite when stubbed.
- typecheck, lint, terminology and examples gates clean.
- Verified against a **real server on a copy of a real state root**, not mock mode.

Refs SAP-2926.


gwitwer:
Round 2 was reviewed at `c2ce69db`; all six findings and both nits are fixed in
`61611cfb`, pushed after it. The automatic reviewer capped out
(`review_limit_reached_2_of_2`), so nothing has looked at the fixes yet.

| # | Finding | Fix in `61611cfb` |
|---|---|---|
| 1 | No changeset | `.changeset/rail-agent-is-never-a-project.md`, `patch`. Leads with "rows you had may disappear on upgrade, and no files are touched", names both kinds of row, and states the reversibility. |
| 2 | Scanner gaps | Collects **every** member of a selector group, and matches `background-color:` and a shorthand with trailing tokens. |
| 3 | Round-1 nits | `prettier --write` on the three specs; the quadruplicated rationale is one copy plus three pointers. |
| 4 | Contradictory rationale | Deleted. |
| 5 | Live clause untyped/untested | `status` is `SessionStatus`, read as `!= null && !== "exited"` — the same reading `bareSessionAt` uses. Two unit cases added. |
| 6 | Silent no-op on Open | `openProject` now opens the folder that **holds** the agent. Pinned in `open-project.spec.ts`. |
| nit | Helper history backwards | My blanket rename had eaten the name inside the docstring. Restored. |

On #2 — you were right that this was the finding that mattered, and the diagnosis
was exact. Widening the scanner surfaced **three** full-brand backgrounds it had
never seen: `.run-workspace-pulse` as you predicted, plus
`.canvas-overview-resize` in three of its four states. The fourth,
`:focus-visible::before`, was already allowlisted, which is precisely why the
other three were invisible. All are legitimate indicators (a 7px status dot, a
2px resize grabber) and are now listed with that reason.

Mutation-verified against all three walk-arounds, including the exact one you
named — inserting a new offender on the line above an allowlisted selector:

```
M1  a new selector above an allowlisted group member  → FAILS (was: passed)
M2  background-color: var(--brand)                    → FAILS (was: passed)
M3  background: var(--brand) no-repeat                → FAILS (was: passed)
    restored                                          → passes
```

On #6 — agreed that it undercut the reversibility claim, which is the part that
mattered. Opening an agent's own folder now opens the folder that holds it, which
is what the rail's own rule says, so "one Add-a-project away" is true for agent
folders too rather than true-except-there.

Local: 2595 unit / 440 e2e pass. The two failures are the documented pre-existing
pair, both reproduced on clean `main` (macOS-only `system-graph-freshness`; and
`session-tabs:268`, which fails on `main` in a full run too and passes in
isolation — `main` fails one more). All 10 CI checks green on `61611cfb`.

/claude-review

gwitwer:
Round 3's finding was correct and the diagnosis was exact, including the detail that `parentOf` answers `/` rather than null. Fixed in `ebade380`.

The root cause was two copies of one decision. `openProject` reimplemented the promotion hop privately with neither guard, and that is worse than it first reads: `rememberProjectDir` takes an explicit choice at its word, so nothing downstream could decline it. Opening a registered agent at `~/my-agent` would have set the project to **home**, un-closed every removed project under it via `reopenProjects`, and scanned the whole tree. My own comment there claimed the opposite of what `parentOf` does, five lines from the file documenting the trap.

`holdingProjectFor` is now the single exported answer, used by both callers. Null means refuse, and refusing is safe: the agent's own folder stays the root. Both refusals are unit-tested **on the helper directly**, not only through `projectRoots`, since the caller that shipped the unguarded copy is exactly what those tests are for.

```
M1  drop the filesystem-root guard  -> 2 tests FAIL
M2  drop the swallow guard          -> 2 tests FAIL
M3  .sel { background-color: ... }  -> FAILS (was: passed)
    restored                        -> 75 pass
```

Both nits taken: the single-line scanner form now matches `background-color` and trailing tokens, and the changeset is **minor** — you are right that "rows you had may disappear on upgrade" should not arrive unasked-for on `~0.9.0`.

2600 unit / 439 e2e pass. Both e2e failures are the documented flake band (`session-tabs:268`, `smoke:3134`), each passing in isolation, and `session-tabs:268` fails on clean `main` in a full run too.

/claude-review

gwitwer:
Round 4 was right, and the pattern is the point: the same hop broke twice and **neither time was it the rule**. Round 3 it had no guards; round 4 it had the guards fed a narrower question than the rail asks. Fixed in `b31c4898`.

`projectToOpen` now owns that argument, so the thing that keeps breaking has a name, one definition, and its own tests. It passes the same union `agentNeedsOwnProject` is given, for the reason already written 100 lines above it — `recentDirs` is capped at 8 and session cwds are not — plus `pendingCwds`. Deliberately **wider** than the rail's derived roots: every extra entry can only make the swallow guard refuse more often, and refusing is the safe direction.

Your nit was the load-bearing one. All five previous cases hand-built `projects`, so none of them could have caught this; the new ones assert the caller's own function.

```
M1  narrow the union back to recentDirs   -> 2 tests FAIL  (round 4's exact bug)
M2  drop pendingCwds only                 -> 1 test  FAIL
    restored                              -> 79 pass
```

2605 unit / 439 e2e pass; the two known failures are macOS-only `system-graph-freshness` and the documented `session-tabs:268` flake, both reproduced on clean `main`.

---

Unrelated to this PR, found while reviewing the result on a real install and worth their own tickets against #724:

- **`SystemGraphCanvas.tsx:472` is a dead end.** It renders a plain `<p>` with `warnings.length` and nothing else, while each `GraphWarning` carries `code`, `message` and `agentKey`. On a real install that reads "15 static projection warnings" with no way to see one. They decompose as 8 `inventory-extraction-failed`, 4 `duplicate-agent-key`, 3 `dynamic-target`.
- **The map degenerates on a real fleet.** That workspace is 77 nodes and **4 edges**, so ~73 nodes are each their own component and `system-graph-layout` stacks them down a single column a mile long. The layered layout is right for a connected system and has no packing for a disconnected one. Same root as SAP-2926's issue D: only `invokes` edges are extracted, so a real install is almost entirely unconnected.

/claude-review

gwitwer:
Round 5 is right, and it caught my own reasoning as the bug. "Wider can only make the guard refuse more often, and refusing is safe" was the sentence I wrote in the previous commit, and it is false: **refusing IS the no-op**. Fixed in `cd1b405e`.

The deeper error was treating it as a size question. The raw union is not merely wider than the rail's list, it is a different *kind* of list: `projectRoots` skips `isAgentDir` in both loops that build `kept`, so the rail's guard only ever measures projects. Mine measured agent directories too — and as you point out, an agent no project contains boots its sessions in its own folder, so `sessionCwds` routinely holds them.

Three cases pin it, chosen as the shapes that actually occur rather than the ones easy to construct: the agent's own session cwd, a sibling agent's folder, and a **real** project under the same parent, which must still refuse.

```
M1  restore the raw union   -> 2 tests FAIL  (round 5's exact bug)
M2  over-filter to []       -> 3 tests FAIL  (the guard must still fire)
    restored                -> 82 pass
```

Dead `isAgentDir` local and its stale comment deleted too.

Worth recording, since three rounds found three bugs in one hop and none in the rule: every failure was in *what gets passed*, and each fix I made was narrower than the invariant. The invariant is that this argument must be the same kind of list `projectRoots` measures — not a superset, not "recentDirs plus more". That is now stated where the argument is built, which is the only place it can be enforced.

2608 unit / 439 e2e pass. Failures are the known macOS-only `system-graph-freshness` and two flake-band members (`session-tabs:268`, `popover-crop:78`), each passing in isolation.

/claude-review

PR #734, opened 2026-08-28
fix(mcp): re-sync the offline AUTHORING_INSTRUCTIONS fallback with the server primer [SAP-2959]

gwitwer:
## What

Re-syncs `packages/mcp/src/instructions.ts` (`AUTHORING_INSTRUCTIONS`) with the canonical server-side primer, moving the offline fallback from 2.6-era text to the current 2.8 release, and adds a frozen sha-256 so the two copies cannot drift silently again.

## Why

`AUTHORING_INSTRUCTIONS` is the copy the MCP server serves when its startup fetch of `GET /v1/mcp/instructions` fails. It is a verbatim duplicate of the server's canonical primer by design — the package must work offline, so it cannot import it — and the only thing enforcing that was a doc comment saying KEEP IT IDENTICAL.

It stopped being true two content releases ago. The fallback was missing both App Links releases:

- **2.7** — the App Links paragraph in "Preview a web app" (a durable `https://apps.sapiom.ai/{org}/{slug}` that outlives the sandbox)
- **2.8** — the shipped `sapiom_dev_app_publish`, version-gated at `@sapiom/mcp >= 0.13`

So the one path with no other source of truth — an offline session, which is precisely the case where the live fetch failed — was told App Links do not exist at all: no durable-sharing paragraph, no local publish tool, no hosted or REST surface. The copy also still carried the pre-2.7 single-alias framing of the hosted capability MCP, which now lives under its own `sapiom-direct` alias.

**The drift ran both ways**, which the ticket's "2.6-era text" framing (and my first description of this PR) obscured. Diffed against `origin/main`, this copy was *behind* on App Links and the alias framing, *level* on the LLM call-surface block, and *ahead* on `ctx.shared` — see Removed below. Blast radius is low either way, since online sessions live-fetch the server text and it wins.

## What the sync removes

The fallback carried an 11-line `ctx.shared` quota contract — the inclusive 256 KiB / 262,144-byte limit, compact-`JSON.stringify` measurement, setter-time validation, no `delete()`, structural guards over `instanceof` — that **is not in the served primer**. Online sessions never saw it; only offline ones did.

Syncing removes it here, and I'm keeping that: the two are one canonical text, the ticket's scope is "copy verbatim," and the digest below cannot hold if the copies differ by a paragraph. It is disclosed rather than silent — there's a test named for the removal, and a Removed section in the changeset. The contract still reaches authors via `packages/agent/README.md` and the scaffold-shipped `sapiom-agent-authoring` skill. Putting it back into the primer is a server-side content release; flagged for the epic rather than reintroducing divergence here.

## Changes

- **`packages/mcp/src/instructions.ts`** — body replaced verbatim with the 2.8 primer; byte-identical to the server copy. Doc comment updated to describe the guard rather than restate a rule that had already failed.
- **`packages/mcp/src/instructions.test.ts`** — new `is byte-identical to the backend primer (frozen sha-256, SAP-2959)` pins `sha256(AUTHORING_INSTRUCTIONS)` to `7f518d9c…4494c0df`, the same digest the server-side spec pins for its current primer. Updates the assertions the old text happened to satisfy (the previous hosted-MCP framing), adds App Links coverage, and adds a test recording the `ctx.shared` removal above.
- **`.changeset/`** — patch release of `@sapiom/mcp`.

## The guard, and why the existing tests missed this

Every `contain` assertion in this file passed against the stale text — they assert what must be present, and nothing noticed what had stopped being copied. So the copy fell two releases behind with a green suite.

The digest is what binds the two copies, but only because **both** ends pin it, which review on the paired server-side PR was right to press on. A pin checked only against its own body is a self-consistency check: it catches an accidental in-place edit here and nothing else. In particular it would not catch the direction that caused this incident — a server-side content release, which touches that repo only.

So the paired PR pins this same digest against the server's *current* primer. Cutting the next release changes that hash and reddens a spec that names this pin, which is the moment its author discovers this body has to move too.

Stated plainly, in the code as well as here: neither pin can block a merge in the other repository, and an author can still move one side alone. What the pair removes is the silent path — drifting now takes a deliberate edit to a line that says what it is for. Re-pointing this digest alone just re-blesses the drift.

This is the fourth reactive heal of the same class of drift (SAP-1367 → SAP-1467 → SAP-2227 → this); the guard is the point of the ticket, not the copy.

## Known-adjacent, deliberately not in this PR

The repo ships three answers on MCP aliases: `docs/mcp-servers.md:7` (Remote `sapiom` / Local `sapiom-dev`), `README.md:141` (`claude mcp add sapiom-dev`), and `packages/mcp/README.md:35` (`claude mcp add sapiom`, agreeing with the primer). Review flagged it here.

It's real, but it isn't this PR's: the alias vocabulary is the canonical primer's, decided server-side in 2.7 and already live to every online session. The contradiction exists on `main` today and this diff doesn't change it — it only moves the offline copy onto the vocabulary the online path already uses. A partial fix would leave `docs/mcp-servers.md` built end-to-end around the inverse mapping, so the repo would still ship three answers; doing it properly is a docs-page rewrite that belongs with SAP-2923. Flagged for the epic.

## Testing

- `vitest run src/instructions.test.ts src/instructions-fetch.test.ts` — 14 passed, including the five regression assertions restored after review (`tool_use`, `"it forces a tool"`, `"treat missing as unknown"`, `` "reports both as `null` today" ``, `not.toContain("If you must pin")`), all verified passing against the new text
- `tsc --noEmit` — clean
- `eslint src --ext .ts` — clean
- `prettier --check` on the touched files — clean
- Independently verified the new body byte-for-byte against the server constant before writing the digest

## Related

- Ticket: SAP-2959
- Epic: SAP-2921 — make App Links discoverable from Studio and docs.sapiom.ai
- Siblings: SAP-2922 (shipped `sapiom_dev_app_publish`, #722), SAP-2923 (the 2.7/2.8 server content releases)
- Precedent: SAP-1486 (#283) did the same sync for the sandbox-preview lifecycle
- A paired server-side PR restores the KEEP THE TWO IDENTICAL rule and adds the matching pin. **This PR should land first** — that one's comments describe the copies as already re-synced.

https://claude.ai/code/session_01M2AUF348cx5Mh4N9bbnkSF

### DIFF
--- a/.changeset/resync-authoring-instructions-fallback.md
+++ b/.changeset/resync-authoring-instructions-fallback.md
@@ -10,6 +10,7 @@ no durable-sharing paragraph, no `sapiom_dev_app_publish`, no hosted or REST pub
 surface. It also still described the older single-alias framing of the hosted
 capability MCP, which now lives under its own `sapiom-direct` alias.
 
-The two copies are now byte-identical, and a frozen sha-256 in `instructions.test.ts`
-— the same digest the server-side spec pins for this content release — fails CI on a
-one-sided edit instead of letting them drift silently a third time.
+The two copies are now byte-identical, and `instructions.test.ts` pins their sha-256.
+The server-side spec pins the same value against its own copy, so the next content
+release reddens a spec that names this pin rather than letting the two drift silently
+a third time.
--- a/.changeset/resync-authoring-instructions-fallback.md
+++ b/.changeset/resync-authoring-instructions-fallback.md
@@ -1,0 +1,15 @@
+---
+"@sapiom/mcp": patch
+---
+
+Re-sync the offline `AUTHORING_INSTRUCTIONS` fallback with the server's canonical
+primer. The fallback is what the server serves when its startup fetch of
+`GET /v1/mcp/instructions` fails, and it had drifted two content releases behind — so
+exactly the sessions with no other source of truth were told App Links do not exist:
+no durable-sharing paragraph, no `sapiom_dev_app_publish`, no hosted or REST publish
+surface. It also still described the older single-alias framing of the hosted
+capability MCP, which now lives under its own `sapiom-direct` alias.
+
+The two copies are now byte-identical, and a frozen sha-256 in `instructions.test.ts`
+— the same digest the server-side spec pins for this content release — fails CI on a
+one-sided edit instead of letting them drift silently a third time.
--- a/.changeset/resync-authoring-instructions-fallback.md
+++ b/.changeset/resync-authoring-instructions-fallback.md
@@ -2,15 +2,23 @@
 "@sapiom/mcp": patch
 ---
 
-Re-sync the offline `AUTHORING_INSTRUCTIONS` fallback with the server's canonical
-primer. The fallback is what the server serves when its startup fetch of
-`GET /v1/mcp/instructions` fails, and it had drifted two content releases behind — so
-exactly the sessions with no other source of truth were told App Links do not exist:
-no durable-sharing paragraph, no `sapiom_dev_app_publish`, no hosted or REST publish
-surface. It also still described the older single-alias framing of the hosted
-capability MCP, which now lives under its own `sapiom-direct` alias.
+Re-sync the offline `AUTHORING_INSTRUCTIONS` fallback with the primer the server
+actually serves. This is the text an agent gets on connect when the server's startup
+fetch of `GET /v1/mcp/instructions` fails; it had fallen behind, so offline sessions
+were told App Links do not exist.
 
-The two copies are now byte-identical, and `instructions.test.ts` pins their sha-256.
-The server-side spec pins the same value against its own copy, so the next content
-release reddens a spec that names this pin rather than letting the two drift silently
-a third time.
+Added, matching what online sessions already receive:
+
+- **App Links** — publishing a durable `https://apps.sapiom.ai/{org}/{slug}` that
+  outlives the sandbox a preview URL dies with, via `sapiom_dev_app_publish` (needs
+  `@sapiom/mcp` >= 0.13), the hosted `sapiom_app_publish`, or the REST route.
+- The current alias framing: local authoring under `sapiom`, hosted one-off capability
+  calls under the distinct `sapiom-direct` alias.
+- Corrected lifecycle wording for `sapiom_dev_agents_check`, `_run_local`, and
+  `sapiom_authenticate`, and the fuller docs links.
+
+Removed, because the served primer does not carry it: the detailed `ctx.shared` quota
+paragraph (256 KiB limit, `JSON.stringify` measurement, setter-time validation, no
+`delete()`). This fallback had it and the served text did not, so most sessions never
+saw it. It remains documented in `@sapiom/agent`'s README and the scaffold-shipped
+`sapiom-agent-authoring` skill.
--- a/packages/mcp/src/instructions.test.ts
+++ b/packages/mcp/src/instructions.test.ts
@@ -110,15 +110,21 @@ describe("server instructions", () => {
   it("is byte-identical to the backend primer (frozen sha-256, SAP-2959)", () => {
     // The `contain` assertions above are what let this copy fall two content
     // releases behind the server without anything going red: each one still
-    // passed against the older text. This is the guard that actually binds the
-    // two copies — the digest is the same frozen value the server-side spec pins
-    // for the matching content release, so a one-sided edit reddens one repo or
-    // the other, with no network call from either test suite.
+    // passed against the older text. This digest is what actually binds the two
+    // copies, and it works from both ends: the server-side spec pins this same
+    // value against ITS current primer, so a content release there reddens that
+    // spec and forces its author onto this pin, while an in-place edit here
+    // reddens this one. Neither suite makes a network call.
+    //
+    // Be honest about the limit: neither pin can block a merge in the other
+    // repository, and an author can still move one side alone. What the pair
+    // removes is the silent path — drifting now takes a deliberate edit to a line
+    // that says what it is for.
     //
     // To change the primer: ship the server-side content release, copy its new
-    // body here verbatim, and update both digests in the same pair of PRs. Never
-    // re-point this digest on its own — that just re-blesses the drift the guard
-    // exists to catch.
+    // body here verbatim, and update both pins to the new digest in the same pair
+    // of PRs. Never re-point this digest on its own — that just re-blesses the
+    // drift the guard exists to catch.
     //
     // Current release: 2.8 (App Links + `sapiom_dev_app_publish`).
     const sha256 = createHash("sha256")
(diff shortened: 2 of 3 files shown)


gwitwer:
Four of five actioned; one declined with reasoning, and one correction to the review. Pushed in 499e2bf (and 8e5b5c4 for #2).

## 3. Deleted regression assertions — you're right, and the cause was worse than sloppiness

Restored all five. The cause is worth stating plainly: I read the base test file from a stale local branch instead of `origin/main`, then wrote the replacement wholesale. `origin/main` is 108 lines; the version I had looked at was 56. I overwrote 52 lines of coverage I never saw. The `toContain` set I "updated" was reconstructed from the wrong file.

Restored and verified passing against the new text: `tool_use`, `"it forces a tool"`, `"treat missing as unknown"`, `` "reports both as `null` today" ``, and `not.toContain("If you must pin")` — with their original comments, which explain why each exists.

Your reasoning for why the digest doesn't cover them is the part I'd gotten wrong: it is re-pointed on every content release by design, so it catches accidental drift but never a deliberate regression made *during* a sync — which is precisely when "If you must pin" would come back. Negative assertions and the digest are orthogonal. They're back.

## 4. The `ctx.shared` quota removal — undisclosed, now disclosed

Correct and it was the most consequential thing in the diff. Confirmed against `origin/main`: the 11-line contract was in this fallback and **not** in the served primer, so online sessions never saw it — this copy was ahead there, not behind, which my "2.6-era text" framing obscured.

I'm keeping the removal. The two are one canonical text, the digest cannot hold if they differ by a paragraph, and the ticket's scope is "copy verbatim." But it now has a test named for it recording what left and why, the changeset has a "Removed" section, and the PR body no longer describes this as purely additive. Restoring it to the primer is a server-side content release — I've flagged it for the epic rather than reintroducing divergence here.

## 5. Changeset and JSDoc publish internal mechanics — correct on both, including the part I'd missed

Verified: `tsconfig.build.json` excludes only `src/**/*.test.ts`, so `instructions.ts`'s JSDoc *is* emitted into `dist/instructions.d.ts` — I confirmed it in the built output. CI mechanics and repo layout were shipping in the tarball.

Both trimmed to what a consumer can act on. The sync rule, the digest, and its limits moved into `instructions.test.ts`, which is excluded from the build. The changeset is now Added/Removed against the served primer, with the quota disclosure.

## 2. Digest direction — fixed before this review landed

You reviewed d20a1be. 8e5b5c4 and the paired PR's f203024 fix exactly this: the server-side spec now pins the same digest against **its** current primer, so cutting a release reddens there and forces its author onto this pin. Both sides now say out loud that neither can block a merge in the other repository — what the pair removes is the silent path, not the possibility.

**On the `.trim()` trap:** I don't think it bites, and I'd rather not "fix" it. `instructions-fetch.ts:25` trims the runtime HTTP body; the digests are taken over the two source constants, which I diffed byte-for-byte before writing the pin. Neither constant has leading or trailing whitespace — they open on `#` and close on `project.` immediately before the backtick — so trimmed and untrimmed hash identically today, and if that ever stopped being true the digest is what would catch it. Hashing trimmed text on both sides would add a normalization step whose only effect is to make one class of real divergence invisible. Happy to be shown wrong if you're seeing a path where the served body and the constant legitimately differ by whitespace.

## 1. Alias contradiction — real, but declining to fix it here

Confirmed all three surfaces, so the repo does ship three answers:

- `docs/mcp-servers.md:7` — Remote `sapiom` / Local `sapiom-dev`
- `README.md:141` — `claude mcp add sapiom-dev -- npx -y @sapiom/mcp`
- `packages/mcp/README.md:35` — `claude mcp add sapiom -- npx -y @sapiom/mcp` (agrees with the primer)

Where I disagree is that this PR ships it. The alias vocabulary is the canonical primer's, decided server-side in 2.7 and **already live to every online session** — which is nearly all of them, since this constant only serves when the fetch fails. The contradiction exists on `main` today and is unchanged by this diff. This PR moves the offline copy onto the same vocabulary the online path already uses; it makes an existing inconsistency easier to notice, not worse.

I also don't think a partial fix helps. Correcting `README.md` alone leaves `docs/mcp-servers.md` built end-to-end around the inverse mapping — a comparison table, "Remote `sapiom`" and "Local `sapiom-dev`" section headings, and prose — so the repo would still ship three answers, just rearranged. Doing it properly means rewriting that page against the canonical vocabulary, which is a docs change with its own review and belongs with SAP-2923 (this epic's docs ticket, "Document App Links on docs.sapiom.ai, in the MCP connect-time instructions"). Folding it in here would also put a docs rewrite inside the PR whose one testable property is that the primer body is byte-identical to the server's.

Flagged for the epic. If a maintainer wants it in this PR I'll do the full page, but I don't think a stale-docs fix should ride along with the sync it's unrelated to.

---

Correction to one thing in the review, since it shaped point #2's framing: this copy was not uniformly "2.6-era." It was behind on App Links and the alias framing, level on the LLM call-surface block, and *ahead* on `ctx.shared`. The drift was bidirectional, which is a slightly stronger argument for the pin than the one-directional story.

gwitwer:
For the record, since the last review on this page reads "Request changes" against findings that were fixed in the commit that merged (`499e2bf`, reviewed at `8e5b5c4f`):

- **#3 deleted regression assertions** — restored. `tool_use`, `"it forces a tool"`, `"treat missing as unknown"`, `` "reports both as `null` today" ``, and `not.toContain("If you must pin")` are all on `main`, with their original comments. Verified passing against the new text.
- **#4 `ctx.shared` quota removal** — disclosed. There's a test named for the removal explaining that the paragraph was in this fallback and not in the served primer, and the changeset has a "Removed" section.
- **#5 / JSDoc** — the changeset is Added/Removed against the served primer with no CI or repo mechanics, and the `instructions.ts` JSDoc no longer names a companion repo or describes enforcement; that detail moved to `instructions.test.ts`, which `tsconfig.build.json` excludes from `dist`.
- **#2 digest direction** — the server-side spec now pins the same digest against its own current primer, so a content release reddens there.

Verified against merged `main`, not the PR head: `AUTHORING_INSTRUCTIONS` diffs byte-identically against the server constant, and the five restored assertions are present.

**Still open, deliberately not in this PR — worth a ticket:** the alias contradiction from #1. `docs/mcp-servers.md` (83 lines, built end-to-end around "Remote `sapiom`" / "Local `sapiom-dev`") and `README.md:141` still teach the inverse of what the primer says, while `packages/mcp/README.md:35` agrees with it. My reasoning for excluding it is in the comment above — pre-existing on `main`, already live to every online session since 2.7, and unchanged by this diff — but it is real and it should be fixed. It's a docs-page rewrite against the canonical vocabulary and belongs with SAP-2923.

I also have a two-line follow-up trim to the published JSDoc (dropping "and what that digest can and cannot catch" from the maintainer pointer) that missed the merge. Not worth its own PR on its own; happy to fold it into whoever takes the docs ticket.

PR #736, opened 2026-08-29
fix(desktop): make the pre-release update channel reachable [SAP-2965]

gwitwer:
## What

The beta update channel has never worked, for any release. `electron-updater`'s `GitHubProvider.getLatestVersion()` runs `semver.valid(tag)` over every `releases.atom` entry when resolving a pre-release channel and **skips the failures**:

```js
const hrefTag = hrefElement[1];        // "harness-desktop-v0.3.8"
if (!semver.valid(hrefTag)) continue;  // → null, so EVERY entry is skipped
...
if (tag == null) throw newError(`No published versions on GitHub`, ...)
```

`harness-desktop-v0.3.8-beta.1` is the proof: published yesterday, correct in every other respect, and installable by nobody. Replaying that loop against the live feed with `currentChannel="beta"` skips all 10 entries and throws.

A leading `v` is the only prefix semver tolerates, so the tag namespace becomes `v*`.

The stable channel is untouched by this — `allowPrerelease: false` resolves via `getLatestTagName()` → `/releases/latest`, which treats the tag as an opaque string. That is why this survived from the first release to 0.3.8 with every check green, and existing installs keep updating across the change.

## Changes

- **`desktop-release.yml`** — trigger `harness-desktop-v*` → `v*`; `prepare` expects `v${version}`; **new** guard failing the build on a version that isn't semver-parseable, so this cannot regress silently.
- **`release-tags.test.ts`** (new) — pins the namespace, the guard, and *that the channel-mirror step still exists*. That step looks redundant beside `generateUpdatesFilesForAllChannels`, but that option is a **no-op for us**: `computeChannelNames()` short-circuits on `publishConfig.provider === "github"`. Deleting the mirror would strand every beta install with no path back to stable.
- **`update-prefs.ts`** — persisted `preRelease` opt-in on the existing store, default off, sanitized like its siblings.
- **`update-policy.ts`** — `resolveUpdateChannel` takes prefs. Precedence **env > toggle > version**. The toggle can only move an install *onto* betas, never drag a `-beta.2` build back to `latest` — that would present a downgrade as an update.
- **`updater.ts`** — re-resolves the channel once prefs load, same async-after-sync shape as the existing auto-install toggle (first check is 30s out).
- **`CLAUDE.md`** — new tag table, the beta-cut procedure, and an explicit "do not add `generateUpdatesFilesForAllChannels`, do not delete the mirror step" with the reason.

## Not in this PR

The UI control for the opt-in. The mechanism and persistence land here; placement is a design call still open. `SAPIOM_UPDATE_CHANNEL=beta` works today for anyone who needs it before then.

## Delivery lane

Standard. No migrations, no infrastructure, no integrations. Affects the desktop release workflow and the desktop updater only; no runtime service, no backend surface.

## Tests

`pnpm --filter @sapiom/harness-desktop typecheck` clean; `pnpm --filter @sapiom/harness-desktop test` → **20 files, 167 tests passing** (12 new: 7 in `release-tags.test.ts`, 5 in `update-policy.test.ts`).

## Production signal

The next tag is the signal, and it is deliberately a beta: cut `v0.3.9-beta.1`, confirm GitHub marks it pre-release, confirm `/releases/latest` still resolves to the newest **final**, then confirm an install with `SAPIOM_UPDATE_CHANNEL=beta` is actually offered it — the exact check that fails today. A stable install must be offered nothing.

Fix-forward: if a `v*` tag misbehaves, `prepare` fails fast (~15s, before any notarization) and no release is published; correct the version or the guard on `main` and re-tag. The stable channel is unaffected either way, so a bad beta cannot reach non-opted-in users.

https://claude.ai/code/session_015sZDBYwf1SJBVxH5wPYyfc

### DIFF
--- a/.changeset/desktop-prerelease-channel.md
+++ b/.changeset/desktop-prerelease-channel.md
@@ -1,0 +1,15 @@
+---
+"@sapiom/harness-desktop": patch
+---
+
+Make the pre-release update channel reachable. Desktop releases are now tagged
+`vX.Y.Z` instead of `harness-desktop-vX.Y.Z`, because the old prefix is not valid
+semver and the updater silently skipped every release when resolving a
+pre-release channel — so an install following betas was told "no published
+versions" no matter what had been published. The stable channel was never
+affected and keeps updating across the change.
+
+Also adds a persisted `preRelease` preference, so an install can follow betas
+without setting `SAPIOM_UPDATE_CHANNEL` — unusable as a real control on macOS,
+where a Finder or Dock launch inherits no shell environment. The control that
+writes it is a follow-up; the preference and the plumbing land here.
--- a/.claude/skills/release-desktop/SKILL.md
+++ b/.claude/skills/release-desktop/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: release-desktop
-description: How to cut a Sapiom Studio desktop release (@sapiom/harness-desktop) in this repo. The release is TRIGGERED BY PUSHING A GIT TAG on main — `harness-desktop-vX.Y.Z` — after versions are bumped via changesets. Use whenever asked to "release", "cut/make a release", "ship a new version", "publish/tag the desktop app", "do a release", or asked whether something "is ready to release / version bumped".
+description: How to cut a Sapiom Studio desktop release (@sapiom/harness-desktop) in this repo. The release is TRIGGERED BY PUSHING A GIT TAG on main — `vX.Y.Z` — after versions are bumped via changesets. Use whenever asked to "release", "cut/make a release", "ship a new version", "publish/tag the desktop app", "do a release", or asked whether something "is ready to release / version bumped".
 ---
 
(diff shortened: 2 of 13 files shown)


gwitwer:
Review addressed in `0b256dcb`.

- **Blocker (release job gate):** fixed — only `on.push.tags` had moved to `v*`; the job's own `if:` still matched the old namespace, so a `v*` tag would build ~40min and publish nothing. `release-tags.test.ts` now pins both filters.
- **Stale docs:** `release-desktop/SKILL.md`, `PUBLISHING.md`, the README release badge, and both workflow headers updated.
- **Changeset:** added (`patch`).
- **Semver guard:** tightened to `(0|[1-9][0-9]*)` so `01.2.3` no longer passes, matching `semver.valid` and the test helper; an assertion keeps the two in sync.
- **Nits:** all three addressed.

Typecheck clean, 169 tests passing.

PR #730, opened 2026-08-28
chore(desktop): prepare 0.3.8-beta.1 release branch

ynadge:
> [!IMPORTANT]
> **Release-only PR — do not merge into `main`.** After approval, tag this PR head exactly, publish the beta, then close the PR. Stable `main` remains at Desktop 0.3.7 with the Harness 0.8.9 pin.

## Summary

Prepare the first Project dependency graph beta Desktop release without placing a hand-versioned prerelease on `main`. This branch sets Desktop to 0.3.8-beta.1 and packages the current workspace Harness on the beta update channel.

## Changes

- set `@sapiom/harness-desktop` to `0.3.8-beta.1`
- remove both temporary stable pin mechanisms together: `desktopHarnessVersion` and the selective pnpm override
- update the lockfile back to the workspace Harness link
- add a beta changelog entry that describes the workspace Harness without hard-coding a version
- retain the bundle/version guard introduced in #729

## Why this branch must not merge

Changesets is not in prerelease mode. Merging a manually versioned `0.3.8-beta.1` into `main` would allow a later dependency changeset to rewrite it as stable `0.3.8`. Keeping this as a reviewed release branch isolates the beta while stable `main` and GitHub Latest remain on 0.3.7.

## Required release procedure

1. Wait for this exact head SHA to pass CI and review.
2. Tag this exact head as `harness-desktop-v0.3.8-beta.1`; do not merge it.
3. Monitor Linux, macOS, Windows, and final publication.
4. Verify the GitHub release is a prerelease and `harness-desktop-v0.3.7` remains Latest.
5. Close this PR without merging after release verification.

## Testing

- `pnpm install --frozen-lockfile`
- side-effect-free Harness resolution: workspace Harness, with no implicit build
- explicit Harness preparation build: passed
- Desktop typecheck: passed
- Desktop unit tests: 19 files / 155 tests passed
- Prettier and `git diff --check`: passed
- Linux beta packaging: AppImage, deb, and `beta-linux.yml` generated
- packaged Desktop smoke: 13 passed / 1 expected Windows-only skip
- smoke verified Desktop 0.3.8-beta.1, workspace Harness, and beta update channel

## Review refinements

- made the PR explicitly release-only so Changesets never sees the prerelease on `main`
- removed the brittle `@sapiom/harness@0.9.0` assertion from release prose

## Related

Refs: SAP-2906

PR #737, opened 2026-08-29
chore(desktop): drop the temporary Harness 0.8.9 pin [SAP-2906]

gwitwer:
## Why

The SAP-2906 release split kept stable Desktop on the pre-graph Harness with a pnpm override plus a `desktopHarnessVersion` guard. Its own comment said what should happen next:

> *"After tagging 0.3.7, remove BOTH this override and Desktop's `desktopHarnessVersion` field for 0.3.8-beta.1."*

[#730](https://github.com/sapiom/sapiom-js/pull/730) would have done that on `main` and was **closed unmerged**. The beta was tagged off that branch instead, SAP-2906 was marked Done, and `main` kept the pin.

The cost was silent: **0.3.8 shipped as a stable final still bundling Harness 0.8.9**, so the Studio rail rebuild in 0.10.0 — the reason that release was cut — is not in it. Its SPA is byte-for-byte the same generation as 0.3.7's. It also leaves `0.3.8-beta.1` (Harness 0.9.0) semver-*below* a final carrying an *older* SPA, so the final supersedes the beta with a downgrade.

## Changes

Two deletions and the lockfile:

- `pnpm-workspace.yaml` — drop the `"@sapiom/harness-desktop>@sapiom/harness": 0.8.9` override
- `packages/harness-desktop/package.json` — drop `"desktopHarnessVersion": "0.8.9"`

No new machinery. `harness-version.mjs` already resolves `configuredVersion ?? workspaceHarnessPackage.version`, so removing the field makes the guard track the workspace Harness and then verify the packaged bundle really resolved it — the fallback the cleanup was designed around.

## Delivery lane

Standard. No migrations, no infrastructure, no integrations. Changes only which Harness the desktop app bundles.

## Tests

With the pin removed:

- `harness-desktop` resolves `@sapiom/harness` **0.10.0**
- guard: `[harness-version] desktop development dependency: @sapiom/harness 0.10.0 matches the workspace Harness`
- `pnpm --filter @sapiom/harness-desktop test` → **19 files, 155 tests passing**
- the built SPA bundle contains the rail rebuild's "Create new agent" copy — i.e. the payload is verifiably present, not just the version number

Note the workspace Harness must be built before the desktop suite runs (`pnpm --filter "@sapiom/harness..." build`); with the pin in place it resolved a prebuilt npm tarball instead. CI already builds topologically.

## Production signal

The next desktop tag. Ship it as **`v0.3.9-beta.1`** rather than straight to stable — 0.10.0's own changeset warns *"Rows you had may disappear on upgrade… on one real install this cut 42 project rows to 3"*, which is precisely what a pre-release channel is for. Confirm the rail renders against a real multi-project workspace, then promote to `v0.3.9`.

Depends on [#736](https://github.com/sapiom/sapiom-js/pull/736) for the beta to be installable at all — the `harness-desktop-v*` namespace makes pre-release tags invisible to the updater.

Fix-forward: if 0.10.0's rail misbehaves on a real workspace, re-add the override and `desktopHarnessVersion` pinning to the last good Harness and cut the next patch; the guard fails the build loudly rather than silently shipping a mismatch.

https://claude.ai/code/session_015sZDBYwf1SJBVxH5wPYyfc

### DIFF
--- a/.changeset/desktop-drop-harness-pin.md
+++ b/.changeset/desktop-drop-harness-pin.md
@@ -1,0 +1,23 @@
+---
+"@sapiom/harness-desktop": minor
+---
+
+Ship the current Agent Studio experience on the stable desktop channel again, by
+removing the temporary `@sapiom/harness@0.8.9` pin. The app now bundles the
+workspace Harness, so the Project rail rebuild and the dependency graph reach
+stable users for the first time.
+
+**Correcting the record for 0.3.8.** That release's changelog claims
+`@sapiom/harness@0.10.0`, and it did not ship it. The entry was generated from the
+workspace dependency graph, which could not see the `pnpm` override pinning the
+desktop app to `0.8.9` — so `0.3.8` carried the same Agent Studio SPA as `0.3.7`
+despite what its notes said. Nothing was wrong with the build; the pin was
+deliberate (see `0.3.7`) and only its removal was missed. This release is the one
+that actually delivers the newer Studio.
+
+**Rows you had may disappear on upgrade, and no files are touched.** The rail now
+derives its project list from one rule: a project is a directory you chose that
+holds agents. Remembered folders that were an agent's own directory, and folders
+known only because a session once ran there, stop being drawn. On one real install
+this cut 42 project rows to 3 with all 89 agents still visible. Nothing is
+deleted — any folder is one **Add a project** away from coming back.
--- a/packages/harness-desktop/package.json
+++ b/packages/harness-desktop/package.json
@@ -10,7 +10,6 @@
   },
   "type": "module",
   "main": "./dist/main/index.js",
-  "desktopHarnessVersion": "0.8.9",
   "scripts": {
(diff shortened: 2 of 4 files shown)


gwitwer:
Review addressed in `7ddc17d1`.

- **Integrity hash:** restored from `9875c338`. Confirmed it survives a normal `pnpm install` and `--frozen-lockfile`, so the drop was an artifact of running install with `--lockfile-only` rather than something pnpm refuses to record. The 8 remaining removals are all `@sapiom/*` npm entries the pin had pulled — expected.
- **Changeset:** added, `minor` (0.3.8 → 0.4.0). The desktop package's own code is unchanged, but what ships jumps an SPA generation and carries a migration that removes rail rows; nothing depends on this version range, so the number's job is to communicate that. It also corrects the false `0.3.8` changelog entry and carries 0.10.0's upgrade caveat into the changelog desktop users actually read.

155 tests passing.

PR #739, opened 2026-08-29
test(agent-core): add edge case unit tests for parseExecInput

forumevi:
### Summary
Adds edge-case unit test coverage for `parseExecInput` in `client.test.ts`.

### Changes Included
- Added test coverage verifying that empty strings and empty JSON objects are handled as expected by `parseExecInput`.
- Added test cases covering invalid non-string and non-object inputs (e.g., JSON primitives like numbers and booleans).

### Motivation
Ensures input parsing handles non-standard and boundary JSON values cleanly without unexpected runtime exceptions across agent workflows.


gwitwer:
Closing: does not compile. parseExecInput is never imported (TS2304), the new describe is nested inside an unclosed it block, and the assertions compare against the function rather than its result. Happy to take a corrected version.

PR #747, opened 2026-08-30
fix(fetch): preserve __sapiom metadata across internal Request cloning

### LINKED ISSUE #690: [Bug]: @sapiom/fetch loses Request.__sapiom metadata when cloning input
### Affected package or area
@sapiom/fetch

@sapiom/fetch 0.5.0, commit d7d480a

### Environment
- Node.js: v24.14.1
- Package manager and version: pnpm 10.34.3 via Corepack
- Operating system: WSL2 / Linux 5.15.167.4-microsoft-standard-WSL2 x86_64
- Runtime or framework (if relevant): native Fetch API / Request

Reproduction code sample:

```js
const r = new Request("https://example.test/public");
r.__sapiom = { enabled: false };

const copy = new Request(r);

console.log(JSON.stringify({
  original: r.__sapiom,
  cloned: copy.__sapiom ?? null,
  clonedHasOwn: Object.prototype.hasOwnProperty.call(copy, "__sapiom"),
}, null, 2));
```

Steps:
1. Create a Request object.
2. Attach `__sapiom = { enabled: false }` to it.
3. Clone it with `new Request(request)`, matching what @sapiom/fetch does internally.
4. Check whether the cloned request still has `__sapiom`.

Actual output from the reproduction:

```json
{
  "original": {
    "enabled": false
  },
  "cloned": null,
  "clonedHasOwn": false
}
```

The metadata is lost during cloning, so @sapiom/fetch reads no per-request override after `new Request(input, init)`.

### Expected behavior

The documented per-request override should be honored for Request inputs.

When a request is created like this:

```js
const publicRequest = new Request("/api/public");
(publicRequest as any).__sapiom = { enabled: false };
await fetch(publicRequest);
```

@sapiom/fetch should bypass Sapiom authorization/payment handling for that single request.

The metadata should be read from the original Request before cloning, or copied onto the cloned request before interceptors run.

### Actual behavior

@sapiom/fetch clones the input with `new Request(input, init)` before reading `__sapiom`.

Native Request cloning does not preserve custom properties, so `__sapiom` is lost. As a result, a request with `__sapiom = { enabled: false }` is treated as if no per-request override was provided, and the request can still continue through Sapiom authorization/payment handling instead of being bypassed.

Relevant code:
- `packages/fetch/src/fetch.ts` documents `Request.__sapiom` per-request overrides, including `{ enabled: false }`.
- `packages/fetch/src/fetch.ts` clones the input with `new Request(input, init)` before reading `(request as any).__sapiom`.
- `packages/fetch/src/interceptors.ts` also reads metadata from the cloned request.

I checked open issue/PR searches for `__sapiom` / `enabled` and did not find an open duplicate.

### Acknowledgements
- [x] This report does not describe a suspected security vulnerability; I will use the Security Policy for private reporting.

ygd58:
Fixes #690

`sapiomFetch` reconstructs the request via `new Request(input, init)` before reading `__sapiom`. Native `Request` cloning does not preserve custom properties, so a `__sapiom` override set on the original `Request` was dropped before it could be read - the documented per-request bypass (e.g. `request.__sapiom = { enabled: false }`) silently had no effect.

**Fix:** `__sapiom` is now captured from the original `input` before cloning and re-attached to the clone, including after the identity-header re-wrap later in the same function (the same class of loss, one function down).

**Scope:** deliberately limited to the reported reproduction and this identical pattern within `sapiomFetch` itself. `handleAuthorization` (`interceptors.ts`) does its own internal `Request` re-wrap before returning, which could still drop `__sapiom` for a caller relying on it past the authorization step (e.g. at payment-retry time) - left as a separate, deeper propagation concern rather than folded into this fix, since it wasn't part of the reported repro.

**Tests:** added a regression test verifying the `enabled:false` bypass calls `globalThis.fetch` directly (skipping authorization) when `__sapiom` is set on a `Request` input. Verified it fails against the pre-fix code (hits the unmocked authorization path and throws) and passes against the fix.

**Changeset:** added (`@sapiom/fetch`, patch).

Verified locally: `pnpm --filter @sapiom/fetch test` - 44/44 passing (full package), `eslint` clean (the one pre-existing warning in `fetch.ts` is unrelated to this diff).

### DIFF
--- a/.changeset/sapiom-metadata-clone-loss.md
+++ b/.changeset/sapiom-metadata-clone-loss.md
@@ -1,0 +1,23 @@
+---
+"@sapiom/fetch": patch
+---
+
+Fixed `__sapiom` per-request metadata (e.g. `{ enabled: false }`) being
+silently lost when a `Request` instance was passed as `fetch`'s `input`.
+
+`sapiomFetch` reconstructs the request via `new Request(input, init)`
+before reading `__sapiom`. Native `Request` cloning does not preserve
+custom properties, so a `__sapiom` override set on the original `Request`
+was dropped before it could be read - the documented per-request bypass
+(e.g. `request.__sapiom = { enabled: false }`) silently had no effect.
+
+`__sapiom` is now captured from the original `input` before cloning and
+re-attached to the clone, including after the identity-header re-wrap
+later in the same function (the same class of loss, one function down).
+
+Deliberately scoped to the reproduction reported in the issue and this
+identical pattern within `sapiomFetch` itself. `handleAuthorization`
+(`interceptors.ts`) does its own internal `Request` re-wrap before
+returning, which could still drop `__sapiom` for a caller relying on it
+past the authorization step (e.g. at payment-retry time) - left as a
+separate, deeper propagation concern rather than folded into this fix.
--- a/packages/fetch/src/fetch.test.ts
+++ b/packages/fetch/src/fetch.test.ts
@@ -47,4 +47,29 @@ describe("createFetch", () => {
 
     expect(typeof fetch).toBe("function");
   });
+
+  it("should honor __sapiom = { enabled: false } on a Request input, surviving the internal clone", async () => {
+    // sapiomFetch internally does `new Request(input, init)` before reading
+    // __sapiom. Native Request cloning drops custom properties, so this
+    // regression-tests that the metadata set on the *original* Request is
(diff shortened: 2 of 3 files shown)


gwitwer:
Closing in favour of #691, which fixes the same __sapiom metadata loss, was opened first, and also re-attaches metadata after the handleAuthorization rewrap. Thanks.

PR #753, opened 2026-08-31
fix(harness): keep the first-run explainer dismissed across desktop launches [SAP-2991]

gwitwer:
## The bug

The "How Studio is organised" card is specced as one-time and re-openable. On the desktop app it opened on **every launch**.

The component's logic was already correct — `dismiss()` recorded seen, `shouldAutoOpen()` honoured it. The **store** was wrong. It kept the flag in `localStorage`, which is keyed by ORIGIN, and an origin includes the port. The desktop app asks the OS for an ephemeral port on every boot (`harness-desktop/src/main/boot.ts:104`, `srv.listen(0, "127.0.0.1", …)`), so launch N is `http://127.0.0.1:52431`, launch N+1 is `http://127.0.0.1:61208`, and the second one has never seen the card.

The card's own comment reasoned correctly that "I have already been told what a project is" must not live in `ui-prefs` — that blob is the UI's ARRANGEMENT and a user may reasonably reset it — and then picked a store **more** volatile than `ui-prefs` rather than less.

## The fix

`HarnessSettings.helpSeen`, behind `GET`/`PATCH /api/settings`: server-side, per-install, origin-independent, and already the home of this class of fact. The shell reads it from the settings it already holds and PATCHes on dismiss, so the card gains one prop pair and loses its storage entirely.

## A second bug, same shape, found on the way

`telemetryNoticeDismissed` was already a `HarnessSettings` field, and `App.tsx` already PATCHed it on dismiss — but it was **missing from `settingsPatchSchema`**, and zod strips unknown keys. The request 200'd, the response looked right, and nothing reached disk. The telemetry notice was being re-shown for exactly the reason the help card was. One line in the same object; the new server test asserts the **re-read**, not the response, which is what catches this class.

## Verification

Not jsdom — a port-dependent bug cannot be proven there.

**Real harness server, two launches, two ports, same state root, fresh browser profile each time** (a new origin *and* empty storage — strictly harsher than a desktop relaunch). This is the desktop's exact code path: `windows.ts:144` loads `http://127.0.0.1:<ephemeral port>/?token=…`.

```
# launch 1 — port 5401
origin=http://127.0.0.1:5401 helpOverlayVisible=true
dismissed; afterCount=0
localStorage=["sapiom-harness-ui-prefs"]      ← no help key anywhere

/tmp/sap2991-state/settings.json → { …, "helpSeen": true }

# launch 2 — port 5402, same --state-root, new browser profile
origin=http://127.0.0.1:5402 helpOverlayVisible=false
autoOpened=false
reopenedFromAccountMenu=true
escClosed=true
```

Against the acceptance criteria:

- **Survives a restart on a different ephemeral port** — the run above.
- **Survives a `ui-prefs` reset** — the launch-2 profile had no `ui-prefs` at all and the card stayed gone. There is also a new e2e spec that removes only the `sapiom-harness-ui-prefs` key, but it is an **invariant guard, not this bug's regression test**: it passes against the pre-fix code too, since the old flag had its own separate key. What it pins down is that a later "just fold it into `ui-prefs`" simplification has to fail. `rest.test.ts`'s re-read assertion is the only test that fails without this change.
- **Still re-openable from the account menu** — `reopenedFromAccountMenu=true` above, plus the existing spec.
- **Storage failure still shows the card** — absent/unreadable settings read as "not seen", which shows the card; a rejected PATCH still closes it and costs at most one extra showing.
- **Audit note** — `docs/harness-browser-storage.md`.

**One criterion I did not meet literally:** I could not launch the Electron shell itself. The installed `Sapiom.app` was running on this machine and holds the single-instance lock, so a dev Electron quits immediately, and I was not willing to kill the developer's running app. The two-port run above exercises the identical server, bundle, origin and settings file — the Electron chrome is the only untested layer.

Also green: `pnpm build`, `pnpm typecheck`, `pnpm lint`, `rest.test.ts` (80/80), and the **full Playwright suite, 480/480**. The harness vitest suite has 13 pre-existing failures in `core/` fs-watcher specs; identical count and names on a clean `origin/main` build of this machine, untouched by this change.

## The audit

`docs/harness-browser-storage.md` lists every `localStorage` write in `web/src` and sorts them into install-level facts that must move and genuine per-viewer arrangement that can stay:

- **Fixed here** — the explainer flag.
- **Open, highest value** — `ui-prefs.closedProjects`, the removal tombstones: the second symptom of this same hazard, tracked separately. Needs a server endpoint rather than a settings field, since it is per-project.
- **Open** — `ui-prefs.sessionNames` (a server record's name, parked client-side because there is no rename endpoint), `ui-prefs.preferredHarness` (a default, not a layout), and `sapiom-harness-theme` (the awkward one: `index.html` reads it in an inline script before any module loads, so fixing it means the boot HTML carrying the value, not just moving the key).
- **Leave alone** — pane widths, folds, collapse, rail axis/sort, inspector height.

## Note on migration

There is deliberately none. On desktop there was never a durable value to migrate. A CLI user on the stable default port who had already dismissed the card will see it once more, and that is cheaper than carrying a legacy `localStorage` read forever in the file whose whole point is that it no longer reads browser storage.

## Changeset

`.changeset/thick-pianos-hammer.md`, **minor** — a new public `HarnessSettings` field, plus a behaviour change to a published package (`PATCH /api/settings` now persists `telemetryNoticeDismissed`, which every shipped version silently dropped).

Fixes SAP-2991.


gwitwer:
All four addressed in c8b5cbcc — every one of them was right.

**1. Internal doc citation — fixed.** This is the finding I'm most glad you caught. I wrote the audit note against the ticket, which cites that handoff, and carried the path straight across into a file that ships in a public repo. It was both a disclosure that bought the reader nothing and a dead link for everyone who can't open it. Replaced with your generic wording. I also scrubbed the same citation from the PR body. The `SAP-####` ids stay — those are already an established public convention here (`origin/main` carries them in commit subjects).

**2. Changeset — added,** `.changeset/thick-pianos-hammer.md`, `minor`, with both facts. Your reasoning on the second one is the part I'd underweighted: the `telemetryNoticeDismissed` line reads as a one-line schema fix from inside the diff, but from a consumer's side it's a released behaviour change that shows up nowhere in a lockfile. Text is close to your draft.

**3. The ui-prefs spec — you're right, and the PR body oversold it.** The old flag lived under its own key, so clearing `ui-prefs` was always a no-op for it and that spec passes against pre-fix code. I've kept the spec, because the invariant it pins down is real — a later "just fold it into `ui-prefs`" simplification has to fail somewhere — but its comment now says outright that it is an invariant guard and not this bug's regression test, and names `rest.test.ts`'s re-read assertion as the only test that does fail without the change. PR body corrected to match.

**4. `fresh` inheriting a dismissal — fixed,** `helpSeen: this.fresh ? false : readMockHelpSeen()`. I'd considered and waved this one off as unreachable, which was the wrong test to apply: `?mockState=fresh` means "brand-new install", and a brand-new install has no settings file to have recorded anything in, so reading a previous page's dismissal made it a *used* one regardless of whether a spec exercises it today.

Re-verified: typecheck clean, full Playwright suite 480/480 (including `welcome.spec.ts`, which is what uses the `fresh` fixture).

gwitwer:
Both fixed in 9acd5535, and both were right.

**1. Changeset overstates the telemetry fix.** Correct, and I'd inverted the actual upgrade experience. Nothing ever reached disk, so there is no earlier dismissal to honour — every install sees the notice one final time after upgrading, and *that* dismissal is the first one that sticks. Reworded to say exactly that. Thanks for weighting this one as the urgent half: changelog copy is the only line in the diff that can't be edited after publish.

**2. `fresh` discards the write it just made.** Also right, and it's a defect I introduced last round by fixing your round-1 nit at the wrong layer — patching the read to lie about a value the write path was still producing. Went with your first option, at module scope rather than in the constructor:

```ts
if (typeof window !== "undefined" && isFreshMockState()) writeMockHelpSeen(false);
```

Module scope for two reasons: there is one settings file and several `MockApi` instances (the same argument `mockMoves` makes about one disk), and running at import time puts the reset before any construction, so it can never land *after* a dismiss. The read went back to a plain `readMockHelpSeen()`, so the "survives a reload" contract holds again.

I checked the spec you predicted would fail, plus its converse:

- `?mockState=fresh&help=1` → dismiss → reload as `?seed=0&help=1` → **stays dismissed**.
- `?mockState=fresh&help=1` → dismiss → reload as `?mockState=fresh&help=1` → **shows again**, which is right: that URL re-asserts a brand-new install on every load, the same way it re-fabricates empty sessions and workflows.

Both pass. Full Playwright suite still 480/480; typecheck, lint and build clean.

PR #754, opened 2026-08-31
feat(harness): Add a project — use the native folder browser, and cut the copy [SAP-2992]

gwitwer:
## What changed

The **Add a project** dialog reimplemented a file browser in-app and over-explained at every step. This replaces the custom picker with the native folder browser — already implemented, already exposed — and cuts the copy to a title, one line, and one action.

### 1. The picker is the OS one now

`DirectoryPicker.tsx` is **deleted**, not bypassed. It built its own path bar, up-one-level button and scrolling folder list — a worse file browser than the one every user already has (no favourites, no search, no keyboard navigation, no drag target, no network volumes), and ours to keep in sync forever.

`FolderField.tsx` replaces it:

- **Desktop** — `window.sapiomDesktop.chooseDirectory` opens the OS folder browser (`harness-desktop/src/main/dialogs.ts`).
- **The `npx` browser host** — no native dialog exists, so the fallback is the smallest one that keeps the feature: the field plus a native `<datalist>` of the folder's children. Browser chrome, not ours. Without it a browser user would be reciting an absolute path from memory.

`TemplateUseDialog` moves to the same field. That is what makes the deletion real — leaving it on the old picker would have kept exactly the two pickers this ticket is about.

### 2. The copy directs instead of explaining

- The three-sentence block about detection probing one level while adding walks the whole tree is **gone**. That was the scanner describing itself to justify a warning nobody could act on. The behaviour it guarded is unchanged and still tested: a shallow probe never decides a folder has no agents, so the deep scan stays on offer.
- **"No agent directly inside this folder" is gone** with it. The one state a person can act on from here — the folder is not there — is the one line that survives.
- **`sapiom.json` no longer appears in user copy**, asserted in `start-dialog.spec.ts`.

### 3. One primary action

"Add a project" had two ink buttons and nothing on screen said how they differed — because they did not. `openProject` (`use-harness-state.ts`) scans the whole tree after remembering the root, so **"Add every agent under this folder" was a duplicate of "Open project"**, not a competing choice. It is removed, and the primary is now **Add project**.

In the detection flow the two outcomes really are different, so the escape is named **"Open as project"** — it reads apart from **"Add agents"** at a glance, where "Open folder anyway" did not.

## Before / after — real dialog width, both themes

| | Before | After (browser host) | After (desktop host) |
|---|---|---|---|
| Light |  |  |  |
| Dark |  |  |  |

The dialog goes from 486px tall to 199px at the same 440px width. The desktop shot shows the **Choose…** button that opens the OS browser; the browser shot correctly has none, because a control that cannot work must never be shown.

The detection dialog, which shares this surface, and the template destination dialog, which moved to the same field:

| Add existing agents (before → after) | Use a template (after) |
|---|---|
|   |  |

## Verification

**A real harness server, not the mock** — `packages/harness/src/cli/bin.ts` against a 29-agent tree with its own `--state-root`:

```json
{ "title": "Add a project",
  "inAppListing": 0,
  "footer": ["Cancel", "Add project"],
  "bodyText": "Choose a folder to work in — any agents inside come with it.",
  "markerLeak": false,
  "datalistSample": [".../examples/approval-chain", ".../examples/autonomous-pr", ".../examples/axios"],
  "missingFolder": { "hint": "That folder doesn't exist yet.", "primaryEnabled": false },
  "geometry": { "modal": [440, 201], "bodyScrollsSideways": false } }
```

Real completion came from the real `GET /api/fs/list`, not a fixture. The only 4xx in the run was the deliberate probe of a folder that does not exist, which both callers fall back from correctly.

**Mutation-tested guards.** `folder-field.spec.ts` injects the Electron preload's bridge shape so the desktop branch — the one that now carries the whole picker — is covered from a browser run. Each of its four claims was proven by breaking the behaviour and confirming the test fails:

| Mutation | Test that failed |
|---|---|
| Render Choose regardless of the bridge | browser host › no Choose button |
| Write a cancelled pick (`null`) into the field | desktop host › cancelling leaves the folder alone |
| Never render the datalist | browser host › the fallback still completes a path |
| Ignore what the OS dialog returned | desktop host › Choose … takes its answer |

**Suites.** `pnpm build`, `pnpm typecheck` and `pnpm lint` clean. Playwright mock suite: 483 specs, all passing (`popover-crop` failed once in a full parallel run and passes in isolation — it is in the run/step-pipeline flake band the config documents, and touches nothing here).

## did_not_work / notes

- **`GET /api/fs/list` could not be removed**, contrary to the ticket's "may become unused". Three callers still need it: the command palette's path completion, `classifyFolder`, and the browser fallback's completion.
- **The in-app listing's own Retry button is gone** with the listing. An unreadable folder now reports through the dialog's `.modal-error` and recovery is re-pointing the field; `wave5.spec.ts` was repointed to assert exactly that rather than deleted.
- **`pnpm lint` does not cover `web/src`** (it is `eslint src`, and `web/src` is ignored by the eslint config), so nearly all of this diff is unlinted by design. Typecheck covers it.
- Scope note: `StartDialog` serves both the "Add a project" and "Add existing agents" entries, so the copy cut necessarily touched both. The competing-button removal is confined to "Add a project", which is what the ticket owns.

Closes SAP-2992

https://claude.ai/code/session_01A2LUBHGheiUjcwzKmkDo6L


gwitwer:
Round 1 addressed — all three findings were legitimate. Pushed in `e0f0c6a`-ish (see latest commit on the branch).

**1. Confidentiality — agreed, fixed.** The three `StartDialog` comments and the one `start-dialog.spec.ts` comment this PR rewrote no longer name an install's directory layout or its incident numbers. The argument survives intact — a one-level probe cannot answer for an eight-level scan, in either direction — stated as "a folder whose agent sits two levels down was refused outright" and "a folder reported as holding one agent has been measured registering dozens". As you noted, the vocabulary is pre-existing across `mock-data.ts`, `unrooted-agents.spec.ts` and others; I have not touched those, since they are outside this diff.

**2. Changeset — agreed, added.** `.changeset/wild-folders-open.md`, patch on `@sapiom/harness`:

> Agent Studio's folder fields now open the operating system's folder browser on the desktop app, and fall back to path completion in the `npx` browser host; the in-app directory listing is removed. The Add a project dialog is reduced to one line of guidance and one action.

**3. The unreachable ancestor retry — agreed, and the diagnosis was exactly right.** The mock resolves a missing tail to its nearest ancestor instead of rejecting, so `listDir` never rejects under Playwright and that branch had no coverage from any suite. Rather than add React DOM-test infrastructure (there is no jsdom or testing-library in this repo), I extracted the effect's body to `lib/folder-completions.ts` and tested it directly — five cases, including the half-typed-path fallback and the root case that must not retry itself.

Held to the same mutation standard as the rest: inverting the fallback guard (`if (!up)` → `if (up)`) fails *falls back to the parent when the typed tail does not exist yet* and *stops at the root rather than retrying itself forever*, and passes the other three.

`pnpm build`, `pnpm typecheck` clean; the affected e2e specs (144) pass.

gwitwer:
Round 2 nits — both taken.

- **Changeset now names the removed action.** Agreed it is the one change an upgrading user notices as an absence: "…a single action — its second button, "Add every agent under this folder", is gone because adding a project already brings in the agents below it."
- **Fixture cast dropped.** `listing()` is now a plain typed return, so a field added to `FsListResponse` fails typecheck here instead of slipping through the `as`. Verified with `tsc --noEmit -p web/tsconfig.json`; the five tests still pass.

gwitwer:
## Handoff — read before merging

CI is green and the bot review says ship, but **that is not the whole picture.** Two things are outstanding, one of which should block the merge. Written up for whoever picks this up next.

---

### 1. The real OS folder picker has never actually been clicked — blocks merge

This PR makes `window.sapiomDesktop.chooseDirectory` the *only* folder picker on desktop. Every layer of that is verified except the one that matters most:

| What | Covered by | Actually exercises the OS dialog? |
|---|---|---|
| Our side of the bridge contract | `folder-field.spec.ts` — injects a fake `window.sapiomDesktop` into a **mock browser** run | No. Stubbed `chooseDirectory`. |
| The bridge is exposed at all | CI `desktop-smoke` | No — shape-only, and deliberately so: invoking it would open a real blocking OS sheet in CI (see the comment at `harness-desktop/src/main/smoke.ts:724`). |
| The browser fallback | A real harness server, 29-agent tree | Yes, but that is the *fallback* path, not the native one. |

So: no CI job and no human has opened the real macOS folder sheet with this branch in it. The pieces are individually sound and the bridge predates this PR, but the end-to-end desktop flow is **inference, not evidence**.

**What to do:** build and run `packages/harness-desktop`, open **Add a project**, and confirm by hand:

1. The **Choose…** button is present (it is feature-detected, so its absence means the bridge did not resolve).
2. Clicking it opens the OS sheet **at the folder already in the field** — the SPA passes the current value as `defaultPath`.
3. Picking a folder writes it into the field and **Add project** creates the row.
4. **Cancelling leaves the field untouched** — `showOpenDialog` resolves `null`, and `null` must never be written in.

---

### 2. The `npx` browser host is a real downgrade, and the PR body undersold it

The webapp **cannot** open the OS picker — there is no bridge, so no Choose button renders. Its fallback is the text field plus a native `<datalist>` of the current folder’s children.

Drilling down still works (picking an option sets the field, which re-fetches that folder’s children), but compared to the deleted in-app browser, browser users **lose**:

- the up-one-level button (you now edit the path text instead),
- arrow-key navigation of the folder list,
- the **“Agent” badges** that marked which folders were agent projects (`hasAgentProject`),
- and datalist rendering varies by browser.

The ticket asked for exactly this trade — *“this is a deletion, not a build … keep a minimal fallback for the `npx` browser host”* — so it is not a deviation. But the PR body described what the datalist **adds** without naming what it **removes**, and that framing was too generous. Calling it out here so the trade is decided rather than absorbed.

---

### Open question for a human

Is the npx host secondary enough that this trade is fine? Three options, cheapest first:

| Option | Cost | Risk |
|---|---|---|
| **(a)** Leave as shipped | none | Browser users type more |
| **(b)** Re-add recent-folder chips to `StartDialog` | one prop — `FolderField` already renders them, `StartDialog` passes `recentDirs={[]}` | none; it is not a picker |
| **(c)** Keep a minimal listing for the browser host only | moderate | Re-opens the “two pickers drift” problem the ticket set out to close |

Note (b) is **not** a regression fix — `DirectoryPicker` was also passed `[]` in this dialog, so chips were never there. It is just the cheapest way to soften (2) without re-litigating the deletion.

---

### Where things are

- Repo is **`sapiom-js`**, not `Sapiom` — easy to get wrong.
- Worktree: `~/sapiom/sjx/2992`, branch `feat/SAP-2992`, head `97d5b79`.
- Three commits: the change, the round-1 review fixes, the round-2 nits.
- Real-server run used for verification:
  `pnpm exec tsx src/cli/bin.ts <a real agent tree> --port <port> --state-root <scratch dir> --no-session --no-telemetry --no-open`
  `--state-root` is **mandatory** — without it you mutate the real `~/.sapiom/harness`.
- E2E: `npx playwright test --config web/e2e/playwright.config.ts` from `packages/harness`. `E2E_PORT` lets you avoid colliding with a running stack.

### Known noise — do not chase these

- **12 unit failures in `src/core/` and `src/server/`** (watcher / registry / graph) fail **identically on unmodified `origin/main`** locally. Pre-existing and environment-dependent; CI is green on them.
- **`popover-crop.spec.ts`** failed once in a full parallel run and passes in isolation. It is in the run/step-pipeline flake band the Playwright config documents, and touches nothing here.
- **`pnpm lint` does not cover `web/src`** — it is `eslint src`, and `web/src` is eslint-ignored. Nearly all of this diff is unlinted by design; typecheck is what covers it.

### Scope notes

- **`GET /api/fs/list` could not be removed**, contrary to the ticket’s guess — the command palette’s path completion, `classifyFolder`, and the browser fallback all still use it.
- The in-app listing’s own **Retry button** went with the listing. Errors now surface through the dialog’s `.modal-error` and recovery is re-pointing the field; `wave5.spec.ts` was repointed to assert that rather than deleted.
- `StartDialog` serves both **Add a project** and **Add existing agents**, so the copy cut necessarily touched both. The competing-button removal is confined to **Add a project**, which is what the ticket owns.

PR #761, opened 2026-09-01
feat(content-generation): public model aliases + select passthrough [SAP-2582] [SAP-2580]

gwitwer:
## Summary

Makes the content-generation model surface alias-first and adds capability-based selection.

Public semantic aliases (`flux-fast`, `veo3-fast`, …) are the supported way to pick a model. Raw provider ids are now deprecated: still accepted, still typed, still sent on the wire unchanged — but the docs and types steer callers to aliases ahead of the platform closing that path in a future release.

Also adds the `select` directives and the `preferSatisfied` response echo, both already supported by the platform.

**This change is non-breaking.** No runtime path changes, nothing is removed, and every request shape that worked before still works and still compiles. Verified with a strict-mode back-compat probe covering a raw id pinned on the image path, a `VIDEO_MODELS` pin, an uncataloged id, and a plain `string` variable rather than a literal — the case that actually breaks under type narrowing. `tsc --noEmit --strict` exits 0.

## What changed

**Model selectors**

- New `IMAGE_MODELS` / `KnownImageModel` and `VIDEO_MODEL_ALIASES` / `KnownVideoModelAlias`, listing the public aliases the platform serves.
- `ImageCreateInput.model` retyped from bare `string` to `LiteralUnion<KnownImageModel>` — editor autocomplete for the known aliases, any string still accepted.
- `VIDEO_MODELS` (raw provider ids) is unchanged and still exported; its `@deprecated` notice now gives a per-entry migration. `minimaxVideo01` is flagged separately because it has no cataloged alias and needs repointing rather than renaming.
- No client-side model validation is added, by design: a newly cataloged alias must work before the SDK catches up, so the platform catalog stays the authority on what is accepted.

**Selection (`select`)**

- `ImageSelect` and `VideoSelect` are separate types rather than one shared shape, so each media type only accepts capability tags it can satisfy. `ImageSelect.requires` is `("referenceImage")[]`; `VideoSelect.requires` is `("audio" | "lipsync" | "referenceImage")[]`. `prefer` is shared through a private base, and `MediaSelect` remains exported as the union for generic code.

  `ImageSelect` carries no `requires` at all. Every image entry in the platform catalog declares `referenceImage: false` and none declares `audio` or `lipsync`, so any tag would narrow to zero candidates — a published option that could only fail. Adding `requires` to `ImageSelect` later is non-breaking; removing it later would not be, so it is omitted until the catalog can satisfy one.

  This matters because it is free now and breaking after publish: previously `images.create({ select: { requires: ["lipsync"] } })` type-checked and failed only as a runtime 400. It is now a compile error (TS2353, via excess-property checking), and `MediaSelect` is passable to both `images.create` and `video.create` — both directions confirmed with strict-mode compile probes, positive and negative, and pinned by spec guards so neither can silently regress.
- `preferSatisfied?: boolean` typed on both results, both dispatch envelopes, and both launch handles. It is omitted entirely when no preference was requested rather than defaulting to `false`, so a degrade is never reported for a comparison that never ran.
- On async paths the flag resolves at submit and the polled queue response carries no submit metadata, so it is threaded off the handle exactly as `cost` and `resolvedModel` already are. The private `withDispatchCost` helper is renamed `withDispatchMetadata`.

**Docs**

- `content-generation/README.md` gains a "Choosing a model" section with the alias tables and a `select` subsection.
- Claims the SDK cannot enforce were removed: catalog order is described as a stable listing rather than "cheapest first" (pointing at `select.prefer` for actual cheapness), and `requires`-against-a-pin is stated as intent rather than guarantee. The same ordering over-claim was fixed in the `IMAGE_MODELS` JSDoc.
- Changeset: `@sapiom/tools` minor.

**Provider-neutral copy check** (`scripts/provider-neutral-copy-check.mjs`)

Review caught a provider token in `content-generation/README.md`, which publishes in the npm tarball and is therefore unretractable. The root cause was an audit gap, not a typo: **no published README was audited at all**, so a green `provider-copy:check` was never evidence for any of them.

The target list is now **derived** from each package's own `files` entries rather than enumerated, so a new file or directory is audited the moment it exists — verified by planting a leak in a brand-new package directory and watching it get caught with no list to update. Both `files` forms count: markdown globs and plain directory entries (`agent-core` ships `skills` and `templates`; `cli` ships `templates`), which together were publishing 13 unaudited files. `.changeset/*.md` is audited too — the one surface that cannot be edited after publish. CHANGELOGs stay excluded as append-only history. Coverage: 79 → 127 files.

`dist` is deliberately **not** walked, which is a judgement call worth a second opinion: its markdown is generated from `src`, which is audited, and walking it would make the audited-file count depend on build state. The trade-off is that markdown committed directly into a published `dist` would go unaudited.

Worth knowing for anyone reasoning about the original leak: `fal-ai/flux/schnell` is an *allowed* contract token repo-wide. Only the `fal-ai/…` ellipsis form fails, because the `…` falls outside the unmask character class. That is precisely why the leak slipped through.

**One exemption.** Deriving from `files` also pulled in package-root READMEs, and `packages/langchain-classic/README.md` has 14 hits. Those are legitimate: the package exports `SapiomChatOpenAI` / `SapiomChatAnthropic` over the `@langchain/openai` / `@langchain/anthropic` peers, so its README cannot document the package without naming providers. Rather than skipping the file, six narrow patterns unmask those specific usages through the existing `REQUIRED_CONTRACT_PATTERNS_BY_PATH` — so the rest of that published README stays audited and a future `fal` or `blaxel` in it is still caught. Verified by planting both.

The script also no longer uses `glob` from `node:fs/promises`, which needs Node ≥22 while `engines` allows 18 and the test matrix runs 20.x — it made `pnpm provider-copy:check` fail at import for contributors on those versions. CI never caught it because the job running that script pins 22.x. Confirmed fixed by executing the check and its test suite under Node 20.20.2.

## Testing

New `index.spec.ts` coverage: an alias `model` forwarded unchanged; a raw id still forwarded untouched (proving no client-side rejection was introduced); `select` present on the body when set, absent when unset, dropped when explicitly `null`; and `preferSatisfied` threaded through the polled and launch-handle paths, and omitted when no preference was requested.

Two pre-existing tests that incidentally pinned a raw provider id now use an alias. The deliberate legacy-echo test is unchanged.

Full `@sapiom/tools` suite passes (34 suites, 619 tests), plus typecheck, lint, prettier, `provider-copy:check` and `examples:check`.

Note: prettier reformatted four pre-existing `applyMediaParams(...)` call sites already non-compliant on `main`. Unrelated churn, required to pass the format check.

## Out of scope

`examples/scene-to-video/index.ts` pins two ids with no cataloged alias (image-to-video and a video merge utility), so there is nothing to migrate them to yet. That needs a platform catalog entry rather than an SDK change and is tracked separately.

Rollback: types, docs and request passthrough only — revert the commit, or patch-release `model` back to `string`. No server state or runtime path is touched.

Refs: SAP-2582, SAP-2580


gwitwer:
/review

Re-review please — all three round-2 findings are addressed in `e520ad9a`, and one of them was fixed at the root rather than at the symptom, which is worth your scrutiny.

1. **README provider token** — fixed at `README.md:236`. But the root cause was an **audit gap**: no published README was in `STATIC_PROVIDER_NEUTRAL_COPY_TARGETS` at all, so the green `provider-copy:check` was never evidence for any of them. All 14 `packages/tools/src/*/README.md` are now audited (93 files vs 79, still passing). That change to `scripts/provider-neutral-copy-check.mjs` is scope beyond the finding — please judge whether it belongs here or in its own PR.

2. **`MediaSelect` split** into `ImageSelect` / `VideoSelect` so each media type only accepts tags it can satisfy, with a `@ts-expect-error` guard against the image vocabulary widening back. Please check the tag split against the platform catalogs independently rather than trusting the PR body.

3. **Unenforceable claims removed** from the README and, additionally, from the `IMAGE_MODELS` JSDoc, which had the identical ordering over-claim but was not in the finding.

Two things to weigh specifically:

- **Non-breaking is the load-bearing claim here.** It was verified with a strict-mode probe including the plain-`string`-variable case, not just literals. If you can construct a caller shape that compiled before and does not now, that is the finding that matters most.
- The stale `withDispatchCost` references are updated in all five places, including the stub package.

gwitwer:
/review

Re-review please — `763a857d` addresses both new findings, and there is one judgement call in it I want you to arbitrate rather than rubber-stamp.

**Finding 1 — resolved by removing `requires` from `ImageSelect` entirely**, which was your nit. With `ImageSelect` reduced to `{ prefer?: "cheapest" }` every member field is optional, so `MediaSelect` is assignable to both inputs and the JSDoc claim becomes true. Confirmed both directions with strict-mode probes: `MediaSelect` into `images.create` / `video.create` compiles clean, and `createImage({ select: { requires: ["lipsync"] } })` still fails TS2353. Note the guard mechanism changed — excess-property checking now does the work that tag narrowing used to — so please satisfy yourself that is genuinely equivalent protection and not a weaker check that happens to pass today.

**Finding 2 — the target list is now derived from each package's `files` entries**, not enumerated, so a new directory's README is audited on creation. Verified by actually creating one and watching it get caught. `.changeset/*.md` added (all 8 pending were already clean). Coverage 79 → 113.

**The judgement call I want you to rule on:** deriving from `files` pulled in package-root READMEs, and `packages/langchain-classic/README.md` fails with 14 hits that are *legitimate* — it exports `SapiomChatOpenAI` / `SapiomChatAnthropic` and cannot document itself without naming providers. It is exempted as a single named file rather than widening the regex, so any second exemption must be argued. That file was never audited before this PR either, so it is not a regression. But the alternative — unmasking just the exported class names and keeping the rest of that README audited — is narrower and arguably better. Please pick, and say if you think this whole script change should be split into its own PR.

One incidental: an `eslint-disable` added for the now-empty `ImageSelect` was removed after confirming the rule is not actually enabled, so the directive was dead weight.

gwitwer:
/review

Re-review please — `7a2d3504` + `89791351` close all three round-4 findings. Note `7a2d3504` predates your last pass, so finding 2 was already fixed when you reviewed `763a857d`.

**Finding 2 — closed at the type level, not by narrowing the copy.** Rather than weakening the changeset claim, `ImageSelect` now carries `requires?: never` and `MediaSelect` is the shared base (`{ prefer? }`) rather than the union. All three rejections are now **TS2322 on the property itself**, where before they were TS2353 (excess-property) and TS2559 (weak-type) — so protection no longer depends on which incidental mechanism fires for a given call shape. Your specific "`requires` and all" hole is now **TS2339**: under the shared shape `MediaSelect` has no such property to carry. The changeset claim is therefore literally true and was kept strong. Please confirm you agree it is now accurate rather than merely narrower.

**Finding 1 — non-`.md` `files` entries are treated as `<entry>/**/*.md`**, via an entry-to-RegExp translator since `files` mixes globs, literal paths and directories. Verified by planting a leak in `agent-core/skills/**` and watching it get caught, not by counting files. 113 → 127.

**Judgement call for you:** `dist` is deliberately not walked — its markdown is generated from audited `src`, and walking it would make the audited count depend on build state. But `dist` markdown does publish where a package's `files` names it. Rule on whether that is the right line.

**Finding 3 — rewritten on `readdir({ withFileTypes })`.** Every remaining API audited against Node 18. **Stated limitation:** only Node 24/25 are installed here, so the audit is static — an executed run under Node 20 is in progress and I will report it. Please treat the Node-compat claim as unverified until I confirm, rather than as tested.

**Exemption — took your narrower option.** Six patterns rather than the three you listed: section comments (`// OpenAI`) and example variable names (`const openai`) are separate hits from the package and class names. Under `REQUIRED_CONTRACT_PATTERNS_BY_PATH`, so no second mechanism. Verified by planting `blaxel` and `fal` in that README and confirming both are still caught — your point that a whole-file skip makes future leaks permanently invisible was the decisive one.

gwitwer:
/review

Superseding my previous request — head is now `96a483a4` (the earlier trigger fired at `89791351`). All round-4 findings are closed, and the Node-compat claim I asked you to treat as **unverified** is now **executed proof**, not a static audit.

**Finding 3 — verified by running it, and the result corrects my own description.** Against the absolute Node 20 binary:
- current script → passed, 127 files, exit 0
- the script as of `7a2d3504` → exit 1, `SyntaxError: The requested module 'node:fs/promises' does not provide an export named 'glob'`

So the break was real and is genuinely fixed. Correction worth recording: it fails as an **import-time SyntaxError**, not the call-time `glob is not a function` I described earlier — slightly worse, since nothing in the script runs at all.

Worth knowing how nearly this went the other way: the first Node 20 run reported success, but under v25.8.0, because `nvm use` does not survive between shell invocations here. That would have been false proof of exactly the claim in question. The numbers above come from invoking the v20.20.2 binary by absolute path.

**Finding 1** — non-`.md` `files` entries handled at line 204 as `${entry}/**/*.md`. 113 → 127, verified by planting a leak in `agent-core/skills/**`, not by counting.

**Exemption** — the whole-file skip is gone; patterns live under `REQUIRED_CONTRACT_PATTERNS_BY_PATH`. **Six patterns, not the three you listed**: section comments (`// OpenAI`, `// Anthropic`) and example variable names (`const openai`, `const anthropic`) are four further distinct hits, and three patterns leave the file failing. Verified `blaxel` and `fal` planted in that README are still caught, so the narrow unmask did not blind it.

**Finding 2** — closed at the type level as described; `96a483a4` adds a comment recording that with `exactOptionalPropertyTypes` off (confirmed unset in every tsconfig here) `requires?: never` is `never | undefined`, so the diagnostic names `undefined` rather than `never`.

**Still yours to rule on:** whether `dist` should be walked. It is currently excluded — its markdown is generated from audited `src`, and walking it would make the audited count depend on build state — but `dist` markdown does publish where a package's `files` names it.

gwitwer:
/review

`debda47d` closes the README finding. **One correction to the premise your `dist` ruling rested on — please re-rule.**

**README:287/298 — fixed.** Both lines now describe `requires?: never` as a prohibition rather than an omission, and the reason is carried across from the JSDoc with the concrete case, since that is what makes it non-obvious:

```typescript
const opts = { prefer: "cheapest", requires: ["lipsync"] };
await sapiom.contentGeneration.images.create({ prompt, select: opts }); // caught
```

That is the case an omitted property would have missed. The `exactOptionalPropertyTypes` wording note and the non-breaking forward path are both kept.

**`dist` — your ruling said "every package build here is `tsc` / `node --check`, nothing copies markdown into `dist`". That is not quite true**, and I relayed it without checking, so this is on me as much as anything. Auditing all 18 build scripts turned up two copy/bundle steps:

- **`@sapiom/harness` is published, ships `dist`, and runs `vite build` into `dist/web`.** Vite copies its `publicDir` **verbatim**, so a `.md` dropped in `web/public` would publish unaudited. There is none today, so your conclusion still holds — but that is the live path by which it breaks, and it is far more specific than "no build copies markdown".
- `harness-desktop` also has a copy step but is `private: true`, so it never publishes.

The comment now names that path explicitly as the re-check condition, states the invariant rather than the count-stability rationale, and records the asymmetry you flagged: `bin`, `lib`, `skills` and `templates` are walked; only generated-from-`src` output is skipped. Count stability is noted as the secondary benefit it is.

Given the corrected premise, please confirm the exclusion is still the right line, or say if the Vite `publicDir` path is enough to warrant walking `dist` after all.

**Node compatibility — executed, not inferred.** `provider-copy:check` passes under Node 20.20.2 (output self-reports the version) and the script self-tests are 6/6 there. Negative control: the same Node 20 against the pre-fix script exits 1 with an import-time `SyntaxError`.

gwitwer:
/review

`b31d11cc` fixes README:305. You were right, and the negative control proves it rather than us agreeing by inspection:

| Variant | Result |
|---|---|
| as written, guard **present** | errors — blames `requires` |
| as written, guard **removed** | **still errors** — blames `prefer` |
| `as const`, guard present | errors — blames `requires` |
| `as const`, guard **removed** | **compiles** |

Only the last pair isolates the guard. The original snippet failed for a reason unrelated to the claim it was making. Fixed with `as const`, plus a comment marking it load-bearing so it does not get "tidied" away and silently break the demonstration.

**Audited every other snippet rather than only the line you named.** All compile as written against the shipped types: the `IMAGE_MODELS.fluxStandard` pin, the video `select` example including its `out.resolvedModel` / `out.preferSatisfied` reads, and the image `prefer`-only call. Also checked a claim nobody raised — that the README alias tables match the exported maps — by extracting both and diffing. No drift in either direction.

One methodological note worth recording, since this PR has now produced three distinct ways a check can mislead. Two probe runs initially reported failures that were artifacts of an ad-hoc `tsc` config: top-level `await` in a CJS-inferred file, and a `lib` mismatch surfacing pre-existing `sandboxes` errors absent from the real `pnpm typecheck`. Diagnostics were scoped to the file under test rather than either being reported as a finding. So alongside "the check does not cover the case" and "the artifact passes for the wrong reason", there is **"the probe's own setup is broken"** — which produces false negatives where the other two produce false positives.

**Verification:** tsc exit 0, eslint clean, prettier clean, `provider-copy:check` 127 files on Node 25 **and** Node 20.20.2, script self-tests 6/6 on both. `dist` ruling noted, nothing further there.

PR #762, opened 2026-09-01
feat(harness): fetch the Studio system prompt from the backend [SAP-2810]

gwitwer:
## What

Agent Studio fetches its coding-agent system prompt from the backend on session start instead of using only the copy baked into this package.

- `packages/harness/src/profiles/system-prompt-fetch.ts` mirrors `packages/mcp/src/instructions-fetch.ts` deliberately — same 5s `AbortController` bound, same fallback set (non-200, empty body, network error, timeout), same never-throws contract. `GET {apiURL}/v1/harness/system-prompt`, public, no auth.
- The server loads it per session start (create and resume) and hands it to `generateSystemPromptFile`, so the file claude-code inlines via `--append-system-prompt` (and codex as `developer_instructions`) carries the served text. The rehydration appendix path is unchanged.
- `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1` pins the bundled prompt: an escape hatch for an air-gapped run, and how `src/test-setup.ts` keeps the ~10 specs that boot the real server off the network — the same global-kill-switch pattern already used there for telemetry.

## Why

The prompt shipped inside the package, so every teaching improvement reached only users who upgraded — and Studio feedback shows they don't. `@sapiom/mcp` already solved this for its own instructions; this is that mechanism on the Studio surface. SAP-2810.

## Companion PR

sapiom/Sapiom#4776 serves the content (`GET /v1/harness/system-prompt`, DB-backed, updated by content-release migration). Merge that first — until it deploys this fetch 404s and falls back to the bundled prompt, which is the current behaviour, so the ordering is safe either way.

## Drift guard

`profiles/default.test.ts` pins sha-256 of `DEFAULT_SYSTEM_PROMPT` to the digest the backend pins for `DEFAULT_HARNESS_SYSTEM_PROMPT`. The two are one canonical text: the bundled copy is what a session gets when the fetch fails, the one path with no other source of truth. Editing either reddens both suites, so a change has to move the body and both pins together. The `@sapiom/mcp` copies drifted 2.6 → 2.8 across two releases with every spec green (SAP-2959) — this is that lesson applied up front. It cannot block a merge in the other repo; it removes the silent path.

## Tests

- `profiles/system-prompt-fetch.test.ts` — 200 body, endpoint URL, non-200, empty body, network error, timeout (fake timers), the disabled switch making no request, environment resolution, and fallback when the credential store is unreadable.
- `profiles/default.test.ts` — cross-repo digest + content guards.
- Verified locally: `pnpm build`, `pnpm --filter @sapiom/harness typecheck`, `pnpm --filter @sapiom/harness lint` all clean; `vitest run src/server src/core/inject src/profiles` → 450 passed. Two files fail identically **on unmodified `origin/main`** in this environment (`auto-bind-rescan.test.ts`, `system-graph-freshness.test.ts`) — verified by reverting the change and re-running; unrelated to this PR.

Closes SAP-2810 (harness half).

https://claude.ai/code/session_01TJxJVFvjVujSTKvfQeG9kG

### DIFF
--- a/.changeset/wise-prompts-deploy.md
+++ b/.changeset/wise-prompts-deploy.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": minor
+---
+
+Agent Studio now fetches its coding-agent system prompt from the Sapiom backend on session start (`GET /v1/harness/system-prompt`), instead of using only the copy baked into this package. Prompt improvements now reach a session after a backend deploy rather than after an npm upgrade. The bundled prompt remains the offline fallback — a non-200, empty body, network error or 5s timeout starts the session on it, exactly as before — and `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1` pins it deliberately.
--- a/.changeset/wise-prompts-deploy.md
+++ b/.changeset/wise-prompts-deploy.md
@@ -2,4 +2,6 @@
 "@sapiom/harness": minor
 ---
 
-Agent Studio now fetches its coding-agent system prompt from the Sapiom backend on session start (`GET /v1/harness/system-prompt`), instead of using only the copy baked into this package. Prompt improvements now reach a session after a backend deploy rather than after an npm upgrade. The bundled prompt remains the offline fallback — a non-200, empty body, network error or 5s timeout starts the session on it, exactly as before — and `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1` pins it deliberately.
+Agent Studio now fetches its coding-agent system prompt from the Sapiom backend on session start (`GET /v1/harness/system-prompt`), instead of using only the copy baked into this package. Prompt improvements now reach a session after a backend deploy rather than after an npm upgrade. The bundled prompt remains the offline fallback — a non-200, empty body, network error or 5s timeout starts the session on it, exactly as before.
+
+The request is unconditional: it fetches configuration rather than reporting usage, so it is not gated on the telemetry opt-in, and it carries no session content, identifiers or API key. Set `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1` (or `true`) to skip it and always use the bundled prompt — see the README's "Outbound requests" section.
--- a/packages/harness/README.md
+++ b/packages/harness/README.md
@@ -39,6 +39,21 @@ session lifecycle) to improve Sapiom. Opt out any time; `--no-telemetry`
 disables collection entirely. Events are also written locally to
 `~/.sapiom/harness/events.ndjson` for your own inspection.
 
+## Outbound requests
+
+Beyond telemetry (above) and the Sapiom calls your own actions make (sign-in,
+Deploy, Prod Run), Agent Studio makes one request of its own:
+
+- **System prompt, on every session start** — an unauthenticated
+  `GET https://api.sapiom.ai/v1/harness/system-prompt`, so the Studio conventions
+  your coding agent is told about can improve without you upgrading this package.
+  It sends no session content, no identifiers and no API key, and it is *not*
+  gated on the telemetry opt-in — it fetches configuration rather than reporting
+  usage. It is bounded at 5 seconds and falls back to the prompt bundled in this
+  package on any failure, so an offline session behaves exactly as before.
+  `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1` (or `true`) skips the request entirely
+  and always uses the bundled prompt.
+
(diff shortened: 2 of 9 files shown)


gwitwer:
Round 1 addressed in a76386c3. Taking each:

**Confidentiality (blocking) — partly agreed, done.** Dropped the incident narrative: nothing in this repo now says which released `@sapiom/mcp` versions shipped stale instructions. On the backend file path I'll note the precedent rather than claim you were wrong to flag it — `packages/mcp/src/instructions.ts` already carries `KEEP IT IDENTICAL to the backend's DEFAULT_MCP_INSTRUCTIONS (Sapiom repo, backend/src/mcp/mcp-instructions.constants.ts)` on main, for exactly this cross-repo pin. I removed the path anyway (it earns little once the generic role is stated) and kept the `SAP-####` refs.

**Consent surface (blocking) — disclosure done; the gating I'm declining, with reasoning.** `packages/harness/README.md` gains an **Outbound requests** section: the request, that it sends no session content, identifiers or API key, the 5s bound and bundled fallback, and `SAPIOM_HARNESS_PROMPT_FETCH_DISABLED=1|true`. The changeset now says the call is unconditional.

I'm not tying it to `DO_NOT_TRACK` / `SAPIOM_TELEMETRY_DISABLED`. It fetches configuration rather than reporting usage — the same shape as `@sapiom/mcp`'s existing unconditional instructions fetch on connect, which is not telemetry-gated either. Folding config fetches into the tracking opt-out would also mean a user who opts out of telemetry silently gets a stale prompt, which is the failure this ticket exists to remove. The dedicated flag is the opt-out, now documented where a user will find it.

**Untested wiring — agreed, and the sharpest note in the review.** You were right that the seam was dead and the mutation you described was green. `loadSystemPrompt` is now threaded through `HarnessServerOptions`, and `src/server/served-system-prompt.test.ts` boots the real server and asserts the served text is what lands in `<generated>/<id>/system-prompt.txt` — on `create()` and again on `resume()` (a redeployed prompt reaching a continued session). Deleting `prompt` from the `generateSystemPromptFile` call now fails that spec. One thing the test surfaced: an injected loader that *throws* took session creation down with it, so the call is now `.catch`-guarded to the bundled prompt, matching how the rehydration brief is handled a few lines above.

**Flag convention — agreed, done.** `isEnvFlagSet` is exported from `cli/consent.ts` (rather than adding a third copy of the normalizer) and used here, so `=true` opts out. Covered by a test.

**Nits — both done.** `PINNED_PROMPT_DIGEST`, with the comment saying plainly that it is a snapshot of the local prompt and that the backend holds the twin pin. `originalFlag` moved into `beforeEach`.

Verification: `pnpm --filter @sapiom/harness typecheck` and `lint` clean; `vitest run src/profiles src/server src/core/inject src/cli` → 443 passed. The two files that fail (`auto-bind-rescan`, `system-graph-freshness`) fail identically on unmodified `origin/main` in this environment — checked by reverting the change and re-running.

Companion backend PR sapiom/Sapiom#4776 (also updated for its own review).

gwitwer:
Fixed in 7559822c, and one correction of mine to record.

**README enumeration — agreed.** I checked both paths you named: `productAnalyticsOptIn` defaults to `true` (`shared/types.ts`), and the SPA's PostHog config falls back to the bundled key and `https://us.i.posthog.com` unless `SAPIOM_POSTHOG_KEY` is explicitly empty (`server/static.ts`) — so it is on by default, a different host, and not the tier the Telemetry section above describes. `npx @sapiom/mcp@latest` is per session, and the MCP server then fetches its own instructions. The section no longer implies completeness: it now claims one Sapiom request of Studio's own, and names the app's product analytics and the `npx` launch as separate paths of their own rather than silently omitting them. I stopped short of documenting their opt-out mechanics here — that belongs with those features, not in this PR.

**Your correction is right and mine was wrong.** `packages/mcp/src/instructions.ts` on `origin/main` contains no backend reference; `instructions.test.ts` uses the generic "backend copy" phrasing. I read that "(Sapiom repo, backend/src/mcp/mcp-instructions.constants.ts)" line in a local checkout sitting several releases behind `main`, and cited it as current without verifying against `origin/main`. The precedent I claimed does not exist, so the confidentiality finding stood on its own merits, not on my having complied anyway. Thanks for checking it rather than letting it pass.

CI was green on `4d354898` (the commit before this doc change) across `test (20.x/22.x)`, `playwright-mock`, `desktop-smoke`, `examples` and CodeQL.

PR #764, opened 2026-09-01
feat(harness): initialize durable Agent Map project state

ynadge:
## Summary

Introduces the additive durable foundation for the plan-first Agent Map. Studio now assigns stable project IDs through a server-private catalog and lazily creates one empty, restart-safe Agent Map workspace record only when its protected read endpoint is opened.

## Changes

- Add strict, path-free Agent Map project/workspace/error contracts with independent schema and record versions.
- Persist `studio-projects.json` plus per-project `agent-map/projects/<projectId>/workspace.json` records using cross-host serialized catalog writes and atomic no-clobber workspace creation.
- Reconcile existing allow-listed workspace scopes to durable project IDs while keeping roots and legacy WorkspaceKeys private/migration-only.
- Mount boot-token-protected project create/root-association routes plus `GET /api/projects/:projectId/agent-map/workspace`, with allow-listed roots, bounded errors, and path-free `no-store` responses.
- Publish path-free project summaries through AppState and add strict real/mock browser parsing.
- Emit bounded initialization/read-failure telemetry without paths, repository material, record bodies, or raw errors.
- Preserve the legacy System Graph, per-agent Canvas, discovery, and watchers unchanged; catalog failures degrade back to the legacy AppState projection.
- Add a minor `@sapiom/harness` changeset covering the new published behavior and state-root files.

### Review refinements

- Use an atomic hard-link commit for first workspace creation so independent store instances return one winning timestamp and emit initialization once.
- Serialize catalog writers across CLI/Electron hosts, re-read after acquiring the file lock, and keep reads fresh across live instances.
- Route root moves and additional bindings through authenticated, allow-listed project association endpoints; a path/hash never selects durable identity.
- Remove private `repositoryId` from public summary types, serialization, browser parsing, mocks, and response tests.
- Reject same-project target-root collisions before persistence and prove the catalog remains restart-readable.
- Skip unsafe live input scopes without treating them as corrupt persistence, while accepting legal path names with surrounding spaces.

This is the main-based foundation PR. SAP-3057 and SAP-3055 can stack on this branch while it is under review.

## Testing

- [x] `pnpm --filter @sapiom/harness exec vitest run` — 179 files, 3,019 tests passed
- [x] `pnpm --filter @sapiom/harness lint`
- [x] `pnpm --filter @sapiom/harness build`
- [x] `git diff --check`

## Related

- Closes: SAP-3056
- Linear: https://linear.app/sapiom/issue/SAP-3056

## Checklist

- [x] Additive migration boundary; no legacy route/store/loader removed
- [x] No local paths or legacy WorkspaceKeys in Agent Map payloads
- [x] Cross-instance lazy initialization, concurrent catalog writers, and restart behavior covered
- [x] Malformed, future-schema, unknown-project, authorization, and storage failures covered
- [x] Self-reviewed; no hardcoded secrets

### DIFF
--- a/.changeset/calm-maps-arrive.md
+++ b/.changeset/calm-maps-arrive.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": minor
+---
+
+Add durable, path-free Studio project identities and lazy Agent Map workspace state. The authenticated local server now exposes project workspace and root-binding association endpoints, and stores the new catalog at `studio-projects.json` with per-project records beneath `agent-map/` in the configured harness state root. The legacy System Graph and per-agent Canvas remain unchanged.
--- a/packages/harness-desktop/src/main/boot.ts
+++ b/packages/harness-desktop/src/main/boot.ts
@@ -55,7 +55,9 @@ import {
 export interface BootResult {
   server: HarnessServer;
   mainWindow: BrowserWindow;
-  /** The tokened local URL the main window loaded (useful for dev smoke tests). */
+  /** Trusted-main-process copy used only by packaged smoke checks. */
+  bootToken: string;
+  /** The UI-credentialed local URL the main window loaded. */
   url: string;
 }
 
@@ -551,14 +553,17 @@ export async function boot(setupWin: BrowserWindow, mode: BootMode): Promise<Boo
   });
 
   // 9. Load the SPA in the main window; close setup once it renders.
-  const url = withDeepLinkParams(`http://127.0.0.1:${server.port}/?token=${bootToken}`, mode.deepLink);
+  const url = withDeepLinkParams(
+    `http://127.0.0.1:${server.port}/?uiToken=${server.uiToken}`,
+    mode.deepLink,
+  );
   const mainWindow = createMainWindow(url);
   mainWindow.webContents.once("did-finish-load", () => {
     if (!setupWin.isDestroyed()) setupWin.close();
   });
   progress(setupWin, { phase: "ready", message: "Ready.", status: "done" });
 
-  return { server, mainWindow, url };
+  return { server, mainWindow, bootToken, url };
 }
 
 /**
--- a/packages/harness-desktop/src/main/index.ts
+++ b/packages/harness-desktop/src/main/index.ts
@@ -179,8 +179,8 @@ if (lock.action === "fail") {
       const coldTarget = coldLink ? parseDeepLink(coldLink) : null;
       bootResult = await boot(setupWin, { devMode, smoke: smokeMode, deepLink: coldTarget ?? undefined });
       if (devMode || smokeMode) {
-        // Dev/smoke hook: print the tokened URL so a harness can verify the
-        // server booted without driving the GUI.
+        // Dev/smoke hook: print the UI-authorized launch URL so a harness can
+        // verify the server booted without driving the GUI.
         console.log(`[harness-desktop] ready: ${bootResult.url}`);
       }
       // BEFORE the smoke branch, deliberately. initUpdater gates itself on
--- a/packages/harness-desktop/src/main/smoke.ts
+++ b/packages/harness-desktop/src/main/smoke.ts
@@ -881,7 +881,7 @@ async function checkUpdateConfig(): Promise<string> {
  */
 export async function runSmokeChecks(boot: BootResult): Promise<SmokeCheck[]> {
   const base = `http://127.0.0.1:${boot.server.port}`;
-  const token = new URL(boot.url).searchParams.get("token");
+  const token = boot.bootToken;
 
   return [
     await check("http-spa", async () => {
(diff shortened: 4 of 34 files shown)


ynadge:
Addressed the durability/contract review on `41c0d40`:

1. **Independent workspace-store race:** first creation now writes a complete temp file and atomically hard-links it into the final name without clobbering. An `EEXIST` loser reads/returns the winner and does not emit `workspace_initialized`. A two-store/different-clock regression proves identical records and exactly one event.
2. **Durable root association:** added boot-token-protected project creation plus allow-listed add-binding and move-binding routes. The durable `projectId` and binding ID choose identity; paths are authorization/binding input only. The server regression moves a root, adds a second repository root, reopens the workspace, restarts the catalog, and retains one project ID/two bindings with path-free responses.
3. **Private repository identity:** removed `repositoryId` from `StudioProjectBindingSummary`, public serialization, browser validation, mocks, and accepted fixtures. The strict parser now rejects that private field.
4. **Same-project target collision:** `moveRootBinding` rejects a target already owned by another binding before write. The regression restarts and resolves the unchanged two-binding catalog afterward.
5. **Shared CLI/Electron catalog writers:** catalog mutations now use a cross-instance file lock and force a fresh read after lock acquisition, preventing whole-array lost updates. An independent-catalog concurrent-writer regression preserves both identities.
6. **Published-package release metadata:** added a minor `@sapiom/harness` changeset documenting `studio-projects.json`, `agent-map/`, and the authenticated endpoints.
7. **Odd legal/unsafe live scopes:** paths with surrounding spaces are retained; an unsafe live scope is skipped instead of poisoning all project reads. Persisted corruption still fails closed.
8. **Previously test-only mutation seams:** `create`, `addRootBinding`, and `moveRootBinding` now have production server callers behind the authenticated boundary.

Fresh evidence: 179 test files / 3,019 tests passed; focused review suite 29/29; harness typecheck, lint, build, and `git diff --check` all exited 0. Build emitted only the existing Vite chunk/dynamic-import warnings.

ynadge:
Resolved the round-two blockers in `70238f6c93186fa82109710adef173a00e671e5f`:

1. **PTY authorization:** `startServer` now generates an ingest-only credential distinct from the host/browser boot token. SessionManager, TaskManager, and `/ingest` share only that credential; `/api` and WebSockets retain the boot token. A real-PTY integration regression captures the injected credential, proves project creation returns 401 with it, proves the boot token is rejected by `/ingest`, and proves the ingest credential still delivers `SessionStart`.
2. **Stale lock ownership:** each catalog lock now carries an owner marker. Stale reclaim and release atomically rename to caller-unique tombstones, revalidate ownership after the rename, restore any newer live lock, and delete only the lock owned by that operation. The deterministic two-catalog regression pauses one waiter after observing an abandoned lock, lets another reclaim/acquire, then proves the delayed waiter restores rather than removes the new lock; both writes survive restart.
3. **Rail preservation:** unsafe/control-character and conflicting-alias live scopes are returned unchanged without `projectId`; only Agent Map assignment is skipped. The regression now asserts that the unsafe scope remains in `workspaceScopes` and has no project identity.

Fresh verification on the pushed SHA:

- Focused regressions: 3 files / 12 tests passed.
- Full harness suite: 180 files / 3,021 tests passed; performance suite: 4 tests passed.
- Harness typecheck, lint, production build, and `git diff --check`: all exited 0. Build emitted only the existing Vite dynamic-import/chunk-size warnings.

ynadge:
Resolved the final re-review findings in `9563d18839e236c67b04a8262b78a5202eb69c78`:

1. **Bootstrap-token exposure:** the server now creates a host/browser-only `uiToken` that is distinct from both the privileged API boot token and the PTY ingest token. A bare request to the origin available in `SAPIOM_HARNESS_INGEST_URL` receives public HTML with no boot/UI credential, and the ingest token still cannot mutate projects. A valid `?uiToken=` launch sets an `HttpOnly; SameSite=Strict` cookie and returns `303` to the same path/query with only `uiToken` removed, so app/PostHog scripts and asset referrers never observe the launch credential. CLI and Electron launch through this flow. Regressions cover the PTY-origin exploit, legitimate launch, cookie reload/deep routes, and preservation of unrelated deep-link params.
2. **Catalog lock serialization/liveness:** the fixed lock is an exclusive `wx` owner record `{ ownerId, pid }`; a live PID is never evicted, regardless of lock age. Dead-owner reclaimers first validate the exact observed owner under a caller-owned claim, atomically move that dead lock, then compete for the fixed name via the same exclusive create as every normal contender. No reclaimer proceeds unless it wins that fixed-name fence. Release verifies exact ownership and unlinks only its own fixed lock. Deterministic regressions cover three competing catalogs recovering a dead owner with all identities retained/no artifacts, plus a deliberately old, slow live owner that cannot be evicted.
3. **Duplicate legacy `WorkspaceKey`:** reconciliation now precomputes key-to-distinct-canonical-root groups before assigning any identity. Every member of a conflicting group remains in `workspaceScopes` unchanged and legacy-only (no `projectId`), and no catalog identity is persisted. The regression runs both input orders and verifies an empty restarted catalog.

Fresh verification on the frozen tree:

- Focused: 4 files, 41 tests passed.
- Full harness: 180 files / 3,024 tests passed; performance suite 3 files / 10 tests passed.
- Live production-wiring E2E: `PASS — live integration proof succeeded`.
- Harness typecheck, ESLint, and production build passed.
- Desktop: 20 files / 169 tests passed; typecheck and build passed.
- Working tree clean; combined `git diff --check origin/main` passed; remote branch resolves to the SHA above.

PR #765, opened 2026-09-01
feat(harness): bootstrap scoped map planner sessions

ynadge:
## Summary

- add server-derived `map-planner` identities and project-scoped resume-or-create/fresh APIs while keeping generic session creation role-free
- append bounded, allowlisted Agent Map context with project/session identity, workspace pointer IDs, binding references, and explicit null/empty revision, proposal, and build-plan slots; never scan or disclose source paths, inventories, secrets, or provider errors
- persist a readiness-gated automatic greeting and FIFO with authoritative copy-on-write queue state, attempt-scoped correlation barriers, dispatch intents, and a content-free accepted-input ledger
- atomically hand a predecessor coordinator directory to the exact rehydrated successor, preserving queued pre-ready input through crashes and repeated replacement while using an older recorded ancestor for the focused continuation brief when needed
- separate the host-only browser API capability from random per-session ingest capabilities; bind ingest authentication to bearer plus harness-session ID and serialize SessionStart identity commits ahead of all following hooks across HTTP and Codex ingress
- durably pin current and historical vendor aliases to one harness session in a private SHA-256 sidecar; allow rotation only from a bounded one-shot trusted `/clear` or `/resume` gesture, including a five-minute hard-capped picker window
- migrate legacy duplicate vendor pointers without blocking boot: the first sessions.json row wins, later pointers are cleared, and the private ledger remains authoritative on every restart
- commit vendor-pointer and adopted-row registry changes copy-on-write behind a registry fence, so concurrent session writes cannot publish a rejected pointer or erase an accepted one; ambiguous owner-ledger writes remain fail-closed
- revalidate live principal and every current active project root at open, route, coordinator, and immediate pre-PTY-write boundaries; clear staged input when authority changes before Enter
- project planner telemetry to bounded content-free fields and a server-pinned/null provider-session envelope; quarantine corrupt queues and classify persistence faults without content or path disclosure
- prevent generic resume, adopt, and input routes from bypassing project-scoped planner ownership; generic adopt returns a bounded 409 for every conflicting current or durable historical alias before adapter probing
- document planner routes, durable state, outbound privacy, authentication, recovery, and migration behavior with race, restart, parser, privacy, and fault-injection regressions

## Dependency

#764 is merged into `main` as `516f13a07a23d82da8cd2ac3ac825bf98526b8f4`; this PR records that squash commit as an ancestor while preserving the reviewed SAP-3055 tree.

## API migration

[...]

PR #766, opened 2026-09-01
feat(harness): open projects in a pinned Agent Map workspace

ynadge:
## Dependencies

SAP-3056 / #764 and SAP-3055 / #765 are merged into `main`. This PR now targets `main` directly and has no remaining stack-order dependency.

## Summary

- add an atomic per-user/project workspace preference store with server-private UUID-to-workflow bindings; public responses expose only opaque agent IDs and safe display metadata
- default a first project visit to Agent Map, restore a valid same-project agent, render transient missing inventory as an in-memory map fallback, and durably repair only a deletion proven by a complete project scan
- preserve opaque agent identity and remembered selection through the authenticated local agent-move route, including null-definition agents and process restart
- render Agent Map as the pinned first project child and as the authoritative plan-first canvas; selecting a child reveals it, while a later explicit project disclosure click may collapse it without changing the workspace selection
- retain the legacy project/path canvas only when the durable Studio project summary is absent; non-map destinations clear the plan-first selection coherently
- persist explicit Agent Map, existing-agent, and newly-created-agent transitions without adding created agents to Agent Map state
- preserve a known agent preference when a PUT races an incomplete scan, apply the server response back to the optimistic client, and strictly reject malformed or over-posted selection requests
- keep the on-disk opaque-agent parser aligned with the strict HTTP boundary by accepting only server-issued UUIDv4 identifiers
- add the harness minor changeset for the preference file, routes, exported types, and project-child layout

## Mainline reconciliation

- merge commit `c729190` joins the prior #766 head with #765's merged mainline commit `ac9a35f`; it preserves both the scoped planner/greeting routes and current-workspace preference routes rather than selecting either conflict side
- server startup initializes both planner orchestration and workspace preferences, and both real and mock clients retain both API surfaces
- workspace preference authorization now reads the live trusted principal, matching planner authorization across sign-in/account changes
- the CI terminology allowance now accounts for both the existing canvas tab and the neutral Agent Map empty-state icon
- browser fixtures now choose their authority explicitly: `mockFixtures=agent-map` and the plan-first deep fixture publish durable Studio summaries, while generic legacy fixtures omit them so unrelated per-agent canvas regressions keep exercising their original surface; production remains summary-driven with no feature flag

## Verification

- `pnpm --filter @sapiom/harness test` — 185 files / 3,127 tests passed; performance suite 3 files / 10 tests passed
- CI-shaped full Chromium Playwright — 478 passed and 7 known flakes passed on retry, exit 0 across all 485 cases
- focused Chromium Playwright for `open-project.spec.ts` and `project-axis.spec.ts` — 28/28 passed, including the exact `mockFixtures=agent-map` entry URL
- focused Studio preference/router/client Vitest — 6 files / 49 tests passed; the final strict persisted-ID regression also passed in its 6-test store suite
- `pnpm --filter @sapiom/harness typecheck` — passed
- `pnpm --filter @sapiom/harness lint` — passed
- `pnpm --filter @sapiom/harness build` — passed; only existing Vite dynamic-import and chunk-size warnings
- `pnpm examples:check` and `pnpm terminology:check` — passed
- `pnpm examples:check:test` — 165 repository-content checks passed
- `pnpm pr-labeler:check` and `pnpm pr-ci-security:check` — passed
- `git diff --check origin/main...HEAD` — passed

## Design contract

The rail and empty Agent Map workspace reuse existing Sapiom components, icons, selection washes, type roles, and tones. The diff adds no CSS, token, tone, or primitive. Detailed Agent Map planning content remains owned by SAP-3058.

Closes SAP-3057

### DIFF
--- a/.changeset/calm-maps-arrive.md
+++ b/.changeset/calm-maps-arrive.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": minor
+---
+
+Add durable, path-free Studio project identities and lazy Agent Map workspace state. The authenticated local server now exposes project workspace and root-binding association endpoints, and stores the new catalog at `studio-projects.json` with per-project records beneath `agent-map/` in the configured harness state root. The legacy System Graph and per-agent Canvas remain unchanged.
--- a/.changeset/steady-agents-map.md
+++ b/.changeset/steady-agents-map.md
@@ -2,4 +2,4 @@
 "@sapiom/harness": minor
 ---
 
-Open durable Studio projects in a pinned Agent Map workspace and remember the selected Agent Map or agent per user and project. Agent identities remain opaque outside the server, survive authenticated moves, and repair safely when a complete project scan proves that an agent was deleted.
+Open durable Studio projects in a pinned Agent Map workspace and remember the selected Agent Map or agent per user and project. Preferences live in the new `agent-map/studio-workspace-preferences.json` state file and are exposed through the path-free `GET` and `PUT /api/projects/:projectId/current-workspace` routes. Agent identities remain opaque outside the server, survive authenticated moves, and repair safely when a complete project scan proves that an agent was deleted.
--- a/.changeset/steady-agents-map.md
+++ b/.changeset/steady-agents-map.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": minor
+---
+
+Open durable Studio projects in a pinned Agent Map workspace and remember the selected Agent Map or agent per user and project. Agent identities remain opaque outside the server, survive authenticated moves, and repair safely when a complete project scan proves that an agent was deleted.
--- a/.changeset/trusted-planners-plan.md
+++ b/.changeset/trusted-planners-plan.md
@@ -1,0 +1,32 @@
+---
+"@sapiom/harness": minor
+---
+
+Add trusted, project-scoped Agent Map planner sessions with deterministic
+resume-or-create/fresh resolution, focused path-free context, a durable
+automatic-greeting state machine, FIFO user input, bounded lifecycle telemetry,
+public planner session/greeting state, and per-session ingest capabilities that
+cannot be replayed across PTYs or used for host `/api` mutations. Rehydrated
+planner replacements atomically inherit the exact predecessor FIFO while their
+focused brief can reuse an older recorded ancestor.
+
+**Breaking:** generic `POST /api/sessions` now strictly rejects unknown fields,
+so clients can no longer attach planner metadata to a generic create request.
+Generic planner input, resume, and adopt routes also reject planner-owned
+sessions. Generic adopt additionally rejects every conflicting current owner
+or durable historical vendor identity, including ordinary pre-`/clear` and
+pre-`/resume` aliases, with a bounded 409 before adapter probing. Migrate
+planner clients to the project-scoped open, message, and greeting-retry routes
+under `/api/projects/:projectId/planner-sessions`.
+
+All coding-agent sessions now pin
+vendor identity to a durable session owner: conflicting `SessionStart` claims
+are rejected. The only rotation exception is a short-lived, one-shot
+server-observed `/clear` or `/resume` terminal gesture; `/resume` picker input
+may refresh its soft window only within a bounded hard deadline.
+
+On upgrade, duplicate legacy vendor resume pointers are migrated with the first
+persisted `sessions.json` row as owner. Later duplicate rows are rewritten with
+`agentSessionId: null` and cannot resume or re-adopt that fenced identity under
+the losing row. Provider transcripts and conversation history are not deleted;
+start a fresh session in the losing row's directory to continue there.
--- a/packages/harness-desktop/src/main/boot.ts
+++ b/packages/harness-desktop/src/main/boot.ts
@@ -55,7 +55,9 @@ import {
 export interface BootResult {
   server: HarnessServer;
   mainWindow: BrowserWindow;
-  /** The tokened local URL the main window loaded (useful for dev smoke tests). */
+  /** Trusted-main-process copy used only by packaged smoke checks. */
+  bootToken: string;
+  /** The UI-credentialed local URL the main window loaded. */
   url: string;
 }
 
@@ -551,14 +553,17 @@ export async function boot(setupWin: BrowserWindow, mode: BootMode): Promise<Boo
   });
 
   // 9. Load the SPA in the main window; close setup once it renders.
-  const url = withDeepLinkParams(`http://127.0.0.1:${server.port}/?token=${bootToken}`, mode.deepLink);
+  const url = withDeepLinkParams(
+    `http://127.0.0.1:${server.port}/?uiToken=${server.uiToken}`,
+    mode.deepLink,
+  );
   const mainWindow = createMainWindow(url);
   mainWindow.webContents.once("did-finish-load", () => {
     if (!setupWin.isDestroyed()) setupWin.close();
   });
   progress(setupWin, { phase: "ready", message: "Ready.", status: "done" });
 
-  return { server, mainWindow, url };
+  return { server, mainWindow, bootToken, url };
 }
 
 /**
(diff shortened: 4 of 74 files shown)


ynadge:
Resolved the blocking review findings in `0563578` (with the SAP-3056 foundation merged normally at `9563d18`):

1. **Real AppState bindings:** `annotateStudioSelections` now derives project roots from the reconciled `workspaceScopes`, not the original scopes that lack `projectId`. `studio-workspace-wiring.test.ts` starts the real server and proves `/api/state` publishes the opaque binding without exposing its private root.

2. **Move-stable opaque identity:** the authenticated `/api/agents/move` completion seam now atomically remaps server-private binding evidence before the registry rescans. The real-server test moves a null-definition workflow, restarts the process, and proves both `agentId` and remembered agent selection survive.

3. **Nested project restore:** path ownership now selects the most-specific containing durable Studio scope. The pure parent+nested regression proves the nested project wins regardless of scope order.

4. **Stable selection is the plan-first authority:** the rail, altitude, focused workflow, canvas shell, and conversation subject now derive from `StudioWorkspaceSelection`. Agent Map renders the agreed existing `EmptyState`; legacy `ProjectRef`/system-graph rendering remains only for projects without a durable Studio summary. Explicit agent opens update the stable selection, and Templates/New Session/Overview/review destinations clear it. Focused Playwright covers the empty Agent Map, explicit agent selection, and non-map transition.

5. **Transient absence is non-destructive:** preference reconciliation now receives registry scan completeness. Empty/degraded inventory returns an in-memory map with `repaired: false`, retains the durable agent preference/binding, and the client never PUTs a guessed map. Only a complete scan can prune the binding and persist a deletion repair. Store coverage proves transient absence restores the original agent after restart, while a proven deletion repairs durably.

6. **Open-project failure isolation:** the core open/remember/scan operation still determines success. The added state/map/preference fetches are best-effort; failures are caught and any restore marker already installed is removed so a later restore can retry. No dedicated failure-injection E2E was added; the existing open-project browser cases all pass.

7. **Disclosure remains operable:** selecting a child reveals its project synchronously. A later explicit label/disclosure click can collapse the selected project without clearing its workspace, and same-logical-selection async responses cannot reopen it. The selected collapse/expand regression passed 5 consecutive repetitions and the full focused set passed three repetitions.

8. **Published contract changeset:** `.changeset/steady-agents-map.md` adds a minor `@sapiom/harness` changeset covering the preference file, current-workspace routes/types, opaque IDs, and plan-first layout.

9. **Safe pruning / checkout identity:** complete scans prune stale project bindings. Definition-ID fallback was removed so another checkout can never inherit a deleted path identity; an authenticated move is the only path-changing preservation seam. Tests cover deletion and a duplicate-definition checkout receiving a different UUID.

10. **Creation and ambiguous focus:** scaffold completion is followed by fresh state resolution and an awaited preference PUT, so a newly-created agent selection persists after its completed rescan. Multi-binding focus chooses the most-specific owning project first, then explicit/current context, then a deterministic binding; it no longer silently clears selection while leaving a stale preference.

11. **Hot-read nit:** I did not add a second cache/feature-flag layer in this correctness pass. Reconciliation remains serialized and persistence is conditional on an actual catalog/binding change. This is the one acknowledged performance follow-up, not a correctness gap in the acceptance behavior.

12. **Final evidence on the merged tree:** focused Vitest `63/63`; focused Chromium Playwright `24/24` with `--repeat-each=3`; harness typecheck, lint, and production build all passed; `git diff --check 9563d18...HEAD` passed; and the diff contains no CSS/design-token/tone/primitive additions.

ynadge:
Resolved the round-2 blocker and associated contract nits in `5dabe08`:

1. **Degraded PUT no longer destroys a valid selection.** `put()` now distinguishes a complete absence from an incomplete scan. During an incomplete scan it accepts a requested agent that remains in the server-private project binding table; an unknown requested ID leaves the prior durable preference untouched. Neither case is misreported as a repair. Only a complete scan (or a definitively foreign project) may downgrade and persist Agent Map.

2. **Restart regression added.** The new store case begins with a valid selected agent, PUTs that agent while inventory is empty/degraded, restarts the store, verifies the transient read is an in-memory map with `repaired: false`, then restores complete inventory and verifies the same opaque agent selection returns. Existing complete-scan deletion and foreign-ID repair coverage remains green.

3. **Client/disk agreement.** Explicit agent selection remains optimistic, but now applies the authoritative PUT response with the existing generation guard. A complete-scan repair therefore cannot leave the visible selection disagreeing with the persisted result, and a stale response cannot override a newer navigation.

4. **Strict request boundary.** I investigated the recorded contract and made the parser strict: the top-level request and both discriminated selection variants reject extra keys, and project/agent IDs must have the server-issued UUID shape. Route coverage proves malformed agent IDs, variant over-posting, and top-level private-field over-posting return the bounded `malformed_state` response.

5. **Nits addressed.** The scan-completeness calculation now has one shared server helper. The minor changeset now names `agent-map/studio-workspace-preferences.json` and the path-free `GET`/`PUT /api/projects/:projectId/current-workspace` routes. The previously acknowledged hot-read caching follow-up remains intentionally deferred.

6. **Fresh evidence at `5dabe08`:** focused Vitest 6 files / 65 tests passed; focused Chromium Playwright 24/24 with `--repeat-each=3`; harness typecheck, lint, and production build passed; `git diff --check 9563d18...HEAD` passed; no CSS/design-token/tone/primitive additions.

ynadge:
Reconciled #766 with the merged SAP-3055 mainline and resolved the failing gate at final head `76e1640`:

1. **Conflict resolution was a union, not a side selection.** The four conflicts in `server/agent-map.ts`, `server/index.ts`, `web/src/lib/api.ts`, and the router tests retain both the scoped planner/greeting API from #765 and the durable current-workspace preference API from #766.

2. **Live principal integration is preserved.** Current-workspace GET/PUT authorization now resolves the trusted principal at request time, matching the planner service across account changes. The router regression changes the principal without restarting and proves the two users remain isolated.

3. **The `examples` gate failure is fixed.** The failure was the exact Agent Studio terminology canary `app-canvas-tab-icon-identifier` changing from one to two visible occurrences after the Agent Map empty state landed. The allowlist now documents both intentional icon identifiers. `pnpm terminology:check`, `pnpm examples:check`, and the 165 content-check fixtures pass.

4. **Combined-tree browser regressions are aligned with the intended authority model.** The legacy System Graph test now models a server with no durable Studio project summaries, while normal fixtures use the plan-first path. The project-sort selector targets project rows rather than the newly inserted Agent Map child. The affected Chromium suites pass 28/28 serially.

5. **Advisory parser finding addressed.** Persisted opaque agent IDs now use the same UUIDv4 shape as the strict HTTP boundary, with a regression proving a previously accepted non-v4 disk value fails as `malformed_state`.

6. **Cross-instance preference locking suggestion is acknowledged, not represented as fixed here.** This store remains atomically persisted and serialized within its server process. Sharing a live preference file between concurrent CLI/Electron server processes would require the same owner-fenced lock protocol as the project catalog; that is a distinct durability change and should not be improvised inside this merge/CI reconciliation.

7. **Fresh final evidence:** full harness 185 files / 3,127 tests plus 10 performance tests; typecheck, lint, production build, exact examples/terminology checks, PR label/security checks, and `git diff --check origin/main...HEAD` all pass. The worktree and remote head match; the PR remains open and unmerged.

ynadge:
Resolved the post-merge `playwright-mock` failure at `a59c336` without changing deployed Studio behavior:

1. **Root cause:** after #765 landed, the generic mock response exposed durable Studio project summaries to every historical browser fixture. The application correctly selected Agent Map on first open, but hundreds of legacy canvas/direct-action tests were still designed to begin in an agent workspace. Their repeated failures and retries exhausted the 15-minute CI job.

2. **Mock authority is now explicit.** `mockFixtures=agent-map` and the plan-first deep scenario publish durable Studio project summaries. Generic historical fixtures omit those summaries and continue to exercise the legacy per-agent/System Graph surface they were written for. This is mock-data selection only: production remains server-summary-driven, always plan-first when a durable Studio project exists, with no feature flag.

3. **The demo entry is covered directly.** `open-project.spec.ts` now runs through the exact `?mockFixtures=agent-map` URL. The deep rail suite opts into Studio summaries except for its explicit legacy containment case.

4. **Fresh verification on the pushed head:** CI-shaped full Chromium Playwright completed in 6.9 minutes with 478 passed and 7 known flakes passing on retry (all 485 cases, exit 0); the focused Agent Map suites passed 28/28; full harness Vitest passed 185 files / 3,127 tests plus 10 performance tests; typecheck, lint, production build, Prettier, and diff checks all passed.

5. **GitHub result:** every current-head check is green. `playwright-mock` passed in 12m38s; CodeQL, both Node versions, examples, desktop smoke, label/security checks, and review also passed.

The PR remains open, unmerged, and mergeable.

PR #770, opened 2026-09-02
docs(content-generation): document images.launch in the package README [SAP-3098]

gwitwer:
## Primary change type

- [x] Documentation

## Problem and motivation

`images.launch` shipped in `@sapiom/tools` 0.25.0 and has been effectively undiscoverable since: it appears only in JSDoc and the changelog entry. `packages/tools/src/content-generation/README.md` documents the video equivalent at length — a "Dispatchable video: `video.launch`" section plus `VIDEO_RESULT_SIGNAL` and `VideoResultPayload` — and never mentions `images.launch` or `IMAGE_RESULT_SIGNAL`. `grep -n "launch" README.md` on `main` returns only video lines.

The async image path is a real capability: the submit returns a handle immediately, it supports `wait()` and workflow signal-resume, and three shipped examples (`personalized-media-at-scale`, `scene-to-video`, `research-to-microsite`) already depend on it.

## Summary and scope

Adds a **"Dispatchable image: `images.launch`"** section, placed with the rest of the image content and mirroring the video section's structure:

- **When to prefer `launch` over `create`** — long generations (the routed sync `create` holds the request open for the whole generate+store and so is bounded by Core's 30s router cap, the failure mode a `create` fan-out hits), and workflow steps that should suspend rather than hold a socket.
- **The handle shape** — `{ requestId, resolvedModel, cost?, preferSatisfied?, dispatch, wait() }`, with `requestId` doubling as the `dispatch.correlationId` a workflow resumes on, and the submit-resolved metadata readable off the handle without awaiting.
- **`wait()`** — `{ timeoutMs, pollMs }`, 2min / 2s. Called out explicitly that, unlike video, `ImageCreateInput` has no `timeoutMs` / `pollIntervalMs` fields, so the override goes on the `wait()` call.
- **`IMAGE_RESULT_SIGNAL`** — the `pause: { signal }` declaration, the constant's value, and the fact that images resume through the same media-agnostic rail as video (signal name off the paused step row, matched on `correlationId`).
- **`ImageResultPayload`** — the full interface, one entry per generated image, plus `toImageResumePayload` for local tests.
- Cross-links both dispatchable surfaces from the intro, so the section is reachable from the top of the file.

Also documents current `wait()` failure semantics honestly: a non-OK poll is treated as "still generating", so a terminal provider failure surfaces as the deadline `Error`, not a failure-specific one. That is [SAP-3097](https://linear.app/sapiom/issue/SAP-3097), which has not landed; when it does, this paragraph is the one line to update.

One accuracy fix in the adjacent video section falls out of writing the image one: `pauseUntilSignal`'s `resumeStep` is typed `const Resume extends string`, so the snippet's bare `finalize` identifier is corrected to `"finalize"`.

**Out of scope:** no code change; the `examples/` READMEs and the remote-MCP `async` surface are untouched.

## Related work

Related issue or discussion: [SAP-3098](https://linear.app/sapiom/issue/SAP-3098/sdk-document-imageslaunch-in-the-content-generation-readme)

## Validation

```text
npx prettier --check packages/tools/src/content-generation/README.md — All matched files use Prettier code style!
node scripts/provider-neutral-copy-check.mjs — passed (124 audited files)
```

Every documented symbol, default, and type was read off `packages/tools/src/content-generation/index.ts` on `main` rather than paraphrased from the ticket:

| Claim | Source |
| --- | --- |
| `IMAGE_RESULT_SIGNAL === "contentGeneration.images.result"` | `index.ts:48` |
| `wait()` defaults 2min / 2s | `DEFAULT_IMAGE_TIMEOUT_MS`, `DEFAULT_IMAGE_POLL_INTERVAL_MS` (`index.ts:577-578`) |
| Handle members | `ImageLaunchHandle` + the `launchImage` return |
| `wait()` merges submit metadata | `withDispatchMetadata(mapResult(raw), handle)` |
| Resume payload fields | `ImageResultPayload` / `toImageResumePayload` |
| Failure indistinguishable from timeout | the non-OK branch of `wait()`'s poll loop |

Build / typecheck / lint / tests were not run: the diff is a single Markdown file inside a package, so it cannot affect them. Prettier is the check that actually applies, and it passes.

### Tests and documentation

N/A for tests — documentation-only change with no executable surface. Documentation is the change itself.

- Breaking or externally visible changes: None. Markdown only.
- Changeset: N/A — docs-only, following the precedent of `docs(authoring): replace workflow copy with agent` (#530), which changed package READMEs without one.

## Security

- [x] This pull request does not publicly disclose a suspected vulnerability.

## AI assistance

Written with Claude Code. The prose and examples were drafted by the model after reading `packages/tools/src/content-generation/index.ts`, `src/stub/index.ts`, and the three `examples/` agents that call `images.launch`; every factual claim was checked back against the implementation (see the table above) rather than against the ticket text.

## Checklist

https://claude.ai/code/session_01GPphPpsV1He677J3MaT37G

### DIFF
--- a/.changeset/document-images-launch.md
+++ b/.changeset/document-images-launch.md
@@ -2,4 +2,4 @@
 "@sapiom/tools": patch
 ---
 
-Document `images.launch` in the content-generation README: the launch handle, `wait()` and its 2min/2s defaults, `IMAGE_RESULT_SIGNAL`, `ImageResultPayload`, and when to prefer `launch` over `create`. The async image path had shipped in 0.25.0 with no README coverage. Also corrects the pause/resume snippets in both the image and video sections — `defineStep`'s `pause` requires `signal` and `resumeStep` and belongs on the pausing step, and the pause/resume helpers are imported from `@sapiom/agent`. Docs only; `src/**/README.md` ships in the package tarball, so this reaches npm as a patch.
+Document `images.launch` in the content-generation README: the launch handle, `wait()` and its 2min/2s defaults, `IMAGE_RESULT_SIGNAL`, `ImageResultPayload`, and when to prefer `launch` over `create`. The async image path shipped in 0.25.0 with no README coverage. Also corrects the pause/resume snippets in both the image and video sections — `defineStep`'s `pause` requires `signal` and `resumeStep` and belongs on the pausing step, and the pause/resume helpers are imported from `@sapiom/agent`.
--- a/.changeset/document-images-launch.md
+++ b/.changeset/document-images-launch.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/tools": patch
+---
+
+Document `images.launch` in the content-generation README: the launch handle, `wait()` and its 2min/2s defaults, `IMAGE_RESULT_SIGNAL`, `ImageResultPayload`, and when to prefer `launch` over `create`. The async image path had shipped in 0.25.0 with no README coverage. Also corrects the pause/resume snippets in both the image and video sections — `defineStep`'s `pause` requires `signal` and `resumeStep` and belongs on the pausing step, and the pause/resume helpers are imported from `@sapiom/agent`. Docs only; `src/**/README.md` ships in the package tarball, so this reaches npm as a patch.
--- a/packages/tools/src/content-generation/README.md
+++ b/packages/tools/src/content-generation/README.md
@@ -82,6 +82,9 @@ const out = await handle.wait(); // polls until ready
 out.images?.[0]?.fileId; // + .downloadUrl for a ready-to-use URL
 ```
 
+To suspend an agent step until the image is ready instead, pause on the handle —
+see [`IMAGE_RESULT_SIGNAL`](#image_result_signal) below.
+
 The handle is `{ requestId, resolvedModel, cost?, preferSatisfied?, dispatch,
 wait() }`. `requestId` is the queue request id, and it is also the
 `dispatch.correlationId` a workflow resumes on. `resolvedModel`, `cost`, and
@@ -117,7 +120,7 @@ const renderImage = defineStep({
   async run(_input: unknown, ctx) {
     const handle = await ctx.sapiom.contentGeneration.images.launch({
       prompt: "a red bicycle",
-      storage: { visibility: "public" },
+      storage: { visibility: "private" },
     });
     return pauseUntilSignal(handle, { resumeStep: "collectImage" });
   },
--- a/packages/tools/src/content-generation/README.md
+++ b/packages/tools/src/content-generation/README.md
@@ -51,8 +51,8 @@ Persisting is best-effort per image: if one fails, that image carries a
 ## Dispatchable image: `images.launch`
 
 `images.launch` is the dispatchable surface for image generation. It submits the
-job and returns a handle immediately (~1–2s), so you decide when — or whether —
-to wait for the result.
+job and returns a handle as soon as the job is enqueued, so you decide when — or
+whether — to wait for the result.
 
 Prefer `launch` over `create` when:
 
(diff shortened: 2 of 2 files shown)


gwitwer:
## Review round 1 — all four accepted, plus one the review didn't reach

Pushed in 86785f9b. Every point checked out against the shipped types; nothing here was a nitpick.

### 1. `pause` snippet — confirmed, and there was a second defect underneath

Both halves are right. `defineStep`'s `const PauseDecl extends { signal: string; resumeStep: string } | undefined` (`packages/agent/src/step.ts:113`) requires both keys, and `pause` feeds `Allowed<Next, Term, CanFail, PauseDecl["resumeStep"]>`, so it gates the *pausing* step's return type and belongs there. I reproduced the pre-fix form:

```
negative-check.ts(7,3): error TS2741: Property 'resumeStep' is missing in type '{ signal: string; }'
  but required in type '{ signal: string; resumeStep: string; }'.
negative-check.ts(9,9): error TS2322: Type '... => Promise<void>' is not assignable to
  '... => Promise<Allowed<readonly [], true, false, string>>'.
```

Both snippets now follow the shape the shipped `examples/personalized-media-at-scale` agent uses — the pausing step declares `pause: { signal, resumeStep }` and returns `pauseUntilSignal(...)`; the resume target declares no `pause` and takes the payload as its input. Fixed in the video section in the same pass, as you asked.

**The thing your report didn't reach:** that second TS2322 above is a *separate* defect that survives the `resumeStep` fix. A `terminal: true` step's `run` must return `Allowed<..., true, ...>`, so the resume target can't fall off the end — `result.outputs[0]?.fileId;` as a bare expression statement doesn't compile even with the `pause` edge corrected. It returns `terminate({ fileId: ... })` now. I only found this because I compiled the snippets rather than reasoning about them; worth noting since the same expression-statement-as-last-line shape appears in other package READMEs.

### 2. Failure semantics — correct, overreach on my part

You're right that I imported video's guarantee. `VIDEO_RESULT_SIGNAL`'s JSDoc says "ready OR failed … the resumed step branches" (`index.ts:34-36`); `IMAGE_RESULT_SIGNAL`'s (`index.ts:41-46`) deliberately says only "terminal state". And `ImageResultPayload` has no status or error field, so "branch on it" pointed at nothing. Dropped the parenthetical; the sentence now matches the image JSDoc. The `wait()`-path paragraph (SAP-3097) is unchanged — that one is sourced from the poll loop.

### 3. Wrong import package — correct

`defineStep` / `pauseUntilSignal` / `terminate` are `@sapiom/agent` exports (`packages/agent/src/index.ts:26,31`). For a section whose entire purpose is discoverability, sending readers to the wrong package was the worst of the four. Snippets now import from both packages, with a closing line naming which symbol comes from where. Also split the duplicate-`const handle` block so each fence pastes as a file — including the pre-existing video Option A/B block, since leaving it inconsistent next to a corrected one seemed worse than the small extra diff.

### 4. Changeset — correct, my precedent citation was weak

`packages/tools/package.json:94` does list `src/**/README.md` in `files`. My #530 citation didn't distinguish a shipped doc from a repo-only one, which is exactly your point. Patch changeset added for `@sapiom/tools`.

### Nit — taken

"~1–2s" came from the ticket description, not from anything in the SDK. Dropped; it now says the handle returns as soon as the job is enqueued, which is what `launchImage` actually guarantees.

### Validation

Each snippet was transcribed verbatim into a scratch package (only the import specifiers rewritten to resolve in-repo) and compiled against `packages/agent` + `packages/tools`:

```text
npx tsc --project .snippet-check/tsconfig.json — clean (4 snippets: 2 pause/resume, 2 inline wait())
npx prettier --check packages/tools/src/content-generation/README.md .changeset/document-images-launch.md — passes
node scripts/provider-neutral-copy-check.mjs — passed (125 audited files)
```

Scratch package removed before commit; `git status` is clean apart from the two intended files.

https://claude.ai/code/session_01GPphPpsV1He677J3MaT37G

gwitwer:
## Round 2 — all three nits fixed

Pushed in 250fa420. All three were real and in scope (each concerns text this PR wrote or moved), so none is disputed.

**Changeset carried reviewer-facing meta into the public changelog.** Correct, and I should have caught it — a changeset body *is* the `CHANGELOG.md` entry. "Docs only; `src/**/README.md` ships in the package tarball, so this reaches npm as a patch" was my justification to you for finding #4, which explains nothing to someone reading the release notes. Trimmed to what a consumer needs: what got documented and what got corrected.

**`visibility: "public"` in the pause snippet.** Also correct. I lifted that block's shape from `examples/personalized-media-at-scale`, which sets `"public"` deliberately — it emails a durable permalink that has to outlive a presigned URL's TTL, and says so in a comment. Neither the reason nor the comment came across with the snippet, leaving a paste-and-go example that publishes generated media for no stated reason. Now `"private"`, matching every other `launch` snippet in the file.

**Missing pointer in the image section.** Fixed — the image section now carries the same "to suspend an agent step instead, see [`IMAGE_RESULT_SIGNAL`](#image_result_signal) below" line the video section got when its inline and pause examples were split. That asymmetry was introduced by the round-1 fix, so it belongs here.

### Validation

Rather than re-checking the snippets I'd already transcribed, I re-extracted all four ` ```typescript ` blocks programmatically from the README (only `@sapiom/agent` rewritten to the in-repo path) and compiled those:

```text
tsc --project .snippet-check/tsconfig.json — clean, 4/4 blocks extracted from the README
prettier --check <README> <changeset>      — passes
provider-neutral-copy-check                — passed (125 audited files)
internal anchor check                      — 7 links, 0 broken
```

One note on that run, since it produced a red herring worth recording: under the package's own CJS `tsconfig`, the two inline-`wait()` blocks fail with `TS1378` (top-level `await`). That is a property of my harness, not the snippets — the blocks use the same top-level-await style as the README's existing intro example and are correct in the ESM context a consumer writes them in. Re-running with `module: esnext` / `target: es2022` is clean, and that is the result above.

Scratch package removed; the branch holds only the two intended files. CI green (`test 20.x`, `test 22.x`, `examples`, `CodeQL`, `desktop-smoke`).

https://claude.ai/code/session_01GPphPpsV1He677J3MaT37G

PR #771, opened 2026-09-02
fix(tools): tell a failed async media generation from a slow one [SAP-3097]

gwitwer:
## Problem

A caller who launched an async image or video generation could not tell a **failed** job from a **slow** one, on either of the two paths that report one.

**1. `wait()` polled to the deadline on a terminal failure.** Every non-OK poll response was treated as "still generating", so a job that terminally failed in three seconds burned the caller's whole `timeoutMs` and then threw `Image generation did not complete within 300000ms` — which tells the caller the opposite of what happened.

**2. The resume payload had no generation-failure channel.** The provider error already reached a resumed workflow step, but on `storageError` — the field documented as *"Present when storage was requested but persisting this output failed."* A step branching on that type concluded persistence broke, not that the asset was never generated.

### `wait()` fails fast, with the provider's message

New `ContentGenerationFailedError` (exported from `@sapiom/tools`), carrying `requestId` and the provider's own `providerError`. A plain timeout `Error` now means only what it says: still running when we stopped waiting.

The poll loop moves to `content-generation/poll.ts` and is shared by `images.launch().wait()`, `video.launch().wait()`, and `video.create()`, so their terminal semantics cannot drift apart.

### A transport blip still keeps polling

Two channels report terminal failure, and both are read:

- **The polled body.** `terminalFailureFrom` classifies `status` plus `error` / `error_type`, mirroring the gateway's own classification of the same wire contract so the two agree. It is deliberately conservative — only an explicit terminal marker ends the poll, so an unfamiliar body never gets reported as a generation failure.
- **The status endpoint**, consulted when the result endpoint answers non-OK. That answer is ambiguous on its own (failed, not-done-yet, and blipped all look alike) and the status endpoint reports terminal state unambiguously. It is a tie-break, not a second poll: some queues report "not ready yet" as a non-OK result response, so it runs on the first non-OK and every 4th after (`STATUS_PROBE_EVERY`), which still catches an immediate failure, bounds detection to four poll intervals, and holds the extra load to ~25%.

The failure marker is read **before** the result predicate. A terminal body can still carry an empty container for the asset it never produced — `{ status: "COMPLETED", error: "…", images: [] }` — and reporting that as a successful empty result is the exact ambiguity this change removes. Conversely, only real error *content* makes a `COMPLETED` job terminal: a present-but-empty `error: ""` does not fail a job that is handing back its asset.

### `generationError` on the resume payload

`ImageResultPayload` / `VideoResultPayload` outputs gain `generationError?: string`. The platform sends it *instead of* the storage fields, never alongside, so a resumed step branches without string-matching a message:

```typescript
const out = result.outputs[0];
if (out?.generationError) throw new Error(`generation failed: ${out.generationError}`);
if (out?.storageError) throw new Error(`could not persist the output: ${out.storageError}`);
```

Nothing in this package produces `generationError` — it appears on the wire once the gateway change that emits it is deployed, and a step running against an older gateway still sees a generation failure on `storageError`. Checking `generationError` first and falling back, as above, reads correctly on both sides of that deploy. The changeset, README, and field JSDoc all say so, so the published CHANGELOG doesn't claim a fix that hasn't shipped.

This is also what makes `VIDEO_RESULT_SIGNAL`'s documented *"carries the result either way (ready OR failed)"* claim expressible — the payload previously had nowhere to put the failure. `IMAGE_RESULT_SIGNAL` now documents the same contract.

Three templates that read the resume payload (`scene-to-video`, `content-repurposing-pipeline`, `research-to-microsite`) are updated to report the right failure.

## Depends on

sapiom/Sapiom#4801 — the gateway half, which emits `generationError` on the wire. This PR's type is inert until that ships; nothing regresses in the meantime.

## Testing

`npx jest --maxWorkers=1 src/content-generation/index.spec.ts` — 143 pass (31 new). Coverage per the ticket's acceptance criteria:

- terminal failure on the polled body (`ERROR`, `COMPLETED`-with-error, `CANCELLED`), asserted to give up on the **first** poll rather than run to `timeoutMs`
- terminal failure discovered via the status endpoint after a non-OK result poll
- transient non-OK (502 / 503) followed by success — keeps polling, resolves, exactly two result polls
- a terminal body carrying an empty `images: []` throws; a non-terminal one still resolves
- a `COMPLETED` job carrying `error: ""` alongside its asset resolves
- probe cadence: 8 non-OK polls then the result gives exactly 9 polls and 2 probes (scripted counts, no wall-clock budget)
- clean success, which never touches the status endpoint
- a job that never finishes still throws the timeout `Error`, not `ContentGenerationFailedError`
- a table over `terminalFailureFrom` for every keep-polling and every terminal shape

`pnpm typecheck` and `pnpm lint` clean in `packages/tools`.

Closes SAP-3097

https://claude.ai/code/session_015XHWBRT86s84urfsTY9JJe


gwitwer:
Four findings, three fixed in 6101551, one I'm pushing back on.

### ✅ `images: []` slips past the terminal-failure check — real bug, fixed

Correct, and it was self-inflicted: I added the "an asset beside a failure marker wins" shortcut late, which put the result predicate ahead of the terminal check. An empty array is not an asset, so the ordering is now terminal-check-first. The carve-out is gone rather than gated, because the shape it defended against (a completed-with-error body that also carries real media) doesn't occur.

I did **not** take the suggested `raw.images.length > 0`. That would change the completion predicate itself, so a non-terminal `{ images: [] }` — which resolves today and did before this PR — would start polling to `timeoutMs`. Trading an empty result for a five-minute hang is a worse answer to a shape that has no failure marker on it. Two tests now pin both halves: a terminal `{ status: "COMPLETED", error, images: [] }` throws, a bare `{ images: [] }` still resolves.

### ✅ Status probe on every non-OK poll — real cost, fixed

Your arithmetic is right and I'd underweighted it. Probing is a tie-break, not a second poll. It now runs on the first non-OK and every 4th after (`STATUS_PROBE_EVERY`): a job that fails immediately is still caught on the first tick, detection is bounded to four poll intervals, and the extra load is ~25% rather than 100%. Test asserts `probes === ceil(polls / 4)` over a run that stays non-OK throughout.

### ✅ Internal gateway symbol in the npm tarball — fixed

Correct that `declaration: true` + `files: ["dist"]` ships this JSDoc verbatim. The private symbol name and the settlement-path sentence are gone; the docs now describe the role ("the gateway's own classification of the same wire contract") without naming internals.

### ✅ Published copy claims the payload half as done — fixed

Fair. The changeset and README stated a fix that depends on a deploy this package doesn't control, and a CHANGELOG can't be corrected after publish. Both now separate the two halves explicitly: `wait()` is live in this release; `generationError` is a type the platform populates once the gateway change deploys, and until then a generation failure still arrives on `storageError`. Both docs point at the `generationError` first, `storageError` fallback, which reads correctly on either side of the deploy. The field JSDoc is softened the same way — it describes what the platform sends rather than asserting an invariant this package enforces.

### ❌ `toImageResumePayload` / `toVideoResumePayload` don't map `generationError`

Not fixing this — they structurally cannot, and adding it would require inventing a field that can't be populated.

Those mappers take a **live** `ImageGenerationResult` / `VideoGenerationResult`. After this PR a live terminal failure doesn't *produce* one of those — `wait()` throws `ContentGenerationFailedError` instead. So the input type has no failure to map from, and giving `GeneratedImage` a `generationError` field would add a property nothing ever sets, on the live path, purely so a mapper could forward it.

That asymmetry is the design: the live path throws, the durable resume path delivers a payload. A test that wants the failure payload writes it literally — it's `{ outputs: [{ generationError: "…" }] }`. I've added that note to the README beside the mapper recommendation so the dead end is signposted rather than discovered.

### Verified

`npx jest --maxWorkers=1 src/content-generation/index.spec.ts` — 138 pass (26 new). `pnpm typecheck` and `pnpm lint` clean in `packages/tools`.

gwitwer:
Both round-2 findings are real. Fixed in 070bcbc.

### ✅ Empty `error` string fails a successful job

Correct, and it's a regression I introduced with the reorder — you're right that the success path was masking it before. `hasQueueError` tested key *presence* (`!= null`), which `""` satisfies.

The asymmetry is the fix: `FAILED` / `ERROR` / `CANCELLED` are terminal on the status alone, so their "job reported `<status>`" fallback stays. `COMPLETED` is terminal-**success** by default — the error content is the only thing that turns it into a failure — so that content has to be real. The branch now gates on `providerErrorMessage`, which was already exactly the non-empty-trimmed test, and `hasQueueError` is gone. That also makes the docblock's "only an explicit terminal marker" promise true, which it wasn't.

Knock-on: `{ status: "COMPLETED", error: {} }` now keeps polling instead of throwing "job completed with an unnamed error". That fallback is unreachable by construction and removed — a contentless error object is not an explicit marker, so the conservative default applies. Four table cases pin the shapes (`""`, `error_type: ""`, whitespace-only, `{}`), plus an end-to-end one asserting `{ status: "COMPLETED", error: "", images: [{…}] }` resolves with the asset.

### ✅ PR body records the wrong rule

Fixed. The description said "a body that actually carries the asset wins over any failure marker beside it", which 61015517 reversed. The whole body is rewritten to state the rule the code implements — failure marker first, with the empty-container reasoning and the `error: ""` carve-out — so the merge commit doesn't preserve the stale version. I also brought the status-endpoint bullet in line with the throttling and dropped the internal symbol name there for the same reason it left `poll.ts`.

### ✅ Cadence test is wall-clock dependent

Fair — asserting `polls > 4` inside a 30 ms budget against real timers is a flake waiting for a loaded runner. Rather than reach for fake timers (the loop awaits real fetches between sleeps, so the plumbing outweighs the payoff here), it now scripts the sequence: 8 non-OK polls then the result. That fixes the counts exactly — 9 polls, 2 probes — with no time budget in the assertion at all. It also reads better, since it now demonstrates the intended end state (job completes) rather than a timeout.

### Verified

`npx jest --maxWorkers=1 src/content-generation/index.spec.ts` — 143 pass (5 new this round). `pnpm typecheck` and `pnpm lint` clean.

PR #778, opened 2026-09-02
fix(harness): refresh MCP credentials at launch

ynadge:
## Summary

Agent Studio now resolves the latest saved credential at the shared Claude launch boundary, so created, resumed, and background sessions cannot keep using a boot-time key. Remote MCP URLs also follow the selected environment, while `--no-auth` remains a hard process-wide opt-out.

## Changes

- Add an additive strict MCP credential reader and serialized provider reconciliation.
- Refresh immediately before each default session/task MCP config is generated.
- Route the remote MCP URL through the resolved production, staging, alias, or custom environment.
- Prevent shared credential adoption and auth mutations when Studio runs with `--no-auth`.
- Add unit and real `startServer()` lifecycle coverage, plus minor changesets for MCP and Harness.

## Testing

- [x] `pnpm --filter @sapiom/mcp test` — 146 tests
- [x] `pnpm --filter @sapiom/harness test` — 3,157 unit/integration and 10 performance tests
- [x] `pnpm --filter @sapiom/mcp typecheck`
- [x] `pnpm --filter "@sapiom/harness..." build`
- [x] MCP and Harness lint
- [x] New lifecycle coverage uses fake credentials and mocked OAuth; it does not open a browser or contact production

## Boundaries

Already-running session reconnection remains SAP-3116. Unified verified auth state and observability remain SAP-3115. Codex MCP injection remains out of scope because its adapter does not consume the generated Claude MCP configuration.

## Related

Closes: SAP-3114

### DIFF
--- a/.changeset/live-studios-refresh.md
+++ b/.changeset/live-studios-refresh.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/mcp": patch
+"@sapiom/harness": patch
+---
+
+Resolve the latest saved credential and selected environment whenever Agent Studio creates, resumes, or runs a background Claude session. Preserve `--no-auth` as a process-wide opt-out and expose a strict credential-store reader for safe live reconciliation.
--- a/.changeset/live-studios-refresh.md
+++ b/.changeset/live-studios-refresh.md
@@ -3,4 +3,4 @@
 "@sapiom/harness": patch
 ---
 
-Resolve the latest saved credential and selected environment whenever Agent Studio creates, resumes, or runs a background Claude session. Preserve `--no-auth` as a process-wide opt-out and expose a strict credential-store reader for safe live reconciliation.
+Resolve the latest saved credential and selected environment whenever Agent Studio creates, resumes, or runs a background Claude session. A confirmed credential removal now clears the launch key, while a malformed or unreadable store preserves the last-known key. Preserve `--no-auth` as a process-wide opt-out and expose a strict credential-store reader for safe live reconciliation.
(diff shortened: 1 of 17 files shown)


ynadge:
Follow-up on the automated review:

1. Fixed in `fb65f48`: the shared Vitest setup now assigns each test file an isolated temporary `HOME` and `USERPROFILE`, then removes it after the file. Default `startServer()` tests therefore cannot read, inject, or become dependent on a developer's real `~/.sapiom/credentials.json`. I reran the entire Harness suite after this change: 3,157 unit/integration tests and 10 performance tests passed, along with Harness typecheck and lint.

2. The invalid-environment behavior is intentionally retained because it is an explicit SAP-3114 goal-state decision: an invalid custom environment must fail instead of silently sending its credential to production. The failure rejects the affected create/resume/task launch; it does not stop the Studio server. The terminal request error handler returns the original `Unknown environment` message to the client, so this is actionable rather than a generic 500 body.

3. The patch bumps are intentionally retained from the approved release plan. Both API additions are optional/additive (`authMode` defaults to enabled), and the launch-key clearing is the bug fix being shipped. I expanded the changeset in `5a9928d` to state the confirmed-removal versus transient-read-failure behavior explicitly.

The `clear()` versus in-flight refresh race remains deliberately scoped to SAP-3116, which owns login/disconnect races and already-running connection invalidation.

ynadge:
Addressed the remaining release-level review feedback in `3151af8`: both `@sapiom/mcp` and `@sapiom/harness` are now marked `minor`. `changeset status` confirms both packages in the minor release group.

PR #780, opened 2026-09-03
feat(harness): add the Secrets tab, and let a secret exist before a deploy

gwitwer:
Adds the right pane's **Secrets** tab — the credentials one agent's runs receive, with add, replace, delete and `.env` import, all write-only.

The tab is the visible half. The load-bearing half is that **a secret no longer waits on a deploy.**

## The problem

Sapiom stores secrets per cloud definition (`POST /v1/workflows/definitions/:id/secrets`). Until a project is linked there is no id to store anything against, so "set a credential" was only possible *after* the deploy you needed the credential for. `design-eng/agent-studio-v2/docs/GAPS.md` SEC-01 names this: *"Studio has no secret editor anywhere."*

## The shape

Two stores behind one list. A row says which it is.

```
                 unlinked agent                  linked agent
add a secret  →  pending (this machine)  →  POST to vault → synced
                        │                          ▲
                        └───── deploy / upload ─────┘
```

- **Before link** → held in `~/.sapiom/harness/pending-secrets.json` and merged into the run-local child's environment, so a local run gets the same env its deployed counterpart will.
- **On deploy** → flushed after `ensureDefinitionId()` and *before* the build (`link --create` has already provisioned the definition, so the writes work pre-build, and a failed build does not also lose the keys).
- **Deployed from the terminal?** The server never sees that, so the tab names the values that have not shipped and offers an explicit upload rather than letting a deployed run quietly miss a credential.

A successful upload **keeps** the local copy. The vault has no read path by design, so dropping it would end local runs' access with no way to restore it; removing it is its own action.

## Two deliberate departures from the design

Ported from the canonical Studio design (`design-eng/agent-studio-v2`), with two changes the platform forces:

1. **Metadata dropped.** The design draws a value hint, a version pill, Updated and Last used. `GET .../secrets` returns `{ keys: string[] }` and nothing else, so those could be drawn only for values this machine happens to hold — populated for some rows and blank for others is identification you cannot trust. In their place is the one fact the platform *can* confirm: `pending` / `synced`.
2. **Dialogs rebuilt.** The design uses Radix; the harness has no dialog library and should not gain one for four modals. They sit on `modal-backdrop` + `useDismissable`, the idiom `EndSessionConfirm` already keeps.

## Security posture

- **No value reaches the browser.** `GET` carries names and states only; the API key stays server-side, as with `account.ts` and `templates.ts`.
- **No credential in a project directory.** The store lives under the harness state root — a `.env` in a project tree is one `git add -A` from being published, and nothing would read it anyway (there is no `dotenv` in `agent-core`, `agent`, or the CLI).
- **Mode 0600**, re-asserted on every write, with atomic temp-then-rename — the postures `cli/machine-id.ts` and `core/workflow-registry.ts` already prove.
- **Analytics redaction activated.** The panel is tagged `surface: "secrets_panel"`, which switches on the guard shipped in `analytics/before-send.ts` that until now guarded nothing: autocapture drops click text and every element attribute but `class`/`id`, because a field value here can *be* the credential.

## Delivery

**Lane:** standard (PR → required checks → merge queue → `main`).
**Affected:** `@sapiom/harness` only. No migrations, no infrastructure, no new dependency. One new state file (`pending-secrets.json`) created on first write; absent installs read as empty.
**Integrations:** the existing Sapiom core surface (`/v1/workflows/definitions/:id/secrets`). No new endpoint.

### Tests

- `core/pending-secrets.test.ts` (15) — round-trip, per-project isolation, 0600 on create *and* hardening a pre-existing 0644 file, no temp file left behind, corrupt-file recovery that does **not** delete the file, in-memory survival of a failed write.
- `server/secrets.test.ts` (14) — the value appearing nowhere in the serialized body, writes going to the definition from `sapiom.json` and never one the caller names, per-key import outcomes, `?local` deletion scope, local copy kept when the vault refuses a delete.
- `server/actions.test.ts` (+3) — run-local env merge, and a pending secret overriding an inherited shell export.

**Mutation-tested**, each failing only its own test when the guard is removed: value-leak, per-key import failure, flush-keeps-local-copy, 0600, env-merge.

**Suite:** 3283 passed / 12 failed. Those 12 are pre-existing — the same 6 watcher/registry files fail identically on a clean `origin/main` checkout, measured directly rather than assumed.

### Verified against a real server

Driven on a real agent tree with a scratch `--state-root`, not mock mode: add on an **unlinked** agent → `pending`; `GET` returns names only; import; delete; flush refuses with 409 rather than pretending; a linked agent pointing at a nonexistent definition reports `unreadable`, which is distinct from empty. Store file `-rw-------`, nothing written into any project directory, the developer's real `~/.sapiom/harness` untouched.

Rendering it found three defects no test caught, all fixed: the canvas board showed through the panel (`.right-pane-panel` sets `display:flex`, which beats the `hidden` attribute — the pane already has `.is-hidden` for this); `.sr-only` does not exist here, so "value hidden" rendered as visible text; and Status was the first column to shed at the default pane width, because the design's shedding order was copied — the value column, which reads `••••` on every row, sheds first now. The icon registry test also caught `Clock` unregistered, which would have drawn a question mark on every pending row.

**Production signal:** the tab lists an agent's credentials and a first deploy reports the keys it uploaded. Regression looks like a deploy emitting `warning` lines naming keys that did not land, or a `pending` row that never turns `synced`.

```text
Fix-forward: the surface is additive and self-contained — revert the two commits to remove the tab and the routes; the local store is inert without them and holds no state anything else reads.
```

## Note for the reviewer

No Linear ID: no existing issue covers the Studio-side tab (the nearest are dashboard/webapp secrets work). Happy to file one and re-title if Main Eligibility wants it.

https://claude.ai/code/session_01FHRGEUo3kc1GjmCDe3k3Un


gwitwer:
## Evidence

**Before** is the absence of the surface: on `main` the right pane is `Canvas | Steps`, and there is no way to set a credential anywhere in Studio before an agent is deployed.

### The tab, on an unlinked agent

Both values are `pending` — set before this agent exists in the cloud, which is the thing that was not possible before. The value column reads `••••` and sheds first at this pane width; Status is what survives, because "is this actually on Sapiom yet" is the question the tab answers.

Secrets tab

### Add, with the destination stated before the click

An unlinked agent says where the value is about to land — *"Saved on this machine and used by local runs. Deploying this agent uploads it to Sapiom."* A linked one says *"Applies to new runs. Existing runs keep the value they started with."* A user who is not told the difference reads "saved" as "deployed."

Add secret dialog

### A full write through the real UI

Typed `stripe-api key`, normalised to `STRIPE_API_KEY` on the way in, saved, listed as `pending`. Zero page errors, and the value appears nowhere in the DOM.

Write round-trip

All three were taken against a real harness server on a real agent tree with a scratch `--state-root`, after the rebase onto current `main` — not mock mode.

gwitwer:
Thanks — all three findings were real. Fixed in `9a04bd8a`.

### 1. Plaintext copy on every write, under-disclosed — fixed

You're right that the copy is written on both paths, and right that the copy describing it was scoped to the pre-link case. The behaviour is deliberate (it is what gives a local run the same environment its deployed counterpart gets, since there is no read path for a stored secret), but that is a fact the user has to be told at the point of entry.

- **Changeset rewritten**, and this is now its own paragraph rather than an implication: *"Studio keeps a plaintext copy of every secret you enter, on every path — not only before linking."* It also names the file, the mode, and that deleting a secret removes the local copy with it.
- **All three dialogs** now disclose it on the linked branch: *"Applies to new runs; runs already in flight keep the value they started with. A copy is also kept on this machine, so local runs get the same value."* The import dialog previously showed a destination line only when unlinked — the common case, pasting a `.env` onto a deployed agent, was silent. It now speaks on both paths.

### 2. Undeletable pending secret on a linked agent — fixed

Confirmed exactly as described, and reproduced before fixing (502 where 204 belonged). The vault call was gated on linkedness rather than on whether the vault holds the key.

Fixed at **both** layers, because either alone leaves the trap available:

- `vault-secrets.ts` treats **404 on DELETE as success** — the vault not holding the key is the state the caller asked for. DELETE is idempotent.
- `secrets.ts` no longer strands the local copy if a `VaultSecretError` with status 404 reaches it, so a future client cannot reintroduce the dead end.

You also identified the reason it got through: `secrets.test.ts` covered "the vault refuses" but not "the key was never there." Added **`vault-secrets.test.ts` (10 tests)** — that module had none at all — covering the 404 path, the Bearer-not-`x-sapiom-api-key` contract, refresh-and-retry-once, `null` vs `[]` on a failed read, and that a refusal message names the key without echoing the value. The 404 line is mutation-tested: removing it fails that test and only that test.

### 3. Provider name in rendered copy — fixed, with one correction

Placeholder is now `API_KEY=…`; the other two lines were already neutral. Same drift cleared from `pending-secrets.test.ts` and `actions.test.ts`.

I added both dialogs to `STATIC_PROVIDER_NEUTRAL_COPY_TARGETS` as you suggested, then checked whether that actually guards this — **it does not, and I don't want that recorded as protection it isn't.** `FORBIDDEN_PROVIDER_COPY_RE` is `/\b(?:anthropic|…)\b/`, and `\b` fails against `ANTHROPIC_API_KEY` because `_` is a word character. Verified both directions on the audited file: a bare `Anthropic` in prose **fails** the check; `ANTHROPIC_API_KEY` **passes** it. So the entry is real but partial — it catches a prose vendor name, not the underscored token. Closing that gap properly means the regex, which felt out of scope here.

### Also, from CI

`playwright-mock` failed on `project-altitude.spec.ts`, which asserted `.right-pane-tab` `toHaveCount(2)`. Its stated point is that the map grows no tab of its own, not that the row is frozen at two, so rather than bump the number I extended it to cover Secrets under the same one-agent rule it already applies to Steps — disabled at map altitude, with its reason on `data-tooltip`.

One note for anyone reading the next run: `open-project.spec.ts:262` is **flaky in this repo, not a regression here**. Alternating clean local runs between `origin/main` and this branch gave pass, pass, fail, fail — the outcome tracked run order, not the ref. It passed in CI on this branch.

PR #790, opened 2026-09-03
fix(harness): keep agent creation in planner flow

ynadge:
## Summary

Agent Map planning is now the only agent-creation route in current Studio projects. The sidebar no longer exposes standalone creation or in-session scaffolding that can bypass the planning flow. The pinned planning tab is labeled “Plan Agents” so its purpose is explicit.

## Changes

- hide the empty-project “Create the first agent here” row
- hide direct create and bare-session scaffold actions from project overflow menus across both rail axes
- rename the pinned Agent Map tab and its hover states to “Plan Agents”
- retain Agent Map access, project removal, and unsearched-checkout information
- retain compatibility behavior for older state payloads without a durable Studio project identity
- add production-shaped browser coverage for all three removed creation doors and the renamed planning tab
- release Harness as a minor change and Desktop as a patch

## Testing

- [x] `pnpm --filter @sapiom/harness... build`
- [x] `pnpm --filter @sapiom/harness test` — 3,306 unit and 10 performance tests
- [x] focused creation-flow Playwright suite — 16 tests
- [x] project-axis Playwright suite — 19 tests
- [x] Prettier and diff checks

## Screenshots/Demos

This removes the reported “Create the first agent here” row and renames the pinned planning tab to “Plan Agents” without restyling the surrounding UI. Browser regressions assert that the empty-state CTA, project-menu create action, and bare-session scaffold action are absent while the planning tab and project removal remain available.

### DIFF
--- a/.changeset/planner-owned-agent-creation.md
+++ b/.changeset/planner-owned-agent-creation.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/harness": patch
+"@sapiom/harness-desktop": patch
+---
+
+Keep planner-managed Studio projects on the Agent Map flow by removing direct agent-creation actions from empty sidebar rows and project menus. Legacy non-planning harness projects retain their existing creation controls.
--- a/.changeset/planner-owned-agent-creation.md
+++ b/.changeset/planner-owned-agent-creation.md
@@ -1,6 +1,6 @@
 ---
-"@sapiom/harness": patch
+"@sapiom/harness": minor
 "@sapiom/harness-desktop": patch
 ---
 
-Keep planner-managed Studio projects on the Agent Map flow by removing direct agent-creation actions from empty sidebar rows and project menus. Legacy non-planning harness projects retain their existing creation controls.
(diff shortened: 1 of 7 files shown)


ynadge:
Addressed the round-one findings in 1d69d1f5:

- made the intended current-Studio-wide scope explicit in both the changeset and PR description
- changed the Harness release entry from patch to minor
- added production-shaped Agent Map coverage proving the empty-state create CTA, project-menu create action, and bare-session scaffold action are all absent
- kept Agent Map and project removal available
- removed the redundant loading assertion and made the test title describe the absence contract accurately

The compatibility callbacks remain narrowly available for older state payloads without a durable Studio project identity; current Studio projects use Agent Map as the sole creation route.

ynadge:
The round-two follow-up is addressed in 6ec81867. The retained flow now has an explicit `mockStudioProjects=absent` fixture mode, and every spec that exercises direct create/scaffold behavior opts into and labels that legacy-server contract. Current Studio behavior remains covered separately by the Agent Map fixture. The expanded local browser run passed all 116 relevant tests.

PR #794, opened 2026-09-03
fix(harness): rename Agent Map tab

ynadge:
## Summary

- rename the pinned Studio project sidebar tab from “Agent Map” to “Plan Agents”
- update the tab hover text for both selected and unselected states
- direct the stop-planning confirmation back to “Plan Agents”
- keep “Agent Map” as the name of the proposal view, artifact, and protocol
- connect the two names in the published Harness README
- add a patch changeset for Harness and the bundled Desktop app

“Plan Agents” is intentionally action-oriented navigation: it tells users what they do from the pinned entry point. Once opened, “Agent Map” continues to identify the artifact they are viewing and the underlying protocol. Internal identifiers are unchanged.

## Testing

- [x] `pnpm --filter @sapiom/harness... build`
- [x] `pnpm --filter @sapiom/harness exec playwright test --config web/e2e/playwright.config.ts web/e2e/project-axis.spec.ts web/e2e/agent-map-planning.spec.ts` — 31 passed
- [x] Prettier check on all changed files
- [x] `git diff --check`

### DIFF
--- a/.changeset/plan-agents-label.md
+++ b/.changeset/plan-agents-label.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/harness": patch
+"@sapiom/harness-desktop": patch
+---
+
+Rename the pinned Agent Map tab in the Studio project sidebar to Plan Agents so its purpose is clear. The underlying Agent Map view and behavior are unchanged.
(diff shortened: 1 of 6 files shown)


ynadge:
Addressed the review in 135412aa:

- kept the requested UI scope intentionally narrow: **Plan Agents** is the action-oriented sidebar entry, while **Agent Map** remains the proposal view/artifact/protocol name
- expanded the PR description with that product-language rationale
- updated the published Harness README to explicitly connect the **Plan Agents** tab to the Agent Map documentation
- kept the changeset at patch because this changes copy only; it does not alter defaults, behavior, APIs, or compatibility

ynadge:
Addressed the round-two navigation-copy finding in ca6ca74b. The planning-session confirmation now says “from Plan Agents,” and the focused browser test asserts that exact recovery path. The Agent Map view/artifact labels remain unchanged under the documented naming rule.

PR #796, opened 2026-09-03
fix(harness): add project session shortcut

ynadge:
## Summary

Add a project-row `+` immediately before the overflow menu so a developer can start a coding-agent session at that project root in one click. This is a new session action, distinct from the former agent-creation meaning of the glyph; Sapiom agent creation remains owned by Plan Agents.

## Changes

- Add the hover-revealed project-root session shortcut beside `⋮`
- Use the stored preferred coding harness for both the shortcut and project-start path
- Open the new session in the workbench only after creation succeeds
- Keep an active planning conversation visible on failure and resumable after success
- Keep direct Sapiom agent creation hidden for plan-first projects
- Document the project-row and tab-strip session shortcuts
- Add a minor harness / patch desktop changeset with corrected user-facing copy

### Key Refinements from Code Review

- Seed and assert the Codex preference instead of only exercising the Claude default
- Make the provider argument required so new project-session callers must choose consistently
- Preserve Plan Agents when session creation fails
- Cover keyboard reveal for both `+` and `⋮`
- Remove unreachable duplicate error feedback

## Testing

- [x] `pnpm --filter @sapiom/harness build`
- [x] `pnpm --filter @sapiom/harness test` — 3,316 tests passed
- [x] 53 targeted Chromium tests across project axis, rail grammar, project opening, and accumulation guard
- [x] Red/green check: the shortcut test fails with the button removed and passes after restoration

## Screenshots/Demos

- Visually checked the hover state in Chromium at 1440×900 against the supplied reference: `+` is directly before `⋮` and reuses the existing row-action treatment.

## Checklist

- [x] No hardcoded secrets
- [x] Self-reviewed

### DIFF
--- a/.changeset/project-session-shortcut.md
+++ b/.changeset/project-session-shortcut.md
@@ -3,4 +3,4 @@
 "@sapiom/harness-desktop": patch
 ---
 
-The project-row `+` now starts a coding-agent session at that project root. Previously it created an agent and was removed in favor of the row's `⋮`; Sapiom agent creation remains owned by Plan Agents.
+The project-row `+` now starts a coding-agent session at that project root. Previously it created an agent. Sapiom agent creation remains owned by Plan Agents.
--- a/.changeset/project-session-shortcut.md
+++ b/.changeset/project-session-shortcut.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/harness": patch
+"@sapiom/harness-desktop": patch
+---
+
+Restore the project-row `+` shortcut so users can start a coding-agent session at that project root without opening the overflow menu. Sapiom agent creation remains owned by Plan Agents.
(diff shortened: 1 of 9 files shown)


ynadge:
Addressed the automated review in 6589ed0f:

- corrected the changeset wording and raised the harness bump to minor
- seeded `preferredHarness: "codex"` in the shortcut test and made the provider argument required
- made row selection use the same stored provider
- delayed leaving Plan Agents until session creation succeeds, with failure coverage
- removed the unreachable duplicate catch/toast path
- restored keyboard-focus coverage for both the shortcut and overflow trigger

Fresh local verification: build, typecheck, 3,316 unit/performance tests, and 53 targeted Chromium tests all pass.

ynadge:
Trimmed the remaining ambiguous changeset clause in c6a0d91d. The published copy now says only that the glyph previously created an agent and that Sapiom agent creation is owned by Plan Agents; it no longer points readers to the overflow menu.

PR #801, opened 2026-09-04
fix(harness): one definition of what a project is [SAP-3142]

gwitwer:
## The bug

`WorkflowsRail.tsx:1232-1240` joins each project row to a workspace scope by
`samePath(scope.cwd, project.root)`, reads the durable Studio project off that
scope's `projectId`, and sets `mapOwnsCreation = studioProject != null`. When it
is false the retired direct-creation UI renders: the project-row create item
(`:1410`) and the empty-project row (`:1484`). Its own comment says the absent
case is "a compatibility payload, not a second creation mode".

The absent case fired routinely. Two definitions of "project" existed and they
disagreed on the commonest shape there is:

| | definition |
|---|---|
| the rail | `projectRoots` (`lib/project-tree.ts`). Rule 1 replaces an agent's OWN directory with `holdingProjectFor(dir) ?? dir`, i.e. the folder that HOLDS the agent |
| the server | `settings.recentDirs` + every session cwd (`index.ts:1209-1212`), one durable Studio project per entry |

That holding folder was never a recentDir and never a session cwd, so it was
never a scope, so the join found nothing.

**Reproduced, on a real server, before the fix:**

```
node dist/cli/bin.js ~/sapiom/agents/property-ops/tenant-screening \
  --port 4131 --state-root <fresh> --no-session --no-telemetry --no-open
```

`GET /api/state` returned one scope, `cwd: .../property-ops/tenant-screening`.
The rendered row was `property-ops`. `find` returned undefined, `agent-map-row`
was absent, and the project menu offered **"Create an agent in property-ops"**.

## The fix, and why it is on this side

`projectRoots` moves to `src/shared/project-roots.ts` (with `paths.ts`, the pure
string helpers it compares with — `web/src/lib/paths.ts` is now a re-export
barrel so no web import changed) and `LocalWorkspaceScopeCatalog` calls it.

The two candidates were: register the derived holding root as a scope, or make
the join tolerant of a scope nested under the row's root. **One source of truth
picked the first.** `project-tree.ts` already says it in the source: "`projectRoots`
is the one place that decides what a project is". Deriving the scope list from it
makes the catalog downstream of that definition and mints a durable Studio
project for exactly the roots the rail draws. A tolerant join would have been a
second, weaker rule, and it would bind a row to a project whose `localRootRef` is
NOT that row's root — the Agent Map for `property-ops` would have been rooted at
`tenant-screening`, and a session started there would resolve to a different
project than the map above it (`resolveIdentityForPath` picks the most specific
active binding).

Deliberately not done: defaulting `mapOwnsCreation` to true, which would render
an Agent Map for a project that has none.

## Real-server evidence

Reproduction tree, `~/sapiom/agents/property-ops/tenant-screening`:

| before | after |
|---|---|
| before | after |

- before: scope `cwd` = `.../property-ops/tenant-screening`; no `agent-map-row`;
  menu = `["project-create-agent-property-ops", "project-remove-property-ops"]`
- after: scope `cwd` = `.../property-ops` with a `projectId`; `agent-map-row` +
  `agent-map-select` render; menu = `["project-remove-property-ops"]`
- `studio-projects.json` after: exactly one project, `property-ops`, binding
  `active`. No stray project for the agent folder.

Control, `~/sapiom/agents` (76 projects, launched at the holding folder, where
the join already worked) — byte-identical before and after:

| before | after |
|---|---|
| before | after |

## Independently confirmed, from the opposite direction

The dialog-shell agent (#800) hit the same boundary while converging dialogs onto
a shared shell, without looking for it: **`CreateAgentDialog` is not reachable on
a current server.** `project-create-agent-<label>` does not render against
`~/sapiom/agents` at all, and it had to pass `?mockStudioProjects=absent` to
screenshot the dialog even once.

That is the control case passing, observed by someone who wanted it to fail. It
pins the defect precisely: launched at a folder that HOLDS agents the gate holds
and the compatibility dialog never appears; launched at an agent's own directory
the rail derives a holding root the server never registered and the same dialog
comes back. The fix is that one join, not a change to how creation works.

## The mock default, and what it exposed

`lib/api.ts` returned `undefined` for `mockStudioProjects` unless a spec opted
in, so **every** UI test ran the retired branch and none ever rendered the
shipped one. A green suite was evidence about the wrong branch. The default is
flipped; `mockStudioProjects=absent` now names the compatibility payload where a
spec wants it. The mock's `workspaceScopes()` also derives from `projectRoots`
now, so the fixture cannot disagree with the server about what a project is.

**192 of 497 specs failed on the flip.** Three causes, three responses:

1. **A helper named the wrong class (1 fix, ~120 specs).** `focusRfqAgent` clicked
   `.workspace-row-main`, which exists only on a MERGED root-agent row. A
   plan-first project renders the agent as a separate child row. Both carry
   `.workflow-item-trigger`; the helper now names that.
2. **The mock landed every boot on the Agent Map (1 fix, ~65 specs).** With a
   durable project and a live session and nothing stored, `App.tsx:352`'s restore
   effect takes the server's default selection, which is `agent-map`. The suite's
   fixture is a RETURNING user with a session bound to an agent, so the mock's
   unstored default is now that agent. `mockStudioProjects=present` still yields
   the map, which is what `agent-map-planning.spec.ts` asserts.
3. **21 spec files describe the compatibility rail (pinned).** Chiefly the
   workspace System Graph a project row opens: on a current server the row opens
   the project's Agent Map instead, so `workspace-graph-view` /
   `system-graph-canvas` is a legacy-payload surface. Those files now say
   `mockStudioProjects=absent` at their navigation, with the reason written in.
   **This takes no coverage away** — every one of them already ran on that payload
   before this PR; it only stops the choice being an accident. What it does NOT
   do is cover their plan-first equivalents. That gap is real and named, not
   silent. See "Known gaps" below.

Two locator fixes were needed regardless: a plan-first group has a SECOND direct
`.workspace-row` child (the pinned Agent Map row), so `:scope > .workspace-row`
is a strict-mode violation there. Now `:not(.is-nested)`.

Net: **497 passed / 0 failed**, identical to `origin/main`'s baseline on this
machine. 21 spec files + 1 helper changed.

## Known gaps, stated

- **`accumulation-guard` and `project-axis` contradict each other, and now it is
  visible.** `project-axis.spec.ts:168` specifies "a root agent is a separate
  target below the pinned Agent Map" — project header + map row + agent row, three
  rows. `accumulation-guard`'s SYMPTOM sweep forbids exactly that: "no project row
  is a single agent wearing its own folder's name". Neither could see the other:
  one ran only on the opt-in payload, the other only on the default. I merged the
  root agent into the plan-first row to satisfy the guard, which broke
  `project-axis`, and reverted it. **Which one is right is a product decision**, so
  the sweep is pinned to the compat payload with the conflict written into the
  fixture list rather than resolved by editing an assertion.
- The plan-first behaviour of the 21 pinned files (project selection, session
  scoping, the deploy/canvas flows under a durable project) has no coverage.
- The rail's `mapOwnsCreation` join is not the only `samePath(scope.cwd, root)`
  join in the SPA. `App.tsx:1836` and `:2702` do the same lookup against a project
  root and had the same latent bug; the shared derivation fixes them by making the
  scope exist, but they are still exact-match joins.

## Review rounds

**This PR had no tests and no review until its conflict cleared**, and that is
worth recording because the symptom is silence. A `pull_request` workflow builds
against the merge ref GitHub computes; a conflicting PR has none, so every job
needing one never fires — no failure, no skip notice, nothing queued. Four jobs
ran (CodeQL, both Analyze, Classify) and six did not, including every harness
suite and the review bot. It read as green. After merging `main` all ten run and
all ten pass.

### Round 1

1. **The boot race persisted the stray this fix removes.** Correct and worse
   than the PR body said. At boot `workflowsCache` is the persisted registry —
   empty on a first run — so rule 1 cannot fire, the agent's own folder survives
   as a root, and `reconcile` MINTS AND PERSISTS a project for it. When the scan
   lands the binding flips to "missing" and the row stays in
   `studio-projects.json` forever, unreachable from the UI.
   Awaiting the boot scan was the first fix and it broke a contract this suite
   states by name — "serves persisted cold inventory without awaiting
   discovery". So the catalog asks the disk instead: a candidate holding a
   `sapiom.json` is an agent directory. One stat per candidate over recentDirs
   plus live session cwds, no discovery, and it falls back to exactly the
   registry once the scan lands.
2. **No migration; the dropped projects were orphaned.** `reconcile` now
   re-points the binding when the move is unambiguous, so the Agent Map
   aggregate, `studioBindings` and persisted planner identities follow the
   project instead of being stranded on an id nothing points at.
3. **The accumulation guard no longer guards the shipped rail.** Not fixed —
   see "Known gaps". It is a product decision between two specs that contradict
   each other, and it is escalated rather than settled by editing an assertion.
4. **One rule, two input sets.** Documented on `ProjectRootSources`:
   `closedProjects` is browser localStorage, so a removed project whose agents
   stay registered is hidden in the rail and still a root on the server.
5. **Public copy.** `src/shared/project-roots.ts` is compiled by `tsc`, which
   keeps comments, so its prose now ships in the tarball where Vite used to
   strip it. The dogfood capture filename and per-install counts are gone.

### Round 2

The migration added for finding 2 opened a durable-write path of its own, and
the review was right about all three faces of it:

- **The adoption pool was unbounded in time.** "Any project whose bindings are
  all missing" grows forever, because nothing deletes a project and every folder
  that ages out of the capped `recentDirs` leaves one behind — so a new root
  appearing months later would adopt a stranger. The pool is now the bindings
  THIS pass just took away.
- **The first containing root adopted, not the nearest.** Scopes arrive sorted
  by canonical path, so `/w` reached an orphan under `/w/team/a2` before
  `/w/team` did, landing the identity on the wrong folder.
- **`displayName` was rewritten.** `moveRootBinding` deliberately leaves it
  alone and this now matches: the folder moved, the project the user named did
  not.

Each has its own spec, and each spec dies when its correction is reverted.

### What CI caught that local runs hid

Two `system-graph-freshness` specs failed on CI with `expected undefined to be
truthy`. Locally they sit inside a set of six files that fail under contention
on this machine, so they read as part of the documented flake cast. **Local
"known flaky" was hiding a real regression.**

Both fixtures wrote `index.ts` straight into `workspaceRoot`, making that root
an agent directory — so this PR correctly promotes it and there is no scope at
that exact path. My first fix resolved the workspaceKey through the CONTAINING
scope, which made the lookups succeed and then failed deeper: budgets 500/500
instead of 200/200, pass count 13 instead of 2. The assertion was satisfied
while the subject had silently moved to the parent workspace. The fixtures are
what needed changing; both now scaffold one level down, as every other fixture
in that file already does.

## Tests

- `src/server/studio-project-scopes.test.ts` (new): boots a real server three
  ways and asserts the rail's own three-step join over the state the SPA receives
  — for every root `projectRoots` returns, a scope exists at that exact root whose
  `projectId` names a project in `studioProjects`.
  - launching INSIDE an agent folder (the bug)
  - launching AT a folder that holds agents (control)
  - recentDirs already naming the holding folders (control)
  - **Mutation-tested**: reverting `index.ts` to the raw candidate list fails test
    1 at the scope lookup (`expected null to be '…/property-ops'`) and leaves the
    two controls green. It is not a count assertion that passes when nothing
    happened.
- `pnpm test:ui` — 497 passed / 0 failed (baseline on `origin/main`: 497 / 0).
- `pnpm test` (vitest) — 14 failed / 3295 passed. Baseline on `origin/main` on the
  same machine: 13 failed / 3293 passed, in the **same six files**
  (`workflow-registry`, `system-graph-relationships`, `system-graph-watcher`,
  `workspace-watcher`, `agent-source-discovery`, `system-graph-freshness`), all
  sleep/watcher timing. Pre-existing, load-dependent; not from this change.
- `pnpm typecheck` clean. `pnpm lint` is `eslint src` and does not reach
  `web/src`, so it says nothing about most of this diff.

## One ordering note

`startServer` does not await the boot agent scan, and the derivation reads the
agent registry: before the first scan lands, an agent folder looks like an
ordinary folder and would be registered as its own project. In practice the scan
beats the first `/api/state` — the real-server run above produced exactly one
durable project with no stray — but the new spec waits for the scan rather than
pretending the race is not there.

### DIFF
--- a/.changeset/one-project-definition.md
+++ b/.changeset/one-project-definition.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+Derive Studio workspace scopes from the same `projectRoots` rule the rail renders, so every project row on a real install has a durable Studio project and its Agent Map owns creation. Launching Studio inside an agent's own folder promoted the rail's row to the folder that holds it, while the server had registered only the agent folder, so the row's join failed and the retired direct-creation UI rendered instead.
--- a/.changeset/one-project-definition.md
+++ b/.changeset/one-project-definition.md
@@ -2,4 +2,6 @@
 "@sapiom/harness": patch
 ---
 
-Derive Studio workspace scopes from the same `projectRoots` rule the rail renders, so every project row on a real install has a durable Studio project and its Agent Map owns creation. Launching Studio inside an agent's own folder promoted the rail's row to the folder that holds it, while the server had registered only the agent folder, so the row's join failed and the retired direct-creation UI rendered instead.
+Fix Studio showing the retired agent-creation controls instead of the project's Agent Map. Launching the harness inside an agent's own folder made the sidebar draw a row for the folder that holds it, which the server had never registered, so the project's Agent Map could not be found and the older create menu came back. The sidebar and the server now decide what counts as a project the same way.
+
+Existing installs need no migration. A project whose folder moves up to the folder holding it keeps its identity, so its Agent Map, its agents and its planning sessions come with it; where that move is ambiguous, because one folder holds several projects that all moved, a new project is created and nothing existing is altered.
--- a/packages/harness/src/core/studio-project-catalog.test.ts
+++ b/packages/harness/src/core/studio-project-catalog.test.ts
@@ -61,6 +61,73 @@ describe("StudioProjectCatalog", () => {
     expect(JSON.stringify(second.projects)).not.toContain("workspace-legacy");
   });
 
+  it("a root that moved UP keeps its project rather than orphaning it", async () => {
+    // The rail's rule replaces an agent's own directory with the folder that
+    // HOLDS it, so a scope legitimately turns into an ancestor of itself. If
+    // reconcile mints for the ancestor and leaves the old entry behind, the two
+    // halves of one project split: the Agent Map aggregate, `studioBindings` and
+    // persisted planner identities all stay keyed to the id that was abandoned.
+    const { root, catalogPath } = await fixture();
+    const holding = path.join(root, "property-ops");
+    const agentDir = path.join(holding, "tenant-screening");
+    await fs.mkdir(agentDir, { recursive: true });
+    const catalog = new StudioProjectCatalog(catalogPath);
+
+    const before = await catalog.reconcile([
+      { workspaceKey: "workspace-agent", cwd: agentDir },
+    ]);
+    const original = before.projects[0]!.projectId;
+
+    // The scan lands, rule 1 fires, and the scope is now the holding folder.
+    const after = await catalog.reconcile([
+      { workspaceKey: "workspace-holding", cwd: holding },
+    ]);
+
+    expect(after.projects).toHaveLength(1);
+    expect(after.projects[0]!.projectId).toBe(original);
+    expect(after.projects[0]!.displayName).toBe("property-ops");
+    expect(after.workspaceScopes[0]?.projectId).toBe(original);
+    // And a session anywhere under the new root still resolves to it.
+    expect(
+      (await catalog.resolveIdentityForPath(agentDir))?.projectId,
+    ).toBe(original);
+  });
+
+  it("refuses to merge two orphans into one root, and mints instead", async () => {
+    // TWO agent folders promoted to one holding root. Re-pointing either one
+    // would silently merge two identities into one, which is not reversible and
+    // is worse than the split it fixes. Ambiguity mints fresh and leaves both
+    // originals alone with their bindings marked missing.
+    const { root, catalogPath } = await fixture();
+    const holding = path.join(root, "team-bots");
+    const first = path.join(holding, "backlog-nudge");
+    const second = path.join(holding, "compliment-bot");
+    await fs.mkdir(first, { recursive: true });
+    await fs.mkdir(second, { recursive: true });
+    const catalog = new StudioProjectCatalog(catalogPath);
+
+    const before = await catalog.reconcile([
+      { workspaceKey: "workspace-first", cwd: first },
+      { workspaceKey: "workspace-second", cwd: second },
+    ]);
+    const originals = before.projects.map((project) => project.projectId).sort();
+
+    const after = await catalog.reconcile([
+      { workspaceKey: "workspace-holding", cwd: holding },
+    ]);
+
+    expect(after.projects).toHaveLength(3);
+    const minted = after.workspaceScopes[0]?.projectId;
+    expect(originals).not.toContain(minted);
+    // Both originals survive untouched rather than one being quietly adopted.
+    expect(
+      after.projects
+        .map((project) => project.projectId)
+        .filter((id) => originals.includes(id))
+        .sort(),
+    ).toEqual(originals);
+  });
+
   it("resolves cwd containment to the most-specific active project without paths", async () => {
     const { root, catalogPath } = await fixture();
     const parent = path.join(root, "workspace");
--- a/packages/harness/src/core/studio-project-catalog.test.ts
+++ b/packages/harness/src/core/studio-project-catalog.test.ts
@@ -85,12 +85,77 @@ describe("StudioProjectCatalog", () => {
 
     expect(after.projects).toHaveLength(1);
     expect(after.projects[0]!.projectId).toBe(original);
-    expect(after.projects[0]!.displayName).toBe("property-ops");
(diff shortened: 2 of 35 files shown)


gwitwer:
/review

The automatic round never fired on this PR: `gh run list --workflow=claude-code-review.yml` shows no run for `gwitwer/studio-rail-creation` at all, while the stacked #803 got its round 1 normally. Not a skip and not a silent pass — the `opened` event produced no run. Triggering it by comment.

PR #802, opened 2026-09-04
fix(harness): keep new-agent prompts in builder sessions

ynadge:
## Summary
Keep Create new agent submissions in the standalone coding session that receives the submitted prompt. Prevent project preference restoration from replacing it with a Plan Agents session.

## Changes
- Record standalone-builder intent before the generic session and project catalog refresh begin.
- Suppress containing parent-project restores until Studio resolves the exact new project root.
- Bind the one-shot intent to the created session, retain it across focus changes, and expire it only when that session exits or disappears.
- Preserve normal explicit Plan Agents entry and clean up failed creation attempts.
- Add regression coverage for an unrestored parent project and for leaving and returning before catalog refresh completes.
- Add patch changesets for the harness and desktop app.

## Testing
- [x] pnpm --filter @sapiom/harness typecheck
- [x] pnpm --filter @sapiom/harness build
- [x] Composer Playwright suite: 18 passed
- [x] Parent-race regression repeated 5 times: 5 passed
- [x] Revisit regression repeated 5 times: 5 passed
- [x] Regressions verified red before their fixes and green afterward

## Screenshots
Not applicable; this changes session selection behavior without changing visuals.

## Checklist
- [x] No hardcoded secrets
- [x] Self-reviewed

### DIFF
--- a/.changeset/manual-builder-session.md
+++ b/.changeset/manual-builder-session.md
@@ -1,0 +1,6 @@
+---
+"@sapiom/harness": patch
+"@sapiom/harness-desktop": patch
+---
+
+Keep new-agent composer submissions in their standalone coding session instead of automatically switching to Plan Agents.
--- a/packages/harness/web/e2e/new-session-composer.spec.ts
+++ b/packages/harness/web/e2e/new-session-composer.spec.ts
@@ -9,6 +9,8 @@
 import { expect, test } from "@playwright/test";
 import type { Page } from "@playwright/test";
 
+import { selectMockSessionFromPalette } from "./mock-navigation";
+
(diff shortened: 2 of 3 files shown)


ynadge:
Addressed both review findings in 40a231c9: containing parent restores are suppressed while the builder intent is pending, the intent is bound to the created session and expires on focus/exit, and the regression now begins with the parent project unrestored.

ynadge:
Addressed the follow-up finding in 58d9d316. The intent remains tied to its specific builder session across focus changes, expires only if that session exits or disappears, and a new regression switches away before catalog refresh then returns and verifies Plan Agents stays inactive.

PR #803, opened 2026-09-04
feat(harness): one create verb, and the folder step comes first [SAP-3143]

gwitwer:
> Stacked on #801 (SAP-3142) and based on its branch, so this diff is only PR 2's
> change. #801 is the data fix; it must be provable on its own. GitHub retargets
> this to `main` when #801 merges.
>
> **NO TEST JOB HAS EVER RUN AGAINST THIS PR.** `test.yml` and `harness.yml` both
> trigger on `pull_request: branches: [main]`, and this PR's base is #801's
> branch, so neither fires — nor do `playwright-mock`, `desktop-smoke` or
> `examples`. The only workflow that runs here is Claude Code Review, which has
> no branch filter. **Read its rounds as code review, not as evidence anything
> runs.** The suites below were run locally; CI coverage arrives when #801 merges
> and this retargets to `main`.

SAP-3143. `SAP-2981` is titled **"Canonical create-agent flow"**, so the prompt
injection this replaces is the mechanism that ticket exists to remove, not
merely something adjacent to it.

## Three controls, one job

| control | testid | what it did |
|---|---|---|
| black CTA, first row of the nav | `rail-create-new` | `onNewSession` → `setComposing(true)` |
| nav row under it | `add-existing-agents` | `setStartMode("detect")` |
| Projects header folder-plus | `rail-add-project` | `setStartMode("open")` |

`agent-studio-demo/docs/IA-REVIEW.md:522` states the rule they broke:

> One create verb, `New agent`, as the rail header's `+`. Not a row among
> places: a verb in a list of destinations is the F-18 confusion.

Two of the three were rows in a list of destinations. Neither of the create-ish
ones asked the one thing creation cannot proceed without — where the agent
lives. The CTA answered it by inventing a folder from a slug of whatever you
typed into the composer.

## What it is now

**One create verb, the Projects header's `+`, `aria-label="New agent"`.** Its
first step is a popover: the projects the rail already shows, plus one way out
to a folder it does not.

| before | the verb's first step | where it lands |
|---|---|---|
| before | step 1 | step 2 |

The whole flow, recorded against `~/sapiom/agents`:
**[pr2-create-flow.webm](https://raw.githubusercontent.com/sapiom/sapiom-js/pr-media/media/803/pr2-create-flow.webm)**

The nav now holds destinations only (Search, Templates). "Add existing agents"
is a real job and stays one click deep — in the rail's own ⋮ settings menu,
where its housekeeping already lives — without standing beside the verb as a
peer: 

## (a) The folder step opens the native picker on desktop

The bridge worked already (`FolderField.tsx:55` → `dialogs.ts:41-44` →
`desktop.mts:55`). The split was one level too deep: `FolderField` asked
`getDesktopBridge()?.chooseDirectory` and used the answer to decide whether to
render a `<datalist>`, so a desktop user still got a modal wrapping a text field
with Finder offered as a button inside it. With a real OS folder browser
available, our dialog is not a smaller version of it.

The bridge now decides whether **the dialog opens at all**, in
`web/src/lib/folder-step.ts` — one shared place, because there is more than one
entry point and two copies of this branch would eventually disagree.

**Which half is proved how, since only one can run here:**

- **Web** — proved end to end. The CLI harness in a browser IS the web host, so
  every `test:ui` run exercises it. `folder-field.spec.ts`'s "browser host (npx)"
  describe still asserts no Choose button and a working datalist.
- **Desktop** — proved by unit-testing the branch (`folder-step.test.ts`, 5
  cases). There is no Electron in this repo's test environment and the Playwright
  tier is the browser host by definition, so a branch only the other host can run
  needs a test that does not need that host. The specs assert the NEGATIVE too —
  that our dialog never opens when the bridge is there — because a test that only
  checked `chooseDirectory` was called would pass on a build that opened both.
  **Mutation-tested**: making it call `openDialog()` unconditionally fails 3 of
  5; dropping the `onPicked` call fails 1. It is not an assertion that survives
  its own mutation.
- `folder-field.spec.ts`'s "desktop host" describe now reaches `FolderField`
  through the ⋮'s detection dialog instead of the create verb, because the verb
  deliberately never opens a dialog on that host any more.

## (b) The create verb stops typing English at the agent

`App.tsx`'s `sendScaffoldPrompt` starts a session at an invented folder and
injects `composerScaffoldPrompt` — an English request that the coding agent
please call `sapiom_dev_agents_scaffold`. It is the mechanism SAP-2981 set out to
remove and it was wired to the most prominent control in the product. It failed
the way a prompt fails: a create that did not happen arrived as a confused model
rather than an error.

`aaf721bb` (#790) had already fixed this for the project-row doors, and the fix
is the same one — **when a project has a durable Studio project, its Agent Map
owns creation**. So `handleNewAgentIn` resolves the folder to a project and hands
over to that map, exactly as selecting the project does. The folder does not have
to be a project yet: `openProject` is what makes it one (and resolves an agent's
own folder to the folder that holds it), and after #801 the server issues a
durable Studio project for exactly that root.

A payload with no Studio project catalog is the one case the map cannot serve.
It falls through to `handleCreateAgentInProject` — the dialog whose endpoint
creates the agent on disk and rescans before any session starts. Still the
harness creating the agent; still no sentence sent to a model.

**Measured on a real server** (`~/sapiom/agents`, 76 projects), pressing the verb
and confirming the project:

```
after picking the project: { agentMapFrame: true, agentMapSelectPressed: 'true', composer: 0 }
pageerrors: [] (one 404 for a resource, see "did not work")
```

The map planner's session does print text into the terminal — that is the
planner's own `SessionStart` hook, the shipped map-planner contract, visible in
the screenshot above. It is not `composerScaffoldPrompt`, which this PR's control
no longer reaches.

## The gap, stated

**The composer's own start still injects.** The composer is no longer a create
door — nothing in the rail opens it — but it remains what the app shows when
there is no session, no agent and nothing selected, and from there
`handleComposerSubmitIdea` still calls `sendScaffoldPrompt`.

The real path genuinely cannot serve it, and the reason is structural: the Agent
Map needs a project that already exists, and the composer's folder is invented
FROM the idea, after it. That inversion is exactly what the new verb's
folder-first step fixes, which is why the verb has no such case.

I did not disable the composer's send. Doing so would leave a fresh install's
home screen with no action at all, and moving new users onto the create verb's
folder step is a design decision I was not asked to make. Flagging it here rather
than silently keeping it: **this is the last `composerScaffoldPrompt` caller, it
is reachable from the home screen, and it should be someone's next ticket.**

## Spec changes (18 files)

Mechanical, through two new helpers in `mock-navigation.ts`:

- `openAddExistingAgents` — the item moved into the ⋮ menu (10 call sites).
- `openFolderStep` — the header `+` opens the where-step, not a folder dialog
  (5 call sites).

Judgement calls, each with the reason written into the spec:

- **`new-session-composer.spec.ts` (16 tests)** opens the composer via
  `?mockState=fresh` — the empty-state home — instead of the deleted CTA. Every
  assertion is unchanged; the subject was always the composer's own behaviour.
  One test, "Back returns to the session the composer was opened over", asserted
  the inverse of what is now true and became "the home screen has nothing to go
  Back to".
- **`smoke.spec.ts`'s CTA test** now asserts the verb is NOT in the nav and that
  pressing it asks where first.
- **`project-axis.spec.ts`'s header chrome** — the `+` is still left of the ⋮ and
  the label still owns the leading edge; what changed is what the `+` means.
- **`dismiss.spec.ts`** — Escape now returns focus to the ⋮, because the item
  that opened the dialog unmounts with the menu. A `triggerRef` pointing at a
  detached node restores focus to `<body>`, which is the bug that spec exists to
  catch; the fix is in the rail, not the spec.
- **`create-agent.spec.ts`** and `folder-field.spec.ts`'s desktop describe route
  through the ⋮, because the create verb's folder step CONTINUES into creation
  once the folder is named and those tests need the folder registered and nothing
  else.

## Two fixes this exposed

- `handleComposerSubmitIdea` now sets `composing` itself. It used to be set by
  the control that opened the composer, and holding it true is what keeps the
  composer mounted across the provisional session a start creates and a failed
  upload closes. Without it the attachment queue a rollback is supposed to retain
  went with the unmount.
- The empty-rail brand halo moved from `.rail-nav-cta.is-empty` to
  `.rail-header-btn.is-empty`, so the one action a first-run user needs is still
  the loudest thing on an empty rail. The dead `.rail-nav-cta` block is deleted;
  nothing in the modal block of `styles.css` is touched.

## Round 1

Findings 1, 2, 3 and 5 are fixed. Finding 4 is answered below with a citation
rather than a change.

**1 — the payload had no test.** `web/e2e/create-verb.spec.ts` (5 tests) asserts
where each branch LANDS, not that a menu opened: the Agent Map for a project that
has one, the create dialog **naming that project** for a payload that does not,
and the web folder step continuing into the flow instead of stopping at
registering the folder. The name is the assertion, because the named failure is
falling through silently with the wrong folder's basename.

Mutation-verified: forcing `scope` to null fails the map and folder-step tests;
stubbing the ⌘K guard fails its test; deleting the toast fails the failure test.

**A trap worth recording, because it nearly made this coverage a lie.** Mutating
a source file and THEN starting a fresh Vite dev server serves a CACHED
transform — the mutation never reaches the bundle and every spec passes, which
reads exactly like "the assertion survived its own mutation". Four consecutive
runs told me the map test was worthless. The file has to change while the server
is already running so HMR invalidates it.

**2 — it swallowed every failure.** Both call sites are `void onNewAgentIn(...)`
from a control that has already closed its popover, so a rejection left the
verb doing nothing at all, with an unhandled rejection as the only trace. It
reports now. The injection point is the workspace RE-READ, not the settings
write: `rememberProjectDir` swallows its own 500 by design
(`use-harness-state.ts:1503-1513`), so `openProject` does not reject and a test
built there would have asserted nothing.

**3 — it opened the project twice on the web fallback.** `openProjectAndResolve`
is now the one prefix shared by the verb and the rail's add-a-project route; the
difference stays in the callers. Two settings writes and two registry sweeps per
click, gone.

**4 — the two branches: the premise is wrong on the happy path, and the diff
cannot show why.** Selecting a project's map STARTS the creation flow one render
later. `use-agent-map-entry.ts:400-423` is an effect keyed on `projectId`:

```ts
if (startedProjectRef.current === projectId) return;
startedProjectRef.current = projectId;
...
if (!selectedResponse) loadPlanner(projectId, "resume-or-create");
```

`projectId` is `plannerProjectId` (`App.tsx:311-312`), read straight off
`studioSelection.kind === "agent-map"` — the one thing the handler sets. So the
map-planner session is resumed or created, and that planner is the surface where
an agent gets designed and generated. The screenshot above is that: the planner's
own SessionStart text, "Start by describing the outcome you want."

The create-verb spec now pins this rather than leaving it to a screenshot — it
waits for `session:/Users/demo/acme-app` in `createOrder` and asserts no
`scaffold:` entry. My first version asserted `createOrder` was EMPTY and was
simply wrong: landing on the map legitimately creates a session.

What genuinely differs between the branches is the KIND of surface, and the
payload forces that: there is no map to open when the catalog is absent. It is
the same asymmetry #790 shipped for the project-row doors. Forcing
`openFreshPlanner` instead was considered and rejected twice over — it reads
`currentProjectRef.current`, assigned during render, so it cannot be called in
the same tick as `setStudioSelection`; and resuming is the right behaviour when
the project already has a planning conversation in progress.

**5 — the changeset** is release copy now: what changed for someone using
Studio, no ticket IDs, no internal mechanisms.

## Also folded in: ⌘K stacked the palette over an open dialog

Found by the dialog-shell work (#800), which correctly stopped at `App.tsx`
rather than reaching into it. The pane-collapse hotkey 14 lines above the ⌘K
handler already bails when a layer is open; ⌘K did not, so the palette opened on
top of a dialog and native Tab then walked out of the palette into the dialog
behind it.

`.modal-backdrop` leads the selector because **`CommandPalette` carries no
`role` at all** — a role-only selector, which is the shape of the sibling
hotkey's, cannot see it and a guard copied from there would look correct and
detect nothing.

**The Overview is carved out, and the first version of this guard broke it.**
The palette is deliberately reachable by shortcut while the Overview is up, and
navigating from it dismisses the Overview rather than stacking behind it —
`welcome.spec.ts`'s "the palette's Browse templates, opened over the Overview,
leaves it (never stacks)" is that contract, and it failed until the carve-out
went in. The Overview wears both `role="dialog"` and `aria-modal="true"`, so the
exclusion is per clause.

This is also why the constant is NOT #800's `DIALOG_LAYER_SELECTOR` and should
not be collapsed into it. That one answers "which layer owns Tab", where the
Overview IS a layer and belongs in the list. This one answers "may ⌘K open
here", where it does not. Same shape, different question.

Scope, per #800's round 2: making the palette CONTAIN focus is pre-existing and
not attempted here.

## Held: two specs from #802 that this PR's IA orphans

#802 (`keep new-agent prompts in builder sessions`) landed while this was open
and added two specs that open the composer with `rail-create-new` on a fixture
that HAS projects. On this branch that control is the create verb: it opens the
"where does it live" step, and creation goes to the project's Agent Map. The
composer survives only as the home screen, which is by definition the state with
no session, no agent and nothing selected — so it cannot appear in a fixture
whose whole premise is that a parent project already exists. **Measured: the
composer renders 0 times at both of their URLs.**

The behaviour is intact — #802's `App.tsx` changes merged cleanly and the
standalone-builder intent is still here. What went is the door. Both specs are
held with `test.fixme` and a comment naming the collision, and the question is
escalated: whether the composer keeps an entry point is a product decision on
top of two PRs that landed hours apart, not a test edit. Nothing was deleted and
no assertion was changed.

This is the same question as the gap below — the composer's own submit is the
last `sendScaffoldPrompt` caller.

## Tests

- `pnpm test:ui` — **502 passed, 0 failed.**
- `create-verb.spec.ts` — 5 new e2e tests, mutation-verified (above).
- `folder-step.test.ts` — 5 new unit tests, mutation-verified (above).
- `pnpm typecheck` clean. `pnpm lint` is `eslint src` and does not reach
  `web/src`, so it says nothing about this diff.

### DIFF
--- a/.changeset/one-create-verb.md
+++ b/.changeset/one-create-verb.md
@@ -2,4 +2,4 @@
 "@sapiom/harness": minor
 ---
 
-Collapse the rail's three folder-or-create controls into ONE create verb on the Projects header, whose first step is where the agent lives: a project you already have, or a folder you pick. On desktop that folder step opens the OS folder browser directly instead of a dialog wrapping a text field. Creation then goes to that project's Agent Map, the same route project rows already take, so the rail's most prominent control no longer reaches the English scaffold prompt injected into the agent terminal. Registering a folder that already holds agents moves into the rail's settings menu.
+One `+` on the rail's Projects header creates an agent, and its first step asks which project or folder it belongs to. It replaces the "Create new agent" button and the header's separate "add a project" control. On the desktop app that folder step opens the system folder browser instead of a dialog with a path field. Registering a folder that already holds agents moved into the rail's ⋮ menu.
--- a/.changeset/one-create-verb.md
+++ b/.changeset/one-create-verb.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": minor
+---
+
+Collapse the rail's three folder-or-create controls into ONE create verb on the Projects header, whose first step is where the agent lives: a project you already have, or a folder you pick. On desktop that folder step opens the OS folder browser directly instead of a dialog wrapping a text field. Creation then goes to that project's Agent Map, the same route project rows already take, so the rail's most prominent control no longer reaches the English scaffold prompt injected into the agent terminal. Registering a folder that already holds agents moves into the rail's settings menu.
--- a/.changeset/palette-over-dialog.md
+++ b/.changeset/palette-over-dialog.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/harness": patch
+---
+
+The command palette no longer opens on top of an open dialog. Pressing its shortcut while a dialog is up used to stack the palette over it, and pressing Tab then walked out of the palette into the dialog behind. The shortcut now does nothing while a dialog is open, and is not handed to the browser. It still opens the palette over the Overview, which is unchanged.
--- a/packages/harness/web/e2e/accumulation-guard.spec.ts
+++ b/packages/harness/web/e2e/accumulation-guard.spec.ts
@@ -22,7 +22,7 @@
 import { expect, test } from "@playwright/test";
 import type { Page } from "@playwright/test";
 
-import { openProjectMenu } from "./mock-navigation";
+import { openProjectMenu, openAddExistingAgents } from "./mock-navigation";
 
 const ACME = "/Users/demo/acme-app";
 
@@ -250,7 +250,7 @@ test.describe("Remove project", () => {
     await page.getByTestId("remove-project-confirm-btn").click();
     await expect(page.getByTestId("workspace-group-acme-app")).toHaveCount(0);
 
-    await page.getByTestId("add-existing-agents").click();
+    await openAddExistingAgents(page);
     await page.getByTestId("folder-field-input").fill(`${ACME}/leasing`);
     await page.getByTestId("aw-add").click();
 
--- a/packages/harness/web/e2e/consent.spec.ts
+++ b/packages/harness/web/e2e/consent.spec.ts
@@ -196,8 +196,10 @@ test.describe("UI event tracking (track() calls)", () => {
   });
 
   test("new session creation emits track('session.created')", async ({ page }) => {
-    // Creating a new agent starts a session from the composer.
-    await page.getByTestId("rail-create-new").click();
+    // The composer is the app's home screen — what it shows with no session, no
+    // agent and nothing selected — rather than something the rail opens. The
+    // rail's create verb goes to a project's Agent Map instead.
+    await page.goto("/?mockState=fresh&mockStudioProjects=absent");
     await expect(page.getByTestId("new-session-composer")).toBeVisible();
     await page.getByTestId("composer-input").fill("Summarize new issues each morning.");
     await page.getByTestId("composer-send").click();
--- a/packages/harness/web/e2e/create-agent.spec.ts
+++ b/packages/harness/web/e2e/create-agent.spec.ts
@@ -26,7 +26,7 @@
 import { expect, test } from "@playwright/test";
 import type { Page } from "@playwright/test";
 
-import { openProjectMenu } from "./mock-navigation";
+import { openAddExistingAgents, openProjectMenu } from "./mock-navigation";
 
 const ROOT = "/Users/demo/acme-app";
 
@@ -202,7 +202,11 @@ test.describe("legacy-server agent creation compatibility", () => {
   test("the empty-project row opens the same dialog", async ({ page }) => {
     // One create flow, not one per door: the empty project's CTA is the same
     // subject and must not be a second mechanism that drifts.
-    await page.getByTestId("rail-add-project").click();
+    // Through the ⋮, not the create verb: the verb's folder step CONTINUES
+    // into creation once the folder is named, which is the whole point of it
+    // asking where first. This test needs the folder registered and nothing
+    // else, so it asks the other question.
+    await openAddExistingAgents(page);
     await page
       .getByTestId("folder-field-input")
       .fill("/Users/demo/blank-slate");
--- a/packages/harness/web/e2e/create-verb.spec.ts
+++ b/packages/harness/web/e2e/create-verb.spec.ts
@@ -191,4 +191,54 @@ test.describe("⌘K over an open dialog", () => {
     await page.keyboard.press("ControlOrMeta+k");
     await expect(page.getByTestId("command-palette-input")).toBeVisible();
   });
+
+  test("⌘P over a dialog is swallowed, not handed to the browser", async ({
+    page,
+  }) => {
+    // THE SAME HANDLER OWNS ⌘P, and ⌘P is the browser's PRINT shortcut. A guard
+    // that returns early WITHOUT preventing the event stops the palette and
+    // hands ⌘P to the browser, which opens a native print preview over the
+    // dialog — worse than the stacking it was added to stop, and invisible to a
+    // spec that only presses ⌘K.
+    //
+    // `defaultPrevented` IS THE ASSERTION, not `beforeprint`. Headless Chromium
+    // never fires `beforeprint` for an unhandled Ctrl+P, so a spec written that
+    // way passes whether or not the event was prevented — measured: it survived
+    // its own mutation. This listener is registered after the app's (same
+    // target, `window`, so registration order decides), and reads what the app
+    // left behind.
+    await page.goto("/?seed=0&mockStudioProjects=absent");
+    await expect(page.locator(".rail-workflows")).toBeVisible();
+    await page.evaluate(() => {
+      const win = window as unknown as { __printKey?: boolean | null };
+      win.__printKey = null;
+      window.addEventListener("keydown", (event) => {
+        if (!(event.metaKey || event.ctrlKey)) return;
+        if (event.key.toLowerCase() !== "p") return;
+        // READ AFTER DISPATCH, not during. The app re-subscribes its handler on
+        // re-render, so a listener added from a spec is not reliably last and
+        // reading `defaultPrevented` inline can sample before the app has acted.
+        // The flag stays set on the event object once dispatch completes.
+        setTimeout(() => {
+          win.__printKey = event.defaultPrevented;
+        }, 0);
+      });
+    });
+
+    await page.getByTestId("rail-create-new").click();
+    await page.getByTestId("new-agent-in-acme-app").click();
+    await expect(page.getByTestId("create-agent-dialog")).toBeVisible();
+
+    await page.keyboard.press("ControlOrMeta+p");
+    await expect(page.getByTestId("command-palette-input")).toHaveCount(0);
+    await expect(page.getByTestId("create-agent-dialog")).toBeVisible();
+    // Seen by the listener, and prevented by the app during dispatch.
+    await expect
+      .poll(() =>
+        page.evaluate(
+          () => (window as unknown as { __printKey?: boolean | null }).__printKey,
+        ),
+      )
+      .toBe(true);
+  });
 });
--- a/packages/harness/web/e2e/create-verb.spec.ts
+++ b/packages/harness/web/e2e/create-verb.spec.ts
@@ -47,6 +47,79 @@ const lastInjectText = (page: Page): Promise<string> =>
   );
 
 test.describe("the create verb lands somewhere", () => {
+  test("both entrances land on the new agent screen, scoped to the project", async ({
+    page,
+  }) => {
+    // THE SWAP THIS PR EXISTS FOR, asserted on both entrances.
+    //
+    // The screen with the good surface had the wrong plumbing (it invented a
+    // folder from a slug and typed English at the coding agent); the path with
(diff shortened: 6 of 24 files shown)


gwitwer:
Round 2 withdrew finding 4 already, but recording the citation here so it is findable rather than only in a review thread.

**Selecting the map DOES start the creation flow** — one render later, which is why the diff cannot show it. `use-agent-map-entry.ts:400-423` is an effect keyed on `projectId`:

```ts
if (startedProjectRef.current === projectId) return;
startedProjectRef.current = projectId;
...
if (!selectedResponse) loadPlanner(projectId, "resume-or-create");
```

`projectId` is `plannerProjectId` (`App.tsx:311-312`), read straight off `studioSelection.kind === "agent-map"` — the one thing `handleNewAgentIn` sets on that branch. So the map-planner session is resumed or created, and that planner is the surface where an agent gets designed and generated.

Real-server evidence, `~/sapiom/agents`: the screenshot in the PR body is the planner's own SessionStart text, "Agent Map planning session … Start by describing the outcome you want." Measured at the same moment: `{ agentMapFrame: true, agentMapSelectPressed: 'true', composer: 0 }`.

`create-verb.spec.ts` now pins this rather than leaving it to a screenshot: it waits for `session:/Users/demo/acme-app` in `createOrder` and asserts no `scaffold:` entry. My first version asserted `createOrder` was EMPTY and was simply wrong — landing on the map legitimately creates a session.

**On the asymmetry itself**, which is the part worth keeping: what differs between the branches is the KIND of surface, and the payload forces that. There is no map to open when the Studio project catalog is absent, so the fallback is the name-plus-template dialog. It is the same asymmetry #790 already shipped for the project-row doors, and it is not a property of the verb.

Forcing `openFreshPlanner` instead was considered and rejected on two grounds. It reads `currentProjectRef.current`, assigned during render (`use-agent-map-entry.ts:167`), so it cannot be called in the same tick as `setStudioSelection` — it would need a pending-fresh-planner ref consumed by an effect, new machinery inside that hook. And it is arguably the wrong behaviour: resuming is what you want when the project already has a planning conversation in progress, where forcing fresh would drop the user into an empty planner and strand the one they were mid-way through.

Settled with the founder as `resume-or-create`, unchanged. [SAP-3143]

ynadge (approved):

gwitwer:
Closing. This PR implements a brief that was reversed after it was built, and its core mechanism no longer exists on `main`.

**1. The dispatch this PR exists for was deliberately removed from main.**

`c6414924` ("submitting in a project dispatches the planner, not English" — its own message calls this "the swap this PR exists for") calls `harness.openPlannerSession(...)`. Neither `openPlannerSession` nor `src/core/planner-greeting.ts` exists on `main`: `planner-greeting.ts` was added by #765 on 2026-09-01 and deleted on 2026-09-05 by `21684912` "feat(harness): activate **ordinary** bootstrap [Agent Map 07/15]".

That deletion is the founder's reversal, already landed: plan-first sessions are ordinary sessions. SAP-3143 now states it directly — "Submit opens a NORMAL session … **Not** the planner / control-turn session type. That session type is explicitly not to be reinforced." So this is not a merge conflict against main; it is a call into machinery main intentionally dropped.

**2. SAP-3143 disowns this diff.** The ticket's current text opens: "PR #803 as built implements an earlier brief that was reversed twice," and names four defects verified in the diff — the prominent CTA became a 14px `rail-header-btn` with `aria-haspopup="menu"`, the create screen stopped being a destination, four plus-type controls still render, and the popover does not match the design language.

**3. The base branch is a dead end.** Base is `gwitwer/studio-rail-creation`, whose PR #801 was closed unmerged. SAP-3142's content landed instead via #819 + `[Agent Map 01..15/15]` (merged 2026-09-06); main's `src/shared/project-roots.ts` is a superset of the abandoned branch's and adds the test it never had.

For the record, the reported `CONFLICTING`/`DIRTY` state was **stale**: the base tip (`fbb6cdf4`) is an ancestor of the head (`bb133c20`), 0 ahead / 13 behind, and a test-merge of head into base was clean. The real conflicts only ever existed against `main`.

**What is being carried forward:** `web/src/lib/folder-step.ts` + its test — the native-picker-on-desktop folder step, which is still an acceptance item on SAP-3143 and is the one piece with no open design question. It moves to a fresh PR from `main`.

**Not carried forward:** the planner dispatch (above); the create verb as built (the four defects); the `⌘P` fix in `f9112731`, which repairs `PALETTE_BLOCKING_LAYER_SELECTOR` — a guard added two commits earlier *in this same PR*, so it fixes a branch-introduced bug with no counterpart on main; and the `styles.css` deletion of `.rail-nav-cta`, which removes the prominent CTA the current brief asks to keep.

Note that `composerScaffoldPrompt` — the English injection SAP-2981/SAP-3143 exist to remove — is **still live on main** (`web/src/App.tsx:1888`). That remains the open work, and it is not what this PR does.

The approval from @ynadge stands against a diff built on the pre-reversal brief, so it is not carried over.

PR #815, opened 2026-09-05
docs(mcp,harness,agent-core): sync offline fallbacks with the vault/launch/receipts/webhooks content release [SAP-3180]

gwitwer:
## Primary change type

- [x] Documentation

## Problem and motivation

Four shipped behaviours had no served teaching text at all: Vault semantics, `ctx.sapiom.agents.launch`, receipts and replay of inbound events, and App Link webhooks — so a Studio session could misuse or miss them. The backend fixes that with content releases 2.9 (authoring primer) and 1.1 (Studio system prompt) in sapiom/Sapiom PR for SAP-3180; this PR is the paired half that keeps the offline fallbacks byte-identical, so a session whose startup fetch fails learns the same things.

## Summary and scope

- `@sapiom/mcp` `AUTHORING_INSTRUCTIONS` and `@sapiom/harness` `DEFAULT_SYSTEM_PROMPT` copied verbatim from the new backend constants; both digest pins moved to the same values the backend specs pin (`ffb497ca…` and `d83fa57c…`).
- The 2.9 primer also drops the `deadlineMinutes` clause (`llm.run` has no such knob), and the `/v1/workflows/` guard in `instructions.test.ts` is scoped to `/v1/workflows/executions/` because the primer now points at the receipts/replay routes under that prefix.
- `sapiom-agent-authoring` skill gains a pointer to where these are taught (primer + docs pages), propagated to both template copies and the plugin copy. No restatement.
- Out of scope: database lifecycle wording (SAP-3100), trigger kinds in `sapiom_dev_agents_schedule` (SAP-3174).

Note: #814 (SAP-3179) edits the same two fallbacks. Whichever lands second must re-copy the backend body and recompute both pins.

Merged `main` (Agent Map 15/15, `e0c1f47a`) into this branch: the bundled prompt is regenerated from the backend 1.1 constant, which now carries main's Agent Map wording plus the SAP-3180 paragraph, so the cross-repo pin holds again (`d83fa57c…`). The legacy-verbatim test is narrowed to the feedback and workspace blocks, and the LLM paragraph's claims are asserted sentence by sentence, because this release deliberately extends that paragraph.

## Related work

Related issue or discussion: https://linear.app/sapiom/issue/SAP-3180 (backend half: sapiom/Sapiom PR titled `docs(mcp,harness): teach Vault semantics, agents.launch, receipts/replay, App Link webhooks [SAP-3180]`)

## Validation

```text
[...]

PR #814, opened 2026-09-05
fix(mcp,harness): name the two MCP servers by role in both offline fallbacks [SAP-3179]

gwitwer:
## Primary change type

- [x] Bug fix
- [ ] Documentation
- [ ] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

The two offline fallbacks — `AUTHORING_INSTRUCTIONS` in `@sapiom/mcp` and `DEFAULT_SYSTEM_PROMPT` in `@sapiom/harness` — named the same two Sapiom MCP servers differently. The Studio prompt said `sapiom` (hosted) / `sapiom-dev` (local), which is what `mcp-config.ts` registers. The authoring primer said `sapiom` (local) / `sapiom-direct` (hosted), which is what a plain Claude Code user is told to register. A Studio session reads both, so "use the `sapiom` alias to author agents" pointed it at the remote server the prompt had just said not to call while authoring.

## Summary and scope

Both bodies are now byte-identical copies of the backend's 2.9 authoring primer and 1.1 Agent Studio system prompt, which describe the servers by role ("the local authoring server", "the hosted capability server") and mention an alias only inside the unchanged `claude mcp add` commands. Both digest pins (`instructions.test.ts`, `default.test.ts`) move to the new bodies, and the alias assertions now check the role phrases.

Out of scope: `claude-code-info.ts` tells a user installing outside Studio to register the local server as `sapiom-dev`, while the primer's command says `sapiom`. That is a separate install flow, noted on the Sapiom ticket.

## Related work

Related issue or discussion: SAP-3179 (https://linear.app/sapiom/issue/SAP-3179). Paired backend PR in sapiom/Sapiom — linked in the first comment. Land both together; the cross-repo digest tests redden either repo alone.

## Validation

```text
pnpm --filter @sapiom/mcp exec vitest run src/instructions.test.ts — 8 passed
pnpm --filter @sapiom/harness exec vitest run src/profiles/default.test.ts — 2 passed
pnpm exec prettier --check <changed files> — clean
```

### Tests and documentation

Updated the two pin tests: new sha-256 digests, and role-phrase assertions replacing the alias-identity assertions. No user-facing docs in this repo change; the served text is the documentation.

## Compatibility and release impact

- Breaking or externally visible changes: None. The fallback text is served only when the startup fetch of the live copy fails.
- Changeset: Added (`@sapiom/mcp` patch, `@sapiom/harness` patch).

## Security

[...]

PR #817, opened 2026-09-05
feat(mcp,agent-core): expose event and webhook trigger kinds from sapiom_dev_agents_schedule [SAP-3174]

gwitwer:
## Primary change type

- [ ] Bug fix
- [ ] Documentation
- [x] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

The backend accepts four trigger kinds (`schedule_cron`, `schedule_once`, `event`, `webhook`) and the public webhook and events ingress routes are live, but `sapiom_dev_agents_schedule` offered `z.enum(["schedule_cron", "schedule_once"])` and neither the served primer nor the authoring skill mentioned the other two. Observed 2026-09-04: a Studio session asked to build an inbound WhatsApp receiver concluded nothing listens for that and proposed hand-building an always-on HTTP endpoint.

## Summary and scope

- `sapiom_dev_agents_schedule` accepts `kind: "event"` (+ `eventType`) and `kind: "webhook"`, mirroring the backend DTO. A webhook create returns the public hook URL, the shown-once secret, and the signing scheme in the tool result (HMAC-SHA256 over `timestamp.eventId.rawBody`, sent as `X-Sapiom-Timestamp` / `X-Sapiom-Event-Id` / `X-Sapiom-Signature`). The description says when a webhook trigger fits versus an App Link `/hook/*` receiver (third-party senders cannot produce our HMAC).
- `_schedule_inspect` and `_schedule_cancel` describe every kind; inspect never carries a secret. New `sapiom_dev_agents_schedule_secret` (`rotate` / `complete_rotation` / `revoke`) makes SAP-2835 reachable without dropping to REST.
- `@sapiom/agent-core`: new `ScheduleKind` members, webhook/event fields on `ScheduleSummary`, `CreateScheduleResult`, and `rotateScheduleSecret` / `completeScheduleSecretRotation` / `revokeScheduleSecret`.
- Offline `AUTHORING_INSTRUCTIONS` re-synced to the server's 2.10 primer; both digest pins move to `7501423911e2…`. The `sapiom-agent-authoring` skill (all four copies) gains a Triggers section.
- Both text changes are version-gated on `@sapiom/mcp` >= 0.15, which the minor changeset makes the next release.

Out of scope: `@sapiom/tools` `schedules` `ScheduleKind` (SAP-3181 owns the tools-side restatements); the harness system prompt (SAP-3179/3180).

Companion PRs: sapiom/Sapiom#4888 (the server-side content release this fallback copies) and sapiom/docs-internal#152 (`guides/triggers`).

**Merge order:** docs-internal#152 first (the primer, the skill, and the tool copy link `https://docs.sapiom.ai/guides/triggers`), then #4888 and this PR, then the "chore: version packages" PR that publishes `@sapiom/mcp` 0.15.0 / `@sapiom/agent-core` 0.14.0.

## Related work

Related issue or discussion: SAP-3174 (https://linear.app/sapiom/issue/SAP-3174)

## Validation

```text
pnpm build                                                     — all packages built
pnpm typecheck                                                 — clean
pnpm lint                                                      — 0 errors (pre-existing warnings in untouched packages)
pnpm --filter @sapiom/mcp test                                 — 15 files, 153 tests passed
pnpm --filter @sapiom/agent-core exec jest src/__tests__/skill-sync.test.ts src/schedule.spec.ts — 26 passed
pnpm --filter @sapiom/harness exec vitest run src/core/inject/skills-plugin.test.ts — 5 passed
```

https://claude.ai/code/session_01RA1XLguY2Pgzsg2UDBXQAj


gwitwer:
Replying to review round 1 (sha 398a195):

**`/v1/workflows/events` in `schedule.ts` JSDoc, the tool description, `eventType.describe()`, and the event hint — keeping it.** The route is not confidential: it is the tenant-facing REST route the docs publish. docs.sapiom.ai already prints `/v1/workflows/executions/:id/signals` (guides/use-signals) and `/v1/workflows/executions/:id/steps/:stepId/io` (guides/choose-a-call-surface), `client.ts` publishes `/v1/workflows` as the base, and sapiom/docs-internal#152 documents `POST /v1/workflows/events` with a curl example. The terminology check that 398a195 satisfied guards the human-readable *product vocabulary* in Studio-facing prose ("Workflow" as a noun for what we now call an agent), which is why its `STATIC_TARGETS` are skills and templates and not tool schemas; the skill wording was changed to pass that guard, not to hide the path. The tool description and hint are what the agent uses to configure a sender, so they carry the real route. The two surfaces do not contradict each other: the skill points at the guide, the guide states the route.

**Breaking type changes under `minor` — fixed in c8a1051.** The changeset now carries a **Breaking (types only)** paragraph: `ScheduleFireRecord.scheduledFor` is `string | null` (event and webhook fires have no occurrence time, so the narrower type was wrong for them), `ScheduleSummary` gains five required nullable fields the server always returns, and `ScheduleKind` widens; each with its migration. I kept the fields required rather than optional because the backend response always has them, and an optional type would force every reader to handle an absence that never happens.

**`https://docs.sapiom.ai/guides/triggers` before the page is live — merge ordering, noted in the PR body.** The page is sapiom/docs-internal#152, CI green. The link ships in three artifacts: this package, the server-served primer (sapiom/Sapiom#4888, which also registers the route in the repo's canonical docs-route list), and the docs site. Release order: docs-internal#152 first, then #4888 and this PR, then the "chore: version packages" PR that actually publishes 0.15.0. That is the same sequence `/capabilities/app-links` followed for 2.7/2.8 and `@sapiom/mcp` 0.13. Dropping the link would leave pre-0.15 clients with no pointer to the REST fallback, which is the worse failure.

packages/mcp/src/tools/agents.ts
```diff
@@ -729,6 +798,59 @@ export function register(server: McpServer, env: ResolvedEnvironment): void {
...
+          );
+          return ok({
+            schedule,
+            webhook: {
+              url,
+              secret,
+              secretVersion: schedule.secretVersion,
+              signing: WEBHOOK_SIGNING_SCHEME,
+            },
+            hint: `Rotated to secret v${schedule.secretVersion}; shown ONCE. The previous secret still verifies until ${schedule.graceUntil ?? "the grace ends"} — reconfigure the sender, then call this tool with action 'complete_rotation' to close the window early.`,
```
antoine-berger: When the grace is still open the engine replays the live rotation but this hint still says "Rotated to secret vN; shown ONCE". An agent reading that will believe a new secret was minted and reconfigure the sender for nothing. Please cover the replay case, e.g. "if `secretVersion` is unchanged, this replayed the current rotation; call `complete_rotation` first to rotate again".

packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
```diff
@@ -590,6 +593,39 @@ Under `run_local`, a dispatch pause auto-resumes with the stub result; a manual
...
 type the resumed step's input with optional fields accordingly.
 
+## Triggers — Run a Deployed Agent Without a Human
+
+A trigger is a persisted cloud object attached to a **deployed** agent by slug; each fire starts an
+independent production run. Create one with `sapiom_dev_agents_schedule` — `kind` picks the shape:
+
+| `kind`          | Fires when                                                                                                                    | Required field                                          |
+| --------------- | ----------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------- |
+| `schedule_cron` | on a recurring cron (validate it first with `sapiom_dev_agents_cron_preview`)                                                 | `cron` (+ `timezone`)                                   |
```
antoine-berger: `timezone` isn't required for `schedule_cron`, only `cron` is.

antoine-berger (changes requested):
Checked the routes, response shapes, signing block and fallback digest against the server, all match. Nice work. Two fixes inline before merge.

packages/mcp/src/tools/agents.ts
gwitwer: Fixed in 85678ae. The tool now reads the trigger's `secretVersion` before rotating and compares it with the rotate response. When the engine replayed the live rotation (same version), the hint reads "No new secret: a rotation to vN is already in progress (grace until …), so this replayed it and returned the SAME current secret … to rotate again, call this tool with action 'complete_rotation' first", and the result carries `replayed: true` so callers need not parse the hint. A fresh rotation keeps the previous wording. Test added for the replay case.

packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
gwitwer: Fixed in 85678ae: the kinds table now says `cron` (optional `timezone`), in all four skill copies, and the tool description was carrying the same overstatement ("+ cron + timezone"), so it now reads "+ cron, optional timezone" too.

antoine-berger (dismissed):

gwitwer:
Heads-up on overlap: #814 (SAP-3179) also moves `packages/mcp/src/instructions.ts` and its digest pin. Its body is the backend's 2.11 primer, which carries this PR's 2.10 trigger-kinds section unchanged plus the role-based server names, so the two are a superset/subset rather than a conflict of intent. Whichever lands second should re-sync `AUTHORING_INSTRUCTIONS` and the pin to the backend's current `DEFAULT_MCP_INSTRUCTIONS`; backend PR sapiom/Sapiom#4884 has the 2.11 digest.

gwitwer:
**Merge conflict resolved (51c5fff, merge of `main`, no force push).** The conflicts were the offline `AUTHORING_INSTRUCTIONS` body and its digest pin. #815 had moved this copy to a 2.9 body whose server-side release, sapiom/Sapiom#4885, has not merged and is itself conflicting against Sapiom `main`. Sapiom `main` serves the 2.10 body (SAP-3174 merged as #4888, digest `47a4e3a3…`), and the rule this pin enforces is that the two copies are one canonical text. So the fallback resolves to the served 2.10 body and the 47a4e3a3 pin.

The 2.9-only assertions from #815 (Vault, `agents.launch`, receipts/replay, App Link webhook forwarding, and the dropped "Say how long you can wait" clause) are kept but `it.skip`ped with the reason in a comment. They come back when #4885 lands — rebased onto 2.10 it becomes 2.11 and carries both sections — and this copy is re-synced to that body. Left a note on #4885.

154 tests pass, 1 skipped; lint and terminology check clean.

antoine-berger (dismissed):

PR #821, opened 2026-09-05
feat(tools): export the llm session resume payload type and schema from the root [SAP-3184]

gwitwer:
## Primary change type

- [ ] Bug fix
- [ ] Documentation
- [x] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

The two `llm` deferred lanes are asymmetric at the `@sapiom/tools` boundary. The **async** lane exports everything an author needs to type and validate a resumed step's input from the package root (`LlmRouteResultPayload`, `llmRouteResultSchema`, `LlmRouteResultSchemaError`). The **session** lane exports none of it: a step paused on `llm.createSession(...)` and resumed on `LLM_SESSION_READY_SIGNAL` has no root-exported type for its input and no validator, so anyone using it writes their own types and their own validation (or reaches for the `./llm` subpath).

## Summary and scope

Additive only. From `packages/tools/src/llm/index.ts`:

- `LlmSessionReadyPayload` — the input a step resumed from `pauseUntilSignal(sessionHandle, …)` receives. An `LlmSession` narrowed to the two shapes the engine's resume forwarder delivers: `state: "ready"` (hand it to `callSession`; `baseUrls` carries the session-scoped URLs when reported), or `state: "failed"` with a non-empty `error` carrying the gateway's structured reason (`deadline_exhausted`, `grant_mint_failed`, `session_ready_failed`, `session_unsupported`). The signal fires once, when the session leaves `pending`, and the forwarder folds any non-ready outcome into `failed`, so no other state arrives on it.
- `llmSessionReadySchema` / `LlmSessionReadySchemaError` — runtime validator + typed error, same shape of API as `llmRouteResultSchema` / `LlmRouteResultSchemaError`. `baseUrls` is optional (as on `LlmSession`) and validated as a complete pair only when present.

From `packages/tools/src/index.ts` (the barrel), new root exports:

- `LlmSessionReadyPayload`, `LlmSession`, `LlmSessionState` (types), `llmSessionReadySchema`, `LlmSessionReadySchemaError`
- Nice-to-haves from the ticket, same class of gap: `RoutingLabel`, `ModelLabel`, `LlmDisclosureResult` (types) and `readDisclosure`.

`scripts/provider-neutral-copy-check.mjs` gains one narrow mask for the const tuple that names the two base-URL keys the schema validates.

Out of scope: no server or gateway change; no change to any existing export's name or shape.

## Related work

Related issue or discussion: [SAP-3184](https://linear.app/sapiom/issue/SAP-3184/sdk-export-the-session-resume-payload-type-and-schema-from-the-package) (part of epic SAP-3183).

## Validation

```text
pnpm --filter @sapiom/tools typecheck                                  — clean
pnpm --filter @sapiom/tools lint                                       — clean
pnpm --filter @sapiom/tools build                                      — ok (cjs + esm)
pnpm provider-copy:check                                               — passed (129 audited files)
pnpm examples:check:test                                               — 165 tests, 0 failures
npx jest --maxWorkers=1 src/llm/sessions.spec.ts src/smoke.spec.ts    — 2 suites, 21 tests passed
```

### Tests and documentation

- `llm/sessions.spec.ts`: new `llmSessionReadySchema` block — accepts a well-formed ready payload with and without `baseUrls` (and the union narrows on `state`), accepts a failed payload for each of the four gateway terminal reasons, rejects malformed payloads with `LlmSessionReadySchemaError` (non-object, missing `sessionId`, non-terminal `state`, partial or non-object `baseUrls`, failed without a reason, mistyped optional fields).
- `smoke.spec.ts`: asserts both lanes' resume contracts (signal + payload type + schema + error) import from the root symmetrically, plus `readDisclosure`, `RoutingLabel`, `ModelLabel`.
- JSDoc on the new type/schema mirrors `LlmRouteResultPayload`'s and cross-links the two.

- Breaking or externally visible changes: None. New exports only.
- Changeset: Added (`@sapiom/tools`: minor).

## Security

      will follow the
      private reporting.

## AI assistance

Written with Claude Code. It derived the payload shape from what the engine's resume forwarder delivers to a resumed step, wrote the schema, tests and changeset, and verified with the commands above.

## Checklist

      any N/A checks above.

https://claude.ai/code/session_01W5yTA5mh9qL5P6e3oBc6GH


gwitwer:
Thanks — addressed in 14e6bb3e and the edited PR body.

**Confidentiality.** Fixed. The PR body and AI-assistance note now use the same generic wording as the shipped copy ("the engine's resume forwarder") and name no internal files.

**Validator strictness.** Partly agreed, partly declined:

- `baseUrls` — agreed. It is now optional on `LlmSessionReadyPayload` (as it is on `LlmSession`) and validated as a complete, well-typed pair only when present. `callSession` needs only `sessionId`, so a ready payload without base URLs parses. Test added.
- State union — declined. `LLM_SESSION_READY_SIGNAL` fires exactly once, when the session leaves `pending`; the forwarder maps the webhook body, not a later GET, and folds any non-ready outcome into `state: "failed"` with the outcome as the reason — the same normalisation the async lane does in `toPayload`. So `ready | failed` is the complete set that arrives on this signal, and `active`/`expired`/`exhausted` are states you get from `getSession`/`wait()`, not from a resume. Accepting them here would advertise shapes that never arrive. I have put that guarantee in the JSDoc. `safeParse` is not added: it would diverge from `llmRouteResultSchema`, which this ticket is explicitly mirroring.

**Root exports.** `LlmDisclosure` (raw wire shape) dropped from the root; it stays on the `llm` namespace and subpath. `readDisclosure` stays: the ticket lists it by name as a root re-export, and `LlmDisclosureResult` stays as its return type.

gwitwer:
Thanks — addressed in 10831242.

**`LlmSession.baseUrls` JSDoc** — fixed. It now says the URLs are present once the session is READY and the gateway reports them, and that `callSession` does not need them. That matches the payload JSDoc, the changeset, and what the validator enforces.

**`safeParse`** — declined for this PR, deliberately. The ticket's brief is to mirror `llmRouteResultSchema` / `LlmRouteResultSchemaError` so the two lanes are symmetric at the root; giving only the session schema a non-throwing path would reintroduce an asymmetry between the lanes, and adding it to both is a separate, cross-lane change. Agreed it is worth doing — as a follow-up covering both schemas together.

**`readDisclosure`** — stays for the reason given in round 1: the ticket lists it by name as a root re-export.

ratataque:
The validator doesn’t fully enforce its return type:

- [`LlmSession` declares `budget` and `error`](https://github.com/sapiom/sapiom-js/blob/10831242e1412c678e7ff08cf11fbcea56df1f86/packages/tools/src/llm/index.ts#L630-L645).
- [`LlmSessionReadyPayload` inherits both fields](https://github.com/sapiom/sapiom-js/blob/10831242e1412c678e7ff08cf11fbcea56df1f86/packages/tools/src/llm/index.ts#L680-L682).
- [The validator](https://github.com/sapiom/sapiom-js/blob/10831242e1412c678e7ff08cf11fbcea56df1f86/packages/tools/src/llm/index.ts#L695-L726) never checks `budget`, and only checks `error` when `state === "failed"`. The final type assertion hides that mismatch.

For example, this currently passes despite violating the returned type:

```ts
llmSessionReadySchema.parse({
  sessionId: "s",
  state: "ready",
  budget: "invalid",
  error: 123,
});
```

I’d suggest replacing the manual checks with a declarative Zod schema and deriving the payload type from it. One definition for validation and typing should scale better as the contract grows and be much cleaner to maintain.

The replacement could look like this:

```ts
import { z } from "zod";

const sessionFields = {
  sessionId: z.string(),
  model: z.string().optional(),
  expiresAtMs: z.number().optional(),

  baseUrls: z.looseObject({
    anthropic: z.string(),
    openai: z.string(),
  }).optional(),

  budget: z.looseObject({
    maxTokens: z.number().nullable(),
    usedTokens: z.number().optional(),
    ttlMinutes: z.number().nullable().optional(),
  }).optional(),
};

const sessionReadyPayloadSchema = z.discriminatedUnion("state", [
  z.looseObject({
    ...sessionFields,
    state: z.literal("ready"),
    error: z.string().optional(),
  }),
  z.looseObject({
    ...sessionFields,
    state: z.literal("failed"),
    error: z.string().min(1),
  }),
]);

export type LlmSessionReadyPayload =
  z.infer<typeof sessionReadyPayloadSchema>;

export class LlmSessionReadySchemaError extends Error {}

export const llmSessionReadySchema = {
  parse(value: unknown): LlmSessionReadyPayload {
    const result = sessionReadyPayloadSchema.safeParse(value);

    if (!result.success) {
      throw new LlmSessionReadySchemaError(
        `invalid llm session ready payload: ${result.error.message}`,
      );
    }

    return result.data;
  },
};
```

This preserves the public `.parse()` entry point and error class. `safeParse` stays internal to the wrapper, so this does not add a new public parsing method. `looseObject` keeps additional fields.

Integration details: add Zod as a dependency, update tests that expect the original object by identity, and cover malformed budget/error values. Zod also rejects non-finite numbers. [Zod API reference](https://zod.dev/api).

Illustrative proposal; not applied or tested.

gwitwer:
Thanks @ratataque — good catch, and correct: the return type inherited `budget` and `error` from `LlmSession` while `parse` only checked `error` on the failed branch and never looked at `budget`, so the closing assertion was hiding a real gap.

**Fixed in 08ccb88a.** The validator now enforces every field the type declares: `error` must be a string whenever present (non-empty on `failed`), and `budget`, when present, must be an object with `maxTokens` as number-or-null, optional numeric `usedTokens`, and optional `ttlMinutes` as number-or-null. Your exact example (`budget: "invalid", error: 123` on a ready payload) now throws `LlmSessionReadySchemaError`, and there are tests for it and for each `budget` member, plus a well-formed budget with the nullable/optional members passing.

**On Zod — agreed on the direction, declined for this PR.** The package's other four resume-payload validators (`llmRouteResultSchema`, `modelRunResultSchema`, `agentResultSchema`, `codingResultSchema`) are all hand-rolled in this same style, and `@sapiom/tools` currently carries a single runtime dependency. Rewriting just this one in Zod would make it the odd one out and add a dependency to a published package for a single validator; migrating all five together (deriving each payload type from its schema, as you sketch) is the right unit, and it is a bigger change than this ticket's "mirror the async lane" brief. I'd support that as its own follow-up — happy to open the ticket if you want it tracked.

ratataque (approved):
as you suggested might be interesting to migrate all the validator with proper schema validation.

but looks good for now

PR #839, opened 2026-09-05
feat(tools): deadlineMinutes on the coding + model run specs, awaiting_capacity in RunStatus [SAP-3201]

gwitwer:
## What

Gives the SDK a typed way to say "I can wait N minutes" on a run, and teaches the run-status union about the deferred state the server will start returning.

- `deadlineMinutes?: number` on `CodingRunSpec` and `ModelRunSpec`, sent on the wire as `deadline_minutes` by both `codingLaunch`/`codingRun` and `models.run`/`models.launch`.
- `RunStatus` gains `awaiting_capacity` — added to the union and to `RUN_STATUSES` (so `codingResultSchema` accepts it), and deliberately **not** to `TERMINAL`.

## Why

Today every run dispatches immediately and is billed at the most expensive lane, because an author has no way to express slack. This is the author-facing half of the label + deadline vocabulary from `plans/model-execution-surface/interfaces.md`: the author states the kind of call (`model`) and how long they can wait (`deadlineMinutes`); the platform derives the billing lane (`run_now` / `priority` / `standard` / `flex`). The caller never names a lane.

## Two details worth checking in review

**Omitted means absent, not zero.** `deadline_minutes: spec.deadlineMinutes` leaves the value `undefined` when unset, so `JSON.stringify` drops the key entirely. The server has to distinguish "no deadline" from "zero minutes", and an existing caller's request body stays byte-identical to before this change. Covered by a test on each surface.

**`awaiting_capacity` is non-terminal.** `TERMINAL` stays `{"completed", "failed"}`. `run()`'s poll loop and a `launch()` handle's `wait()` only stop on those two, so a deferred run keeps polling instead of resolving with a null result. There's a test that drives a run through two `awaiting_capacity` polls into `completed` and asserts the loop kept going.

## Scope

`agents.run` (`AgentRunSpec`) is left alone. It dispatches a deployed orchestration rather than an LLM call, so it has no billing lane to derive — the ticket lists that as a nice-to-have "if it costs nothing", and it doesn't.

`ModelRunStatus` **does** gain `awaiting_capacity`, reversed from this PR's first revision after review. The original reasoning was that the ticket scopes the new status to `RunStatus` and the gateway doesn't emit the deferred state on the model-run surface. That was the wrong way round: `deadlineMinutes` ships on `ModelRunSpec` (an explicit acceptance criterion), and the only reason to send a deadline is for the platform to defer — so the day it does, a union too narrow to hold the value mis-types `handle.status()` and makes `modelRunResultSchema.parse` reject a real payload. Reserving the member costs a consumer one branch that is currently unreachable; omitting it costs a silent type lie. It's kept out of `MODEL_TERMINAL` exactly as on the coding side, and its doc comment says it is reserved rather than mirrored.

## Ships on its own

The server ignores `deadline_minutes` until SAP-3202 lands, so this is zero behavior change. Includes a **minor** changeset for `@sapiom/tools` — the wire ticket's `execution-surface.contract.json` fixture is checked against the published package, so this has to publish before that ticket can go green.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm lint` — all clean across the workspace.
- `npx jest --maxWorkers=1 src/models/{launch,run-launch,coding-result,resume-payload}.spec.ts` — 40 passed. (Full suite left to CI per repo policy.)
- `provider-neutral-copy-check` and `agent-studio-terminology-check` pass.

Closes SAP-3201

https://claude.ai/code/session_019s65cc2Q5ravSTn9cZUj7e


gwitwer:
Thanks — finding 1 is a real bug and finding 3 is a real inconsistency this PR introduced. Both fixed in d50f5b3b, along with 2 and 4. One partial disagreement, on the taxonomy half of 2.

### 1. `run()` throwing past its default timeout — fixed

Correct, and it was the headline call: `coding.run({ task, deadlineMinutes: 60 })` would have thrown at 20 minutes the moment the platform actually deferred it, with no way out on a surface that takes no `timeoutMs`.

Fixed one level up from where you suggested. Rather than patching `codingRun`/`run`, the *handle's* default is now derived from the spec, so `launch()` + `wait()` gets the same protection as `run()`:

```ts
async wait({ timeoutMs = defaultWaitMs(spec.deadlineMinutes, 20 * 60_000), pollMs = 3_000 } = {})
```

`defaultWaitMs` returns `deadlineMinutes * 60_000 + surfaceDefaultMs` — the deadline bounds when the run *finishes*, so the surface default rides on top as slack for dispatch latency and poll granularity rather than replacing it. It only ever widens (a deadline under the default, or a non-positive one, leaves the default alone), and an explicit `wait({ timeoutMs })` still wins.

Three tests per surface, driving a run through `awaiting_capacity` while a stubbed `Date.now` jumps 25 minutes past the coding default (15 past the model one): the deadline case completes, the no-deadline case still throws `timed out after 1200000ms (last status: awaiting_capacity)`, and an explicit `timeoutMs: 1000` still throws at 1000ms. You were right that the old test only proved the loop iterates.

### 3. `deadlineMinutes` on `ModelRunSpec` with no deferred state to land in — fixed

You've got the better argument here and I've taken it. `ModelRunSpec.deadlineMinutes` is a hard requirement of the ticket (there's an explicit acceptance criterion for `models.run({ prompt, deadlineMinutes: 30 })`), so holding it off isn't available — which leaves widening, and widening is the right side to err on anyway. A union that's too wide costs a consumer one unreachable branch; one that's too narrow silently mis-types a real runtime value and makes `modelRunResultSchema.parse` throw on a real payload.

`ModelRunStatus` now has `awaiting_capacity`, kept out of `MODEL_TERMINAL` exactly as on the coding side, and the inline status list in `modelRunResultSchema.parse` matches. Tested: a deferred model run keeps polling to completion, and `parse` accepts the status.

### 2. Changeset copy — fixed as to tense; keeping the lane vocabulary

The tense point is right and worth more than it looks, because a changeset compiles into a CHANGELOG that can't be retracted. Both the changeset and the README now say plainly that the field is accepted and sent on the wire, that the platform does not honor it yet, and that setting it does not change dispatch or price until the server ticket lands.

On dropping `priority` / `standard` / `flex`: keeping them, deliberately. Naming that vocabulary is an explicit requirement of SAP-3201 ("Doc comment on the field states the vocabulary: the deadline derives the billing lane (`run_now` / `priority` / `standard` / `flex`)"), and the point of stating it is precisely that the caller *doesn't* name a lane — it tells an author what the knob they're turning controls, so they can reason about why 5 minutes and 5 hours price differently. `run_now` is also already public on this surface (`CodingRunOutcome.lane`, `ModelRunOutcome.lane`), so the axis isn't new, only the rest of its range. Happy to revisit if someone with the pricing surface wants it held back.

### 4. Exhaustive-switch break — added

The changeset now carries the heads-up, with the guidance a consumer needs at the break: handle `awaiting_capacity` as non-terminal, the run is still in flight.

**Verification:** `pnpm build`, `pnpm typecheck`, `pnpm lint` clean (only pre-existing unused-import warnings in unrelated packages). `npx jest --maxWorkers=1 src/models/` — 46 passed across the four spec files, up from 40. Copy and terminology checks pass.

gwitwer:
Round 2 addressed in 7cfd75ed. Both new items were fair; the poll-count one was self-inflicted by round 1's fix.

### New 1 — flat poll interval against a widened budget: fixed

Right, and it's the direct cost of the round-1 change: widening the budget raised the poll ceiling with it. Your arithmetic holds — a `deadlineMinutes: 480` model run parked for its window went from ~300 polls to ~14,700.

The interval now doubles while the status is `awaiting_capacity`, capped at 60s:

```ts
function nextPollMs(status: string, currentMs: number, callerPollMs: number): number {
  if (status !== "awaiting_capacity") return callerPollMs;
  return Math.min(currentMs * 2, DEFERRED_POLL_CAP_MS);
}
```

That's ~484 polls for the 8-hour case, back in the same order as the old ceiling. Every other status — `running` included — snaps straight back to the caller's `pollMs`, so a run that's actually executing is still observed at full cadence and the terminal transition is caught within one interval. Worst case a parked run notices dispatch up to 60s late, which is nothing against a deadline measured in hours.

Two tests, recording the delay each sleep asks for while firing it immediately: `[2, 4, 8, 1, 1]` across `awaiting_capacity ×3 → running ×2 → completed` (doubling *and* snap-back), and `[60_000, 60_000]` from a 40s `pollMs` (the clamp).

**Declining the second half — exposing `pollMs` on `run()`.** That changes the public signature of `codingRun`/`run`, which the ticket doesn't ask for, and the escape hatch already exists: `launch()` + `wait({ pollMs })`. The backoff is the part that actually needed fixing, because it helps `run()` callers who by definition can't pass anything. If someone wants tunable polling on the blocking surface, that's worth its own ticket with a considered options shape rather than a parameter bolted onto the end of `(spec, transport, baseUrl)`.

### New 2 — stale copy in the PR body and the JSDoc: fixed

Both were genuinely contradictory, and the PR body one would have landed in the merge commit.

The **Scope** section is rewritten to say `ModelRunStatus` does gain the member, that this reversed the first revision after review, and why the original reasoning was backwards.

The **JSDoc** now says the member is *reserved, not mirrored* — the gateway does not emit it on this surface today; it's declared because this surface accepts a deadline, and the only reason to send a deadline is for the platform to defer. I've stated it that way rather than citing a backend guarantee, because there isn't one to cite yet: SAP-3202 hasn't landed. Reserving costs one unreachable branch, omitting costs a silent type lie.

### On the lanes and the round-1 correction

Noted, and thanks for the correction on `CodingRunOutcome.lane` / `ModelRunOutcome.lane`. Your standing objection to `priority` / `standard` / `flex` in an unretractable CHANGELOG is recorded here for whoever owns the pricing surface; if they want the range held back to `run_now`, it's a one-line edit before publish and I'd take that patch.

**Verification:** `pnpm build`, `pnpm typecheck`, `pnpm lint` clean. `npx jest --maxWorkers=1 src/models/` — 48 passed across the four spec files (46 → 48). Copy and terminology checks pass.

packages/tools/src/models/index.ts
```diff
@@ -417,18 +494,24 @@ export async function codingLaunch(
...
+      pollMs = 3_000,
+    } = {}) {
       const deadline = Date.now() + timeoutMs;
+      let intervalMs = pollMs;
       // eslint-disable-next-line no-constant-condition
       while (true) {
         const d = await fetchDoc();
-        if (TERMINAL.has(d.data.attributes.status)) return toResult(d);
+        const status = d.data.attributes.status;
+        if (TERMINAL.has(status)) return toResult(d);
```
GrimalDev: if the timeout is exeeded, it could still respond the result instead of a timeout, is it fine?

GrimalDev, line 528: if the timeout is exeeded, it could still respond the result instead of a timeout, is it fine?

GrimalDev (changes requested):

packages/tools/src/models/index.ts
gwitwer: Yes, and it's deliberate — good question to ask, because the ordering is load-bearing and wasn't written down.

The loop checks `TERMINAL` before it checks the budget, so if a poll comes back `completed`/`failed` when the clock has already run out, the result wins and `wait()` resolves. That's the behavior I want: runs are billed, so throwing a timeout while holding a finished run's result would charge the caller for something we then discarded. The timeout exists to stop waiting, not to suppress an answer we already have. Both loops now carry a comment saying that, and there's a test that drives a 1ms budget with 25 minutes elapsed and a `completed` poll, asserting the result comes back.

Your question did turn up something that *was* wrong, though, one line below. The backoff I added in the previous commit escalates the poll interval up to 60s while a run is parked, and that sleep was unclamped — so against a shorter `timeoutMs` the loop could sleep a full minute past the budget and then throw `timed out after 5000ms` after actually waiting ~65s. The message would have been a lie and the bound much softer than advertised.

Fixed in 98357aef: the sleep is clamped to the remaining budget.

```ts
function sleepMs(intervalMs: number, deadline: number): number {
  return Math.max(0, Math.min(intervalMs, deadline - Date.now() + 1));
}
```

So the final poll now lands at the budget's edge instead of a minute beyond it — which also gives the run its best chance of being seen finishing before we give up, i.e. it makes the case you asked about *more* likely to resolve with a result. The clamp is deliberately kept out of the backoff state so a shortened last sleep doesn't reset the escalation. Tested: a 40s `pollMs` against a 5s budget asks for ~5s, not the 60s the backoff wanted.

`pnpm build` / `typecheck` / `lint` clean; `src/models/` is 50 tests passing, up from 48.

GrimalDev: nice fix! thanks for the clarifications, I agree but wanted to be sure it was deliberate.

GrimalDev (approved):
LGTM

PR #840, opened 2026-09-05
fix(harness): show one message in the canvas Render-failed state [SAP-3199]

gwitwer:
Fixes SAP-3199.

## The bug

When an agent cannot render, the canvas pane drew **two error strings on top of
each other**: the app's Render-failed card and the rendered document's own prose,
in the same area, unreadable.

Two things paint the same failure:

1. `CanvasPane.tsx` draws the Render-failed card (glyph, short claim, one-line
   reason, Ask/Retry, full reason behind Details). `.canvas-render-error` is
   `position: absolute; inset: 0` with **no background**, so it is a transparent
   layer over the iframe.
2. The document in that iframe (`buildErrorPanelHtml`) paints its own paragraph:
   "Could not extract this agent's step graph: ... Use the workbench actions to
   ask your coding agent to fix it or retry the deterministic render."

Nothing hid the second one, so both showed through each other.

The stylesheet already claimed otherwise. The comment above
`.canvas-render-error` read "the document's own error text is hidden by the board
snippet". That was the intent; the mechanism was never there. This PR implements
it and corrects the comment.

## The fix

The document stands its prose down while it is embedded, exactly the way the
title, badge and legend already do. `canvas-template.ts` has a rule that visually
hides `.canvas-header` and `.canvas-legend` because they are app chrome the SPA
draws around the iframe, with the reason spelled out: "floating a second copy
over the board would just crowd the diagram". The failure reason is the same kind
of duplicate, so it gets the same treatment, one line lower in the same block.

- `buildErrorPanelHtml` tags the paragraph `canvas-render-error-note`.
- A new `:root[data-canvas-embedded] .canvas-render-error-note { display: none }`
  sits beside the existing header rule.
- A four-line head script sets `data-canvas-embedded` when
  `window.parent !== window`. It is its own script, before the body, so the flag
  is set before first paint and a parse error in the much larger run-state script
  cannot leave it unset.

Opened standalone there is no card coming, so the prose stays and is the only
message. That direction is pinned by its own spec.

Nothing about the card itself changed, and no user-visible string changed.

## Verification

Real harness server (`node dist/cli/bin.js`, `--state-root` of its own) against a
real agent tree, selecting a project folder with no `index.ts`, which is the repro
in the ticket. Light and dark, before and after, below.

Specs, in `e2e/canvas-render.spec.ts`:

- **embedded** shows one message. The reason reaches the parent exactly once, and
  no element in the document paints a second copy. "Paints" is hit-tested with
  `elementFromPoint` rather than inferred from the box, because the title and
  badge sit inside a clipped 1px `.canvas-header` while keeping full-size rects
  of their own. Every text-bearing element is walked, not just the two known
  ones, so a third string added later cannot quietly reintroduce the overlap.
- **standalone** keeps the prose as the only message.

Mutation-tested, three ways, each confirmed red before restoring:

| Mutation | Result |
| --- | --- |
| Delete the stylesheet rule | embedded spec fails |
| Drop the class from the paragraph | embedded spec fails |
| Hide unconditionally (drop the `window.parent` check) | standalone spec fails |

Local: `pnpm build`, `pnpm typecheck`, `pnpm lint` all clean. `canvas-render.spec.ts`
and `canvas-template.spec.ts` 13 passed; `canvas-render.test.ts` and
`canvas-template.test.ts` 52 passed. Per repo policy the full suite is left to CI.

Before and after screenshots against a real server, light and dark, are in a
comment below, with the measured `documentProseVisible` table. Review rounds 1
and 2 are answered in their own comments.

Linear: https://linear.app/sapiom/issue/SAP-3199

https://claude.ai/code/session_01WoapZZPdPVtz11TQCK5eQH


gwitwer:
## Round 1 addressed

Both findings were real. Pushed in 9f77036 and 23bad72.

### 1. Changeset

Added `.changeset/one-message-on-render-failure.md`, `patch` for both
`@sapiom/harness` and `@sapiom/harness-desktop`, following the
`one-studio-dialog-shell.md` precedent you pointed at.

### 2. The fail-safe pointed the wrong way

Correct, and it is the sharper of the two. My own rationale for splitting the flag
into a head script was that the run-state bundle can abort as a whole. The split
then protected the hiding and left the showing in the fragile script, so a document
that could not post lost its prose and gained nothing: an empty dotted board and no
message anywhere. That is worse than the overlap this PR set out to fix, and it
falsified the `catch` comment in `bootCanvasError` that promises the panel stays
visible.

I did not take the suggested shape exactly, because clearing the attribute on the
`return` and `catch` paths inside `bootCanvasError` closes two of the three ways to
end up with no card and leaves the third, which is the whole-script abort the split
was supposed to survive. One mechanism covers all of them instead:

- a successful post marks the root `data-canvas-error-posted`
  (`canvas-run-state.ts`),
- the head script's `data-canvas-embedded` is now **optimistic**, and at `load` a
  document with no such mark takes the flag back and shows its own prose.

`load` fires after `DOMContentLoaded`, so `bootCanvasError` has always had its turn
by then. The normal path keeps the no-flash property, since the flag is still set in
the head before the body paints.

New spec, as you asked, in `e2e/canvas-render.spec.ts`: the rendered document's
`#sapiom-render-error` payload is corrupted to `{ not json`, the document is embedded,
and it must still show exactly one message, its own prose, with zero posts to the
parent. Mutation-tested both halves:

| Mutation | Result |
| --- | --- |
| Head script never withdraws the flag | corrupt-payload spec fails |
| `bootCanvasError` never marks a post | overlap spec fails |

### Nit: a centre point outside the frame

Fair, `elementFromPoint` answers null out there and a message below the fold would
have passed. The walk now reports a box that lies wholly outside the frame in a
separate `offscreen` list, asserted empty alongside `painted`, and clamps the probe
point into the frame for a box that straddles the edge. Zero-area boxes are skipped,
since `display: none` paints nothing.

Local: build, typecheck and lint clean. `canvas-render.spec.ts` 9 passed,
`canvas-template.spec.ts` 4 passed, the three canvas vitest files 58 passed.

Also removed three throwaway screenshot probes that went into 9f77036 by mistake.
Before and after screenshots against a real server are coming in a follow-up comment.

gwitwer:
## Round 2 addressed

### 1. The withdrawal does not un-hide the header and legend. Disagreeing, with a spec.

`.canvas-header` / `.canvas-legend` are clipped by an **ungated** selector
(`canvas-template.ts:170-174`):

```css
.canvas-header,
.canvas-legend {
  position: absolute; width: 1px; height: 1px; overflow: hidden;
  clip-path: inset(50%); ...
}
```

No attribute, no `:root[...]` prefix. `data-canvas-embedded` gates exactly one rule
in the whole stylesheet, the one this PR added:

```
$ grep -n "data-canvas-embedded" packages/harness/src/core/canvas-template.ts
182::root[data-canvas-embedded] .canvas-render-error-note { display: none; }
294:  root.setAttribute("data-canvas-embedded", "");
296:    if (!root.hasAttribute("data-canvas-error-posted")) root.removeAttribute(...)
```

So withdrawing the flag on a successful board takes nothing away from it: the chrome
was never held back by that flag, and the prose rule is the only thing keyed on it.
The chrome contract at `canvas-template.ts:164` is untouched by this PR.

Round 1's suggested shape and round 2's objection both read the flag as the thing
holding the chrome down. It is not, which is also why the "split the two concerns"
alternative would be a no-op: they are already split.

### 2. No spec covered the embedded successful render. Fair, and added.

That was a real gap regardless of #1, and it is the assertion that settles the
question, so it is worth having on its own merits. `an embedded successful render
still hides its header and legend` renders `ORDER_TRIAGE`, embeds it in the same
sandboxed parent shell, and asserts:

- the graph paints (`rect.canvas-node-rect` visible),
- `.canvas-header` and `.canvas-legend` are attached and still clipped to a box of
  at most 1x1,
- `.canvas-title` is not in the viewport,
- and, so the spec cannot pass for the wrong reason,
  **`data-canvas-embedded` is `null`** — the withdrawal really did happen on this
  document, and the chrome stayed hidden through it.

It passes on the current code. Mutation-tested against the world round 2 assumed, by
gating the header rule on the flag:

```
:root[data-canvas-embedded] .canvas-header,
:root[data-canvas-embedded] .canvas-legend { ... }
```

```
1) [chromium] › e2e/canvas-render.spec.ts:243:1 ›
   an embedded successful render still hides its header and legend
   Error: expect(received).toBeLessThanOrEqual(expected)
```

So the spec does detect that regression, and the unmutated code does not have it.

### 3. Round 1's shape

Agreed, and it is why I did not follow it literally. `if (!el) return` is the healthy
render, not a degraded error document, so clearing the flag there would have been the
one place it genuinely should not be cleared. The marker-plus-withdrawal covers the
missing element, the `catch`, and the whole-script abort with one rule.

### Checks

`canvas-render.spec.ts` and `canvas-template.spec.ts`: 15 passed. Canvas vitest files:
58 passed. Build, typecheck and lint clean.

gwitwer:
## Before and after, real server, real agent tree

Two isolated checkouts, each with its own build and its own agent tree, so neither
side can read the other's cached render:

- **before**: a detached worktree at `origin/main` (082d82c0). Verified the built
  `dist/core/canvas-template.js` contains no `data-canvas-embedded`.
- **after**: this branch. Verified the same file does contain it.

Each ran `node dist/cli/bin.js <tree> --port <p> --state-root <own>`, and the tree is
three real examples (`hello-agent`, `news-roundup`, `proposal-generator`) plus
`research-notes`, a project folder with a `sapiom.json` marker and no `index.ts`.
That is the ticket's repro, and it produces the ticket's reason: "No index.ts found
in ... Run this from an agent project, or pass its directory."

Measured in the page, not just looked at. `documentProseVisible` asks the iframe
whether the document's own paragraph is visible underneath the card:

| | card shown | document prose also painting |
| --- | --- | --- |
| before / light | yes | **yes** |
| before / dark | yes | **yes** |
| after / light | yes | no |
| after / dark | yes | no |

### Light

before, two strings through each other:

before light

after, one message:

after light

### Dark

before:

before dark

after:

after dark

The card is unchanged in all four: same glyph, same short claim, same one-line reason,
same Ask/Retry, same Details disclosure holding the full reason. The only difference is
that the document underneath has stopped painting its own copy.

ynadge (approved):

PR #841, opened 2026-09-05
feat(harness): project rows carry a live-session indicator [SAP-3200]

gwitwer:
Project rows show a standing green dot when their project has live coding-agent sessions. The count appears in the tooltip and accessible name, and the dot disappears when the last live session exits. Group headers show the same indicator for their member agents.

Related: [SAP-3200](https://linear.app/sapiom/issue/SAP-3200/agent-studio-project-rows-show-an-indicator-when-the-project-has-a).

## Design dependency

[sapiom/design-eng#185](https://github.com/sapiom/design-eng/pull/185) is still open and must merge before this PR. The visual implementation follows its light/dark rail-live-indicator frames and reuses the existing `.session-dot[data-status="running"]` recipe. The design's membership explanation should reflect the durable project identity described below.

## Behavior and membership

- Starting and running sessions count, including sessions that are not yet bound to an agent. Exited sessions do not count.
- The project indicator stays visible at rest, ahead of the hover-revealed row actions. Its label is `1 live session` or `N live sessions`; zero renders no dot.
- Each Studio row resolves the server-issued project ID through its workspace scope. `liveSessionsForStudioProject` selects its sessions using the same exact identity as the session tab strip.
- The project mark uses that list's length, and `GroupSections` receives the same list before filtering by member-agent binding or an unbound session's exact agent folder. Shared agent paths therefore cannot pull another project's sessions into a group.
- Independently opened nested projects remain separate. A project's sessions still count when they live outside the displayed root, including another root belonging to the same durable project. Sessions missing a project identity do not fall back to cwd matching inside a durable Studio project.
- Rows from older servers without a Studio project retain the existing folder-based project rule. Groups narrow that project's session list by agent membership.

Agent rows retain their existing appearance, and the rail continues to list agents rather than individual sessions. The changeset covers `@sapiom/harness` and `@sapiom/harness-desktop`.

## Visual evidence

Existing screenshots demonstrate the indicator's visual treatment; the project-ID correction changes session membership without changing its styling.

| | Light | Dark |
|---|---|---|
| Before, two live sessions | before light | before dark |
| After, two live sessions | after light | after dark |
| After, no live sessions | quiet light | quiet dark |

[Existing recording of the count appearing and clearing](https://raw.githubusercontent.com/sapiom/sapiom-js/pr-media/media/SAP-3200/after-live-mark.mp4).

## Validation

Verified on the branch with current released main merged in:

- **124 Vitest tests passed** across `project-live`, `session-scope`, `css-brand-surface`, and `api`.
- **11 Playwright tests passed in one serial run** across `rail-live-indicator.spec.ts` and the new `rail-live-project-identity.spec.ts`.
- The three new browser regressions each failed against the previous implementation, then passed with the fix: nested project isolation, project-scoped groups sharing agent paths, and identity-based counting outside the displayed root with no fallback for unidentified sessions.
- Harness server/web typechecks passed.
- Terminology check passed across 949 files with no stale allowlist entries.
- Explicit ESLint checks of the touched web source and new browser spec passed with no errors; the existing four unused-variable warnings in `WorkflowsRail.tsx` remain. The package's default lint command excludes `web`, so these checks explicitly include it.
- Prettier checks passed for the new browser spec, group-membership helper, and changeset; `git diff --check` passed.

The full repository suite and native desktop smoke are covered by CI. No full local suite or native desktop run was performed for this correction.

AI assistance: Claude Code and Codex.


gwitwer:
Round 1 addressed in 4bcfb7e4. Findings 1, 2 and 3 taken; one nit taken, one declined with reasoning.

## 1. Changeset: added

`.changeset/project-live-indicator.md`, `minor` for `@sapiom/harness` and `patch` for `@sapiom/harness-desktop`, matching `project-session-shortcut.md`. Written for an anonymous npm reader: no ticket ids, no design-eng links.

## 2. Second membership answer: you were right, and it is gone

I confirmed the failure rather than taking it on faith. `liveSessionsForProject` (`session-scope.ts:164`) is what renders the tab strip at `App.tsx:887` and `App.tsx:1433`, and its own docblock names `POST /api/agents/move` as the reason membership is derived one way. The move is reachable in this repo (`agent-move.ts`, drag-to-move on the Project axis), so the divergence was not hypothetical: A's row and B's row would both have said "1 live session" for one session, and selecting B would have opened an empty strip.

I took your second option, which is also what my own justification argued for: the binding clause is gone and the rail now calls `liveSessionsForProject(sessions, project.root).length`. `liveSessionsInProject` is deleted rather than kept alongside, so there is no near-identical name for the next caller to pick by autocomplete. `project-live.ts` now holds only `liveSessionsOnAgents` and `liveSessionsLabel`, which is the part session-scope has no answer for: a group has no directory behind it, so it cannot be asked the containment question. Its module docblock records the rejected draft and why, so the clause is not re-added later as an improvement.

Coverage did not shrink: `liveSessionsForProject` is already pinned by 8 assertions in `session-scope.test.ts` (exited sessions, nested roots, null and empty root, ordering), so the deleted tests were duplicating it. `project-live.test.ts` now covers the group rule only, including the exit walk 2 -> 1 -> none.

## 3. Group mark dark for the default session flow: documented, and now covered on screen

The behaviour is intended, and you are right that it reads like a bug. It is now stated in two places: `liveSessionsOnAgents`'s docblock and the `liveCount` JSDoc on `GroupRowProps`. The reasoning is that a session is unbound until its agent is known, and an unbound session at a project root is working on no member yet, so crediting a group would print a guess as a fact.

I also took the fixture suggestion, since it closed a hole I had flagged myself. `?mockBoundSession=1` seeds exactly one session bound to `gateway`, the head of the deep fixture's biggest derived group, behind its own query parameter so that nothing already running against `mockFixtures=deep` moves. The new spec asserts the mark on the `gateway` header with a counted name, its absence on `mailer` and `Ungrouped`, exactly one group mark in the rail, and the same session counted once by the project row above it.

## Nits

**Opacity assertion: taken.** Correct, the element's own opacity reads 1 inside a faded ancestor. The spec now walks up to `.workspace-row`, multiplies opacity along the chain, and asserts the mark is not inside a `.workspace-row-action`, then checks both facts at once. Mutation 7 below confirms it.

**Per-render allocation: declining.** Taking this would mean adding a counting variant beside `liveSessionsForProject`, which is the second answer finding 2 asked me to remove; the two findings pull opposite ways and correctness wins. The measured cost is a filter and a sort over the session list once per project row: a rail with a dozen projects and a handful of sessions, inside a render that is already building the whole project tree. If it ever shows up in a profile, the fix is to hoist one grouping pass above the map rather than to give the mark private membership logic.

## Mutation test, re-run after the changes

7 mutations, 7 killed, 0 survivors. Each applied to a file while the dev server was already running, so HMR invalidated the transform rather than serving a cached one.

| Mutation | Result |
|---|---|
| the mark never renders | 4 failed |
| the mark renders regardless of liveness | 4 failed |
| exited sessions count as live | 4 failed |
| the accessible name drops the count | 3 failed |
| every group header claims a live session | 1 failed |
| the group rule ignores the binding and matches cwd only | 1 failed |
| the mark is hover-revealed (`opacity: 0` at rest) | 1 failed |

## Checks

`tsc --noEmit` (server and web), `eslint src`, `vitest` on the changed area (`project-live`, `session-scope`, `css-brand-surface`, `api`: 115 pass), and playwright over `rail-live-indicator`, `group-axis`, `project-axis`, `session-scope`, `session-tabs`, `rail-grammar`, `project-map-groups`, `agent-map-planning` (96 pass, plus the 8 in `rail-live-indicator`). Full suite left to CI.

The merge gate on design-eng#185 still stands: it is still OPEN, and this must not merge first.

gwitwer:
Round 2 addressed in 4411f4d9. Both new items taken; the nit declined again, with the reason sharpened.

## The project/group asymmetry: documented, not fixed

You are right, and it is the mirror of the round-1 bug: a session bound to an agent under project A but rooted outside A lights the group header and leaves A's row dark. `POST /api/agents/move` is the way to produce it.

I took the docblock line, and wrote down why it is left standing rather than closed, since "known and deliberate" is only useful if the next reader can see the reasoning:

- A group is a label over agents with nothing on disk behind it, so binding is the only membership it has. It is the same rule `liveSessionsForFocus` applies to one agent, and that agent's own tab strip lists the very same session, so the group header agrees with the rows inside it.
- Intersecting the group rule with the project's containment would make a group mean something different from its own agent rows.
- Adding a binding clause to the project rule is exactly the second membership answer round 1 had me remove.

Every alternative is worse than the odd mark, so the mark stays odd and the rules stay singular. That is now in `liveSessionsOnAgents` under a heading, and in the PR body's `did_not_work`.

## PR body: refreshed

Correct, it was stale in three places. Updated: `did_not_work` no longer claims the group mark has no positive e2e (it names the two real remaining limits instead), the spec counts read 11 unit and 8 e2e, and the Checks table lists the `rail-live-indicator` run that covers `mockBoundSession`. The Derivation section now leads with `liveSessionsForProject` rather than describing the rule I deleted, and `did_not_work` records that the first draft shipped a second membership rule and that round 1 caught it, rather than quietly rewriting history: the near-identical name was the durable half of that mistake and is worth leaving visible.

## Nit, declined again

Correct that `liveSessionsForProject` sorts as well as filters, and that both call sites read only `.length`. Declining for the same reason as round 1, which the sort does not change: the only way to avoid the work is a counting variant that duplicates membership, and a second function answering "is this session in this project" is the round-1 blocking bug. A per-render sort of a session list that is single-digit in every real install, inside a render already building the whole project tree, is not worth reopening that door. If it ever appears in a profile, the fix is one grouping pass hoisted above the project map, feeding both marks, and that keeps a single membership rule.

## Checks after the change

`tsc --noEmit` (server and web) pass, `prettier --check` clean, `project-live.test.ts` 11 pass, `rail-live-indicator` 8 pass. The edit is comment-only, so the mutation results from round 1 stand unchanged.

The merge gate on design-eng#185 still stands: it is still OPEN, and this must not merge first.

gwitwer:
Rebased onto `main` (twice: main moved 19 commits during the first verification pass). Conflicts resolved, and the one red check here is not from this branch.

## Conflicts

Two, both in files the Agent Map series rewrote under me:

- `WorkflowsRail.tsx`: main renamed the handler to `focusProjectAgent` while I added `sessions={sessions}` to the same `GroupSections` call. Took both.
- `api.ts`, twice: the first rebase hit `plannerSessionRecords`, which main had DELETED, so only my seeding change survived rather than resurrecting a field the migration removed. The second rebase hit main's new `restoredSessions` fixture, which forces every mock session exited; combined, so the bound-session fixture joins the seeded list and a restored Studio exits it along with the others.

Re-verified on the final base rather than trusting the merge, since main rewrote all three files: `tsc --noEmit` (server and web), `eslint src`, vitest (`project-live`, `session-scope`, `css-brand-surface`, `api`: 124 pass), and playwright over `rail-live-indicator` (8), plus `smoke`, `group-axis`, `project-axis`, `session-scope`, `session-tabs`, `rail-grammar`, `project-map-groups`, `agent-map-planning` (167 pass together).

## The red check is main's, and there is a PR for it

`examples` fails on `pnpm terminology:check`, in `AgentMapPane.tsx`, which this branch does not touch:

```
Stale terminology allowlist entries found:
  agent-map-pane-icon-identifier [workflow] - packages/harness/web/src/components/AgentMapPane.tsx / ^Workflow$ (expected 2, matched 3)
```

`AgentMapPane.tsx`, the guard script and the allowlist are all byte-identical here to `origin/main`, so the failure cannot originate on this branch, and `main`'s own `test.yml` run at 8495be29 is red for the same reason. #844 added a third neutral Agent Map state carrying the same private `icon="Workflow"` identifier without bumping the allowlist count.

Fixed in #849, deliberately as its own PR rather than folded in here: it is nobody's feature and it blocks every PR rebased past 8495be29. **#841 cannot go green until #849 merges**, and I have not merged it, since that is not mine to do.

The merge gate on design-eng#185 also still stands: it is still OPEN, and this must not merge before it.

gwitwer:
Both red checks are `main`'s, not this branch's. I measured rather than assumed, by running the same suites against a clean `origin/main` worktree.

## `examples`: main's terminology allowlist. Fixed in #849

Covered in the previous comment. #849 is green, including the `examples` job that fails here.

## `playwright-mock`: an Agent Map page error, reproduced on clean main

Two hard failures, both in `agent-map-initialization.spec.ts`, which this branch does not touch:

```
✘ queued and running generation show a compact state without creating a session
✘ storage errors remain separate from generation failures

expect(errors).toEqual([])
+ "Cannot read properties of undefined (reading 'dimensions')"
```

Both specs assert the page raises no `pageerror`, and the Agent Map generating and load-error states raise one. The read is not in harness source (`grep -rn dimensions web src` finds only an unrelated comment), so it comes from the canvas bundle the Agent Map series loads.

**Reproduced on a clean `origin/main` worktree at 8495be29** (same two tests, same message, failing through all retries), and the full mock suite gives the same profile on both trees:

| Tree | Result |
|---|---|
| clean `origin/main` | 2 failed, 2 flaky, 563 passed |
| this branch | 2 failed, 2 flaky, 571 passed |

Same two hard failures on both. The extra 8 passes here are this PR's own specs.

`main` has no green baseline to compare against: every recent `harness.yml` run on `main` is `cancelled`, superseded by the next merge in the Agent Map series, so nothing has completed `playwright-mock` on `main` for the whole series.

I have not fixed this one. It is a page error inside another team's in-flight Agent Map work, the failing specs are theirs, and guessing at a canvas bundle I do not own is how a small PR turns into someone else's debugging session. It needs whoever owns the Agent Map series.

## `project-map-navigation` in the CI run is a flake, not a third failure

CI also showed `project-map-navigation.spec.ts:98` failing (both providers, all retries). It is not reproducible: it passes on this branch in isolation under CI settings (17 pass), passes in a narrower combination with `smoke`, and passed on the second full-suite run here. It is in the load-sensitive family the harness suite is already known for, and it does not fail on clean `main` either.

## Where this leaves #841

Rebased, mergeable, and every check that is actually about this branch passes: `test (20.x)`, `test (22.x)`, `desktop-smoke`, CodeQL, both Analyze jobs, and `claude-review`. It cannot show all-green until main's `agent-map-initialization` breakage is fixed by its owners, and until #849 merges.

The merge gate on design-eng#185 also still stands.

ynadge (approved):

PR #852, opened 2026-09-07
fix(agent): stop requiring nested and prefault-defaulted input fields [SAP-3218]

gwitwer:
## Primary change type

- [x] Bug fix
- [ ] Documentation
- [ ] Feature
- [ ] Tests
- [ ] Dependency update
- [ ] Maintenance or refactor

## Problem and motivation

A deployed agent's published `inputSchema` listed a field as `required` even when
it carried a Zod `.default()` — as long as the field sat below the top level.
`buildManifest` filtered only the top-level `required` array, so a default inside
a nested object, an array item, or a union/intersection branch survived. The
engine's AJV pre-gate then rejected a **partial** input for those fields while an
entirely omitted input validated:

```ts
z.object({
  name: z.string(),
  opts: z.object({ verbose: z.boolean().default(false) }),
});
```

`{ name: "x", opts: {} }` → rejected. `{ name: "x" }` → rejected too, but for the
right reason. Meanwhile a schema where every top-level field was defaulted
accepted `{}` and rejected `{ opts: {} }` — a partial input was stricter than no
input at all, which is the opposite of the guidance we ship ("give every field a
`.default()` so a zero-input run validates").

`.prefault()` was required at every depth for a related reason: Zod's output-mode
JSON Schema emits no `default` keyword for a prefault at all, so nothing marked
it omissible.

Reported via Studio feedback `2c3afb03-9cdd-4f0a-b58e-53172810c367` (production,
sapiom-mcp 0.14.0): a coordinator sent partial input to a child agent, the
pre-gate rejected it for defaulted fields, and the whole run failed.

Second half of the report: `stepInputContract` / `workflowInputContract`
returned the **raw** `z.toJSONSchema` output — no normalization at all — so
anything displaying the contract (dashboard Run form, inspect tooling) showed
defaulted fields as required and disagreed with what the engine enforced.

## Summary and scope

Convert with Zod's `io: "input"` mode. That is the mode which describes what a
caller may **send** rather than what a parse **returns**, so Zod itself drops
every defaulted field from `required` at every depth. Three further consequences,
all of them what an input contract wants:

- `.prefault()` and `.catch()` fields become optional and carry their value;
- a `.pipe()` / `.transform()` is described by its **input** type, so the
  pre-gate checks what the caller actually sent — and a `.transform()` in an
  input schema no longer fails the build with `Transforms cannot be represented
  in JSON Schema`;
- a plain `z.object()` is not emitted closed, matching its strip-not-reject
  parse.

`buildManifest`'s two local normalizers are replaced by one exported
`normalizeInputJsonSchema` in `introspection.ts`, which recursively drops
defaulted keys from `required` and strips `additionalProperties: false`
(`z.strictObject()` still emits it in input mode). It is pure JSON in, pure JSON
out, so the engine — which must run `z.toJSONSchema` with its own Zod module
instance to keep `.meta()` registry lookups on one instance — can reuse it. With
`io: "input"` the `required` rule is an invariant guard rather than the
load-bearing mechanism; the comments say so.

`stepInputContract` / `workflowInputContract` already called `zodToJsonSchema`,
so routing the normalization through it makes the displayed contract identical to
the published one by construction. A spec pins that equality.

**`.prefault()` decided explicitly** (the ticket asked for a decision): treated
exactly like `.default()`. From a caller's point of view both substitute a value
for a missing input, so an omitted prefault field is legal and must not be
required. The difference — a prefault value is then parsed, a default is not — is
invisible to a caller.

**Out of scope.** The engine needs no change once the manifest is right
(`useDefaults` is unnecessary; the sandbox Zod parse fills defaults). Note
though that `apps/workflows-engine/src/workflows/_core/introspection.ts` in the
Sapiom monorepo keeps its own copy of `stepInputContract` (deliberately, for the
Zod-module-instance reason above) and still returns the raw conversion — so
engine-**native** workflows retain the display-vs-enforced mismatch until that
copy passes `{ io: 'input' }` too. Different repo, different code path, filed
separately rather than smuggled in here.

## Related work

Related issue or discussion: [SAP-3218](https://linear.app/sapiom/issue/SAP-3218/sdk-published-inputschema-keeps-nestedprefault-defaulted-fields)

## Validation

```text
pnpm install --frozen-lockfile — ok (lockfile accepts the new ajv devDependency)
pnpm build                     — exit 0
pnpm typecheck                 — exit 0
pnpm lint                      — exit 0 (only pre-existing warnings in core/axios/fetch/langchain*)
npx jest --maxWorkers=1 src/introspection.spec.ts       — 18 passed
npx jest --maxWorkers=1 src/manifest-input-gate.spec.ts —  8 passed
npx jest --maxWorkers=1 src/manifest.spec.ts            — 25 passed
npx jest --maxWorkers=1 src/sdk.spec.ts                 — 56 passed
```

The full suite is left to CI (per repo policy on local jest runs).

### Tests and documentation

Two new specs plus one extension:

- `packages/agent/src/manifest-input-gate.spec.ts` — the end-to-end proof the
  ticket asked for. Compiles a real AJV 2020-12 gate from the manifest
  `buildManifest` publishes and asserts it accepts every input the Zod schema
  parses: partial nested input, array items missing their defaults, a union
  branch missing its default, an omitted `.prefault()`, `{}` when everything is
  defaulted, and extra unknown fields. It also asserts the gate still **rejects**
  a genuinely missing field, a wrongly-typed nested value, and an array item
  missing a non-defaulted key — several cases assert Zod and the gate agree, so
  the test cannot pass by making the gate uniformly permissive. This adds `ajv`
  as a `devDependency` of `packages/agent` (specs are excluded from both the CJS
  and ESM builds, so nothing new ships).
- `packages/agent/src/introspection.spec.ts` — schema-shape matrix: nested,
  array items, union, intersection, prefault, catch, transform, pipe,
  strictObject, typed catchall preserved, `.meta({ examples })` preserved,
  `normalizeInputJsonSchema` fixing an output-mode schema (the engine's reuse
  path) without mutating its input, a property literally named `required` or
  `additionalProperties` left alone, and the contract helpers returning the same
  schema the manifest publishes.
- `packages/agent/src/manifest.spec.ts` — a nested-default case next to the
  existing top-level one; the stale "nested objects out of scope for v1" comment
  is gone.

Documentation: N/A. `packages/agent/README.md` and the authoring skill already
tell authors to put a `.default()` on every field; that guidance was correct and
the implementation did not match it. No user-facing doc changed.

- Breaking or externally visible changes: `zodToJsonSchema` and the two contract
  helpers now emit the input-mode, normalized schema. Every in-repo consumer
  describes a step input, which is exactly what this mode is for, so the change
  is a fix in each case. It is externally visible in that a published manifest
  gets a smaller `required` set and no `additionalProperties: false` — strictly
  more permissive at the pre-gate, with the sandbox Zod parse still
  authoritative. **An already-deployed agent must be redeployed to publish the
  corrected schema**; this does not retroactively fix manifests in the engine's
  database.
- Changeset: Added (`patch` for `@sapiom/agent`).

## Security

      will follow the
      private reporting.

## AI assistance

Written with Claude Code. It read the existing normalizers and their comments,
probed Zod 3.25.76 (the floor of the `^3.25.76 || ^4.0.0` peer range) to confirm
the `zod/v4` subpath accepts `io: "input"` and to check its output across
`.default()`, `.prefault()`, `.catch()`, `.coerce`, `.enum`, `.union`,
`.nullable()`, `.transform()`, `.pipe()`, `z.record`, `z.strictObject`,
tuples, intersections, `$defs`/recursive schemas and `.meta({ examples })`, then
wrote the change and the specs. Verified by the commands under Validation; every
behavioral claim in this description is covered by a passing assertion.

## Checklist

      any N/A checks above.

https://claude.ai/code/session_01EyRQfxZWtu9BhJxx1dhDXu

### DIFF
--- a/.changeset/pink-lions-normalize-input.md
+++ b/.changeset/pink-lions-normalize-input.md
@@ -1,0 +1,28 @@
+---
+"@sapiom/agent": patch
+---
+
+fix(agent): a defaulted field is no longer required at any depth in a published input schema
+
+A step's published `inputSchema` listed a field as `required` even when it
+carried a Zod `.default()`, as long as the field sat inside a nested object, an
+array item, or a union/intersection branch. The engine's AJV pre-gate then
+rejected a **partial** input for those fields while an entirely omitted input
+validated — so `{ opts: {} }` failed where `{}` passed, contradicting the
+authoring guidance that a default makes a field omissible. A `.prefault()` field
+stayed required at every depth for the same reason.
+
+`zodToJsonSchema` now converts in Zod's `io: "input"` mode — the mode that
+describes what a caller may SEND — and runs the result through a new exported
+`normalizeInputJsonSchema`, which recursively drops defaulted keys from
+`required` and strips `additionalProperties: false`. Three consequences:
+
+- `.default()`, `.prefault()` and `.catch()` fields are optional at every depth;
+- `stepInputContract` / `workflowInputContract` now return the same normalized
+  schema the manifest publishes, so a displayed contract (the Run form, inspect
+  tooling) agrees with what the engine enforces;
+- a step whose `inputSchema` contains a `.transform()` no longer fails the build
+  with "Transforms cannot be represented in JSON Schema"; it is described by its
+  input type, which is what a caller sends.
+
+Redeploy an agent to publish the corrected schema.
--- a/.changeset/pink-lions-normalize-input.md
+++ b/.changeset/pink-lions-normalize-input.md
@@ -1,28 +1,37 @@
 ---
-"@sapiom/agent": patch
+"@sapiom/agent": minor
+"@sapiom/agent-runtime": patch
 ---
 
-fix(agent): a defaulted field is no longer required at any depth in a published input schema
+A defaulted field is no longer required at any depth in an input schema
 
 A step's published `inputSchema` listed a field as `required` even when it
 carried a Zod `.default()`, as long as the field sat inside a nested object, an
-array item, or a union/intersection branch. The engine's AJV pre-gate then
-rejected a **partial** input for those fields while an entirely omitted input
-validated — so `{ opts: {} }` failed where `{}` passed, contradicting the
-authoring guidance that a default makes a field omissible. A `.prefault()` field
-stayed required at every depth for the same reason.
+array item, or a union/intersection branch. An AJV pre-gate then rejected a
+**partial** input for those fields while an entirely omitted input validated —
+so `{ opts: {} }` failed where `{}` passed, contradicting the authoring guidance
+that a default makes a field omissible. A `.prefault()` field stayed required at
+every depth for the same reason.
 
 `zodToJsonSchema` now converts in Zod's `io: "input"` mode — the mode that
 describes what a caller may SEND — and runs the result through a new exported
-`normalizeInputJsonSchema`, which recursively drops defaulted keys from
-`required` and strips `additionalProperties: false`. Three consequences:
+`normalizeInputJsonSchema`, which drops defaulted keys from `required` and
+strips `additionalProperties: false` on every schema node it can reach.
 
-- `.default()`, `.prefault()` and `.catch()` fields are optional at every depth;
-- `stepInputContract` / `workflowInputContract` now return the same normalized
-  schema the manifest publishes, so a displayed contract (the Run form, inspect
-  tooling) agrees with what the engine enforces;
-- a step whose `inputSchema` contains a `.transform()` no longer fails the build
-  with "Transforms cannot be represented in JSON Schema"; it is described by its
-  input type, which is what a caller sends.
+**If you call `zodToJsonSchema` directly**, its output changes: a smaller
+`required` set, no `additionalProperties: false` from `z.strictObject()`, a
+`.pipe()`/`.transform()` described by its input type rather than its output
+type, and a schema containing a `.transform()` now converting successfully where
+it previously threw `Transforms cannot be represented in JSON Schema`. Every
+change is in the "what may a caller send" direction. If you need the
+value-a-parse-returns schema instead, call `z.toJSONSchema(schema)` yourself.
 
-Redeploy an agent to publish the corrected schema.
+Also:
+
+- `stepInputContract` / `workflowInputContract` return the same normalized
+  schema the manifest publishes, so a displayed contract (a Run form, inspect
+  tooling) agrees with what a pre-gate enforces.
+- `@sapiom/agent-runtime`'s step-input pre-gate normalizes the manifest it is
+  handed, so an agent deployed on an earlier version accepts a partial nested
+  input without being redeployed. A `.prefault()` field still needs a redeploy —
+  the older schema records no `default` keyword for it to key off.
--- a/packages/agent-runtime/src/manifest-validation.spec.ts
+++ b/packages/agent-runtime/src/manifest-validation.spec.ts
@@ -1,0 +1,82 @@
+/**
+ * The pre-gate normalizes the schema it is handed, so a manifest BUILT before
+ * the generator was fixed still accepts a partial input (SAP-3218) with no
+ * redeploy.
+ *
+ * Schemas here are written as literal JSON on purpose: they stand in for a
+ * stored manifest read back from the engine, which is exactly what this gate
+ * receives. (Constructing them with Zod would also load a second Zod module
+ * instance through the SDK's CJS build, which this package's test environment
+ * cannot do.)
+ */
+import { validateManifestStepInput } from './manifest-validation';
+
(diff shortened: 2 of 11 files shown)


gwitwer:
Thanks — all three findings actioned in ddef328f.

**1. The pre-gate now reuses the helper.** You're right, and this was the most valuable of the three: `manifest-validation.ts` calls `normalizeInputJsonSchema(schema)` instead of its private `relaxAdditionalProperties`, which deletes the duplicated recursion and fixes nested `.default()` fields for **already-deployed** agents with no redeploy. That's a bigger win than the generator fix alone, since the reported production agent is already out there. I did not defer it — the PR body advertising reuse with nobody reusing it was a fair hit.

`packages/agent-runtime/src/manifest-validation.spec.ts` pins it with a manifest written as literal JSON in the shape an older SDK published (`opts.required = ["verbose","retries"]`), asserting a partial input is accepted while a genuinely missing field, a wrongly-typed nested value, and a corrupted-payload case are still rejected. Literal JSON rather than Zod on purpose: constructing schemas here loads a second Zod module instance through the SDK's CJS build, which OOMs this package's jest environment — worth knowing independently of this PR.

`.prefault()` still needs a redeploy, exactly as you noted, and the JSDoc and changeset both say so.

**2. Changeset raised to `minor`.** Checked the precedent rather than taking it on faith: `0.13.0` was cut as a minor for "Add `PACKAGE_INVENTORY_PROTOCOL`, `packageInventorySchema`, and the public package-inventory types" — an additive export, same shape as this. It now also has a paragraph naming what changes for a direct `zodToJsonSchema` caller (smaller `required`, no `strictObject` marker, pipe/transform by input type, previously-throwing schemas now converting) and how to get the old output back (`z.toJSONSchema(schema)`). Added `"@sapiom/agent-runtime": patch` for the gate change.

**3. Real bug, and worse than "not new."** You're right that rule 2 made it worse, but I checked whether the case actually corrupts data rather than assuming, by running the old blind walk over the payload from your repro:

```
author default:   {"type":"object","properties":{"a":{"type":"string","default":"x"}},"required":["a"],"additionalProperties":false}
after blind walk: {"type":"object","properties":{"a":{"type":"string","default":"x"}},"required":[]}
```

Both rules fire on author data. Confirmed.

I did **not** take the suggested fix of skipping `default` / `const` / `enum` / `examples` values, because it has an inverse hazard: inside a `properties` map a property may legitimately be *named* `default` or `items`, and its value is a real subschema that must be normalized. Only position distinguishes payload from subschema, not name. So the traversal is now structural — it descends only through keywords that hold subschemas (single, list, and map forms enumerated), and copies everything else through untouched. Three specs cover both directions: a JSON-Schema-shaped `default` preserved verbatim, a JSON-Schema-shaped `.meta({ examples })` preserved verbatim, and a property *named* `default`/`items` still normalized. Your point about the JSDoc reasoning over property *names* and missing payload values was correct; that comment is rewritten to explain both hazards.

One consequence worth flagging for the record: the walk no longer strips `additionalProperties: false` from an unrecognized vendor keyword's subschema. In practice the only source is `z.strictObject()`, which appears at real schema nodes that the structural walk reaches — but it is a narrowing, so I'd rather name it than have it found later.

Verification: `pnpm build` / `pnpm typecheck` / `pnpm lint` all exit 0; `introspection.spec.ts` 21 passed, `manifest-input-gate.spec.ts` 8, `manifest.spec.ts` 25, `sdk.spec.ts` 56, `manifest-validation.spec.ts` 6. Full suite to CI.

https://claude.ai/code/session_01EyRQfxZWtu9BhJxx1dhDXu

gwitwer:
Fixed in 2346a590 — you're right, and it's the exact failure mode I'd called out in my own earlier note and then committed anyway. AJV is built without `useDefaults`, so `default` is annotation-only and cannot change a verdict; both assertions held either way, which makes that case worse than no test.

Dropped rather than rewritten. The property is only observable against the emitted schema, and it's already covered honestly there — `introspection.spec.ts`'s "never rewrites a JSON-Schema-shaped default value" inspects the emitted `default` directly. Asserting on `normalizeInputJsonSchema` from inside the gate's spec would just duplicate that a layer away from where it's observable. The spec docstring now records why this file doesn't cover it, so it doesn't get re-added by someone filling an apparent gap.

While there I checked the five remaining cases actually discriminate, rather than assuming. Against the previous relax-only gate:

```
OLD gate accepts partial nested input?  false
```

So the stored-manifest repair case does fail without the fix.

`pnpm build` / `pnpm typecheck` / `pnpm lint` all exit 0; `manifest-validation.spec.ts` 5 passed, and the four `@sapiom/agent` spec files unchanged at 21 / 8 / 25 / 56.

https://claude.ai/code/session_01EyRQfxZWtu9BhJxx1dhDXu

PR #853, opened 2026-09-07
docs(mcp): say an App Link is a redirector, not a reverse proxy [SAP-3217]

gwitwer:
## What

`sapiom_dev_app_publish`'s **description** and its **publish summary** now say the link is a wake-on-visit REDIRECTOR, not a reverse proxy: the root 302s to whichever preview URL is serving the app, sub-paths are not proxied, so the app's own API lives at that preview URL.

Three facts ship with it, because review showed the bare claim just relocates the confusion:

| Fact | Why it has to be there |
| --- | --- |
| Re-resolve per use, don't store | an org app's tokenized URL 401s *inside a single wake* once the token expires — "changes across wakes" does not predict that |
| Org-scoped app's API is **browser-only** | the redirect and `/__status` both run the link's own access check, which reads a browser session and no API key. But only for an org app — a public app's link and `/__status` need no session, so those are the machine route |
| `webhooksEnabled` is off by default | it is `false` on upsert, is not in this tool's schema, and the disabled path 404s indistinguishably from a typo'd slug |

`currentPreviewUrl` is deliberately never offered as an address: the backend's `publish()` leaves the binding stale at `wakeStatus = 'idle'`, so it is null in exactly the state this summary is read in, it can point at an expired sandbox, and reading it wakes nothing.

The summary branches on visibility, so a public app is not warned about a gate it does not have.

## Why

Studio feedback `7a59535a-97c8-4b30-bcff-dcf4004fbbdf`:

> It was not obvious that an App Link is a redirector rather than a reverse proxy. The root path answers a 302 to the current preview URL, and sub-paths are not proxied at all, so anything calling the app's API has to use the underlying address rather than the durable one.

The copy pushed the other way — "DURABLE Sapiom link", "safe to hand to a teammate" — which reads as a stable base URL for API calls too.

## The offline primer is deliberately NOT touched

An earlier revision of this PR synced `instructions.ts` to backend content release 2.11. That is now dropped, because the two bodies have drifted **in both directions** across two unrelated release pairs:

- **sapiom-js main** carries SAP-3180's vault / `agents.launch` / receipts-and-replay content (#815). Its backend half has not merged.
- **The backend constant** carries the trigger-kinds release. Its sapiom-js half has not merged.

Neither body is a superset, so copying either over the other deletes a shipped release's content. So the redirector paragraph goes into the backend's `DEFAULT_MCP_INSTRUCTIONS` (where the ticket asks for it) and into this tool's own copy, and it reaches the primer with that reconciliation rather than ahead of it. The reasoning is recorded in `tools/app-publish.ts`, in the backend migration, and behind a ⚠️ on `SAPIOM_JS_FALLBACK_DIGEST` so the next person does not "fix" it by copying one side over the other.

Worth noting the primer on main already says webhooks are "off by default" and that "no `sapiom_dev_*` tool sets it yet" — SAP-3180 reached the same conclusion independently.

## Verification

- `pnpm --filter @sapiom/mcp typecheck` — clean
- `pnpm --filter @sapiom/mcp test` — **149 passed** (15 files)
- `pnpm --filter @sapiom/mcp lint` — clean
- New spec cases assert the claims rather than phrasings, including the public-visibility summary branch the suite could not reach before (`mockHappyBackend` only ever returned an org-scoped link)

Changeset: `@sapiom/mcp` patch.

## Related

- sapiom/Sapiom#4926 — `sapiom_app_publish` description, publish result, and the primer paragraph as content release 2.11
- sapiom/docs-internal#153 — the "What the URL does" section on docs.sapiom.ai (**merged**)
- Linear: [SAP-3217](https://linear.app/sapiom/issue/SAP-3217)

https://claude.ai/code/session_01PAEku9HKnZUHmdnsQ3v5D1

### DIFF
--- a/.changeset/app-link-redirector-not-proxy.md
+++ b/.changeset/app-link-redirector-not-proxy.md
@@ -1,0 +1,5 @@
+---
+"@sapiom/mcp": patch
+---
+
+Say that an App Link is a redirector, not a reverse proxy. `sapiom_dev_app_publish`'s description and its publish summary now state that the link's root answers a 302 to whichever preview URL is currently serving the app, that sub-paths are not proxied, and that the app's own API therefore lives at the preview URL — which must be re-resolved per use rather than stored, because it changes when a wake recreates the sandbox and for an org-scoped app carries a short-lived token that expires. They also say who can take those routes: for an org-scoped app the redirect and `GET {link}/__status` both need a logged-in member's browser session rather than an API key, so its API is browser-only and a machine caller needs the app published `public` (whose link and `__status` need no session) or inbound traffic on `/hook/…`, which requires `webhooksEnabled`, off by default. The summary branches on visibility, so a public app is not warned about a gate it does not have.
--- a/packages/mcp/src/tools/app-publish.test.ts
+++ b/packages/mcp/src/tools/app-publish.test.ts
@@ -135,12 +135,17 @@ const jsonRes = (body: unknown, status = 200) => ({
 });
 
 /** The happy-path backend: upsert → bundle → publish. */
-function mockHappyBackend(): ReturnType<typeof vi.fn> {
+function mockHappyBackend(
+  overrides: Partial<typeof APP_LINK> = {},
+): ReturnType<typeof vi.fn> {
+  // `overrides` exists for `visibility`: the summary branches on it, and the default
+  // fixture is org-scoped, so the public branch is unreachable without this.
+  const link = { ...APP_LINK, ...overrides };
   const fetchMock = vi
     .fn()
-    .mockResolvedValueOnce(jsonRes(APP_LINK, 201))
+    .mockResolvedValueOnce(jsonRes(link, 201))
     .mockResolvedValueOnce(jsonRes(BUNDLE))
-    .mockResolvedValueOnce(jsonRes(APP_LINK));
+    .mockResolvedValueOnce(jsonRes(link));
   globalThis.fetch = fetchMock as unknown as typeof globalThis.fetch;
   return fetchMock;
 }
@@ -210,6 +215,39 @@ describe("sapiom_dev_app_publish tool", () => {
     expect(description).toMatch(/same slug again replaces the app in place/i);
     expect(description).toMatch(/TEXT-ONLY/);
     expect(description).toMatch(/10 MiB/);
+  });
+
+  it("says the link is a redirector, not a reverse proxy (SAP-3217)", () => {
+    // Studio feedback 7a59535a: "DURABLE link" plus "safe to hand to a teammate" reads
+    // as a stable BASE url, so an agent points the app's own fetches at sub-paths of it.
+    // The host serves the root (a 302), `/__status`, and `/hook`/`/hook/*` — every other
+    // sub-path is its 404, never the app. Assert the claim, not one phrasing of it.
+    const { description } = setup();
+    expect(description).toMatch(/REDIRECTOR, not a reverse proxy/);
+    expect(description).toMatch(/sub-paths are NOT proxied/);
+    expect(description).toMatch(/__status/);
+    expect(description).toMatch(/\/hook\//);
+    // Plus the facts that keep the bare claim from misleading in a new way. Re-resolve
+    // rather than store: an org app's tokenized URL 401s inside a single wake, which
+    // "changes across wakes" does not predict. `/hook/` forwards only for a link with
+    // `webhooksEnabled`, off by default and not in this schema.
+    expect(description).toMatch(/re-resolve it per use/);
+    expect(description).toMatch(/short-lived token that expires/);
+    expect(description).toMatch(/webhooksEnabled.*off by default/s);
+    // And WHO can take the recovery routes, scoped correctly. The gate reads a browser
+    // cookie and no API key, so an org app's API is browser-only — but the backend
+    // authorizes a public + browser resolve with no token at all, so a public app's link
+    // and `__status` need no session, and stating the gate unconditionally would rule out
+    // the one route a cron job has. `currentPreviewUrl` is deliberately never offered as
+    // an address: it is null right after a publish, and reading it wakes nothing.
+    expect(description).toMatch(/browser session/);
+    expect(description).toMatch(/NOT an API/);
+    expect(description).toMatch(/browser-only/);
+    expect(description).toMatch(/for an org-scoped app that means/);
+    expect(description).toMatch(
+      /public app's link and `\/__status` need no session/,
+    );
+    expect(description).not.toMatch(/currentPreviewUrl/);
   });
 
   it("is not authenticated without a cached credential, and makes no call", async () => {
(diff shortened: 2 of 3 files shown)


gwitwer:
All three findings were real. Fixed in 6df6b1ad, and the branch is now synced to backend content release **2.11** (not 2.9 — see the last note).

**1. `/hook/` unconditional — agreed, and it was the worst of the three.** Verified: `AppLinksService.upsert` sets `webhooksEnabled: data.webhooksEnabled ?? false`, the flag is in `APP_LINK_MANAGEMENT_FIELDS`, and it is absent from this tool's schema — so every app published here has it off, and the disabled path answers a 404 that `routes/hook.ts` deliberately makes indistinguishable from an unknown slug. Your framing was the deciding one: this traded a wrong inference for a wrong *instruction*, in copy that cannot be edited after publish.

Took the **copy** option, not the schema one. Adding `webhooksEnabled` to the tool is a capability change, not a docs fix, and as you note it would then collide with `assertMayConfigure` on the republish branch and need its own error copy. All three surfaces now say the condition and that this tool cannot set it.

**2. `__status` shape — agreed, and it led somewhere bigger.** Chasing the response shape surfaced a larger problem the backend reviewer independently hit: `routes/gate.ts` verifies an Auth0 cookie and *nothing else*, and `__status` shares that gate on purpose ("its `ready` payload is the tokenized preview URL, so it must be exactly as hard to reach as the link"). So both routes I offered were browser-only — addressed to "anything that calls the app's own API", which is precisely the caller holding an API key instead of a session. The copy now names the field (`url`, not `previewUrl` — worth flagging since your suggestion used the latter) *and* says an org-scoped app's API is browser-only, with the two routes a machine actually has: publish the app `public`, or receive traffic on `/hook/…`.

**3. Token expiry — agreed.** "Not stable across wakes" does not predict a 401 minutes into a single wake, and org-scoped is the default. All surfaces now say re-resolve per use rather than store, and give the reason.

The summary also branches on visibility now, so a public app is not warned about a gate it does not have.

**On the 2.9 → 2.11 change:** the backend cut this as 2.9 at `1789800000000`; release 2.10 (trigger kinds) then merged ahead of it, which would have left the lower-timestamped migration running last and silently dropping this paragraph. Re-sequenced to 2.11 at `1790000000000` and re-snapshotted on 2.10's body, so this fallback goes 2.8 → 2.11 in one step and keeps 2.10's Triggers section verbatim. 2.10's own sapiom-js pairing has not landed here yet — whichever of the two merges second rebases onto the other's digest rather than re-pointing the pin alone, and the pin comment says so.

You were right that CI is the gate on the sync: `pnpm --filter @sapiom/mcp test` is 148/148 green locally with the digest recomputed independently from this body to `4df868d9…`, matching the backend pin.

gwitwer:
Findings 2 and 3 fixed, finding 1 partly — the changeset and the test, not the sequencing. Plus a test-loss you caught indirectly. All in bf57acde.

**3. Public branch untested — agreed, and it was worse than untested.** `mockHappyBackend` only ever returned an org-scoped link, so the public branch was unreachable and nothing would have caught it telling a public app *"a machine caller needs the app published as `public`"* — advice already taken. Added a public-visibility case; `mockHappyBackend` now takes an overrides object for exactly that field. 150 tests, up from 148.

I did **not** harden `asAppLink` to require `visibility`. It gates whether the response is usable as a link at all (its `id` addresses the next two calls), and adding a field requirement there would turn a summary-wording concern into a publish-failure path. The backend DTO always sends `visibility`; the test is the right guard for the branch.

**2. `currentPreviewUrl` — agreed, and the backend reviewer found a stronger reason than citability.** It is a *binding* field: `publish()` sets `wakeStatus = 'idle'` and leaves the binding values alone, so on a first publish it is **null in exactly the state the summary is read in**, it can point at a sandbox that has since expired, and reading it wakes nothing. Not just uncited — wrong. Gone from the description, the summary and the primer, with a `.not.toMatch(/currentPreviewUrl/)` on the description so it cannot come back.

That also exposed the sentence that made it load-bearing: the gate was stated unconditionally, but the backend authorizes a public + browser resolve with no token at all, so a public app's link and `/__status` need no session. Browser-only is now scoped to the org-scoped case, with the public route named.

**Test loss you surfaced by asking:** checking your finding 3 showed the *description* assertions from the previous round were never in the file — a later assert in the same edit aborted the write, so only the summary patch landed. Restored, plus the new scoping claims.

**1. Triggers section — agreed on both actionable parts, pushing back on the framing.**

Done: the changeset now names the section, the four kinds, the signing recipe and the secret tool, and says they are gated on `>= 0.15` and not registered here. And there is now a test pinning `## Triggers` and the `>= 0.15` gate, since you are right that nothing asserted it and the digest goes green on any 2.11 body.

Not done: deferring to 2.10's own pairing. Two reasons.

The first is that it is not available. My backend PR sets `SAPIOM_JS_FALLBACK_DIGEST` to 2.11's digest; the pin requires this body to *be* 2.11. Splitting them re-opens the drift SAP-2959 closed.

The second is that "every consumer of this release reads the four-kind claim" is already true without this PR. `fetchInstructions` hits `GET /v1/mcp/instructions` at startup and serves that; `AUTHORING_INSTRUCTIONS` is the fallback *only* on a non-200, empty body, network error or timeout. Backend 2.10 is merged, so an online session on 0.14.0 reads the four kinds today. Leaving the fallback on 2.8 does not withhold the claim — it makes the offline path contradict the online one, which is the failure mode that motivated the pin in the first place (a session that fell back offline was told App Links do not exist). The `>= 0.15` gate is the mitigation, and it is the same pattern 2.8 used for `sapiom_dev_app_publish`.

**Round 1 correction noted and appreciated** — `url` vs `previewUrl`. Flagged it in my last reply for the same reason you flagged it here.

`typecheck` and `lint` clean, 150/150, primer digest independently hashed to `8d673ba1…` matching the backend pin.

gwitwer:
Rebased onto main (#815, SAP-3180 landed), and the rebase changed what this PR should do.

**Dropped the primer sync.** The previous revision synced `instructions.ts` to backend content release 2.11. That is no longer the right move, because the two bodies are now drifted **in both directions**:

| Body | Has | Missing |
| --- | --- | --- |
| sapiom-js main (#815) | SAP-3180's vault / `agents.launch` / receipts-and-replay | the backend's trigger-kinds release |
| backend `DEFAULT_MCP_INSTRUCTIONS` | the trigger-kinds release | SAP-3180's content |

Neither is a superset. Syncing this repo to 2.11 would have deleted #815's vault/receipts/`agents.launch` sections from the fallback — a shipped release, and not mine to remove. Syncing the other way would drop the trigger kinds. So the redirector paragraph lands in the backend constant (which is the file SAP-3217 actually names) and in this tool's own copy, and reaches the primer with the reconciliation rather than ahead of it.

That also retires the Triggers-section finding from the last round: this PR no longer touches the primer, so there is nothing to disclose in the changeset and no Triggers assertion to pin. Both are removed. The reasoning is recorded in `tools/app-publish.ts`, in the backend migration header, and behind a ⚠️ on `SAPIOM_JS_FALLBACK_DIGEST` telling the next person not to reconcile by copying one side over the other.

**Two things worth flagging from #815:**

1. Its App Link paragraph independently says webhooks are "off by default" and that "no `sapiom_dev_*` tool sets it yet" — the same conclusion as this PR's round-1 correction, reached separately. Good sign the correction was right.
2. The backend halves of both release pairs are outstanding. Reconciling is landing those two, then moving both pins in one pair of PRs. Not this ticket.

**What survives from the review rounds**, all in `app-publish.ts` and its tests: the redirector claim; re-resolve-don't-store with the token-expiry reason; the session gate scoped to org apps with the public machine route named; `webhooksEnabled` off by default and not settable here; `currentPreviewUrl` never offered as an address. Plus the public-visibility summary case the suite could not reach before.

`typecheck` and `lint` clean, 149/149 (down from 150 — the Triggers assertion went with the primer revert).

gwitwer:
CI is green. The `test (20.x)` / `test (22.x)` failure was a flaky test in `packages/harness`, not this change — I re-ran the failed jobs with no code change and both passed.

**What failed:** `src/core/subsession-coordinator.test.ts > shares one batch wait deadline across readiness and adapter identity`. `packages/mcp` passed 15/15 in the same run, and this PR's diff is three files — `packages/mcp/src/tools/app-publish.ts`, its test, and a changeset. Nothing in `packages/harness`.

**Why it flakes:** the case arms a real `setTimeout(… , 50)` that calls `setReady`, then asserts the readiness path *times out* against `batchWaitTimeoutMs: 100` / `readinessTimeoutMs: 250`. It is asserting that the 50 ms timer loses a race with the 100 ms deadline. On a loaded runner — the harness suite is 236 files and runs `vitest run` twice, including the perf config — a 50 ms timer landing before a 100 ms deadline is well within normal scheduling jitter, and then `sessionState` is not `awaiting-ready` and the assertion fails.

**Evidence it is not mine:**

| Check | Result |
| --- | --- |
| Targeted, 5 runs | 5/5 pass |
| Whole file, 3 runs | 27/27 pass each time |
| Re-run of the failed CI jobs, no new commit | both Node versions pass |
| My diff vs `packages/harness` | no overlap |

I have **not** touched the test. Retiming someone else's coordinator test inside a docs PR is the wrong place for it, and the fix is a judgement call for whoever owns that file — fake timers, or a batch wait shorter than the readiness timer rather than longer. Flagging it here so it can be picked up separately; it will keep costing re-runs on unrelated PRs until then.

PR #854, opened 2026-09-07
fix(tools): resolve a refused child dispatch from run, throw a typed error from launch [SAP-3219]

gwitwer:
## Problem

`agents.run` / `agents.launch` (surfaced in steps as `ctx.sapiom.agents.run`) was documented as **"failure is data, not an exception — resolves on any terminal status, does not throw."** That held only for a child that was *dispatched* and then failed. When the **dispatch itself** was refused, the call threw an untyped `Error`:

- unknown slug → 404 from `POST /agents/v1/definitions/:slug/executions`
- input refused by the engine's pre-gate (`StepInputValidationError` → 400 `step_input_invalid`)
- network / transport fault

A step throw is not an exception the author simply catches — it feeds the engine's retry machinery. So a coordinator written straight from the authoring guide lost its **entire run** to one bad slug: the step retried a guaranteed-identical failure to `maxAttemptsPerStep` and then failed. Studio feedback `2c3afb03-9cdd-4f0a-b58e-53172810c367` (production, sapiom-mcp 0.14.0). The workaround was to wrap every call in try/catch and route the throw into the same branch as `status: "failed"`.

`wait()` also threw on poll timeout, so even the awaited path had a second way out.

## Approach

The ticket offered two options: make the contract true, or make the docs true. Review (@xoxo-david) surfaced a third, better one, and the PR now implements it — the two entry points report a refused dispatch according to what each **owes its caller**:

| | Refused dispatch | Why |
| --- | --- | --- |
| **`run()`** | Resolves `status: "rejected"` | Owes a *result*. Its result is already discriminated on `status`, and to a coordinator "the child failed" and "the child never started" are one fact: this stage did not deliver. So `status !== "completed"` stays the single branch and a fan-out coordinator needs no try/catch. |
| **`launch()`** | Throws `AgentDispatchError` | Owes a *pausable handle*. A dispatch that created no child has none — an object with a null `executionId` that cannot be paused on would misrepresent what it is. |

An earlier revision of this PR had `launch` resolve a dispatch-less handle carrying a `rejection` field. That forced every author to branch on `child.rejection` and left anyone who skipped the check holding a handle that lies about being one. The split is both better DX and a smaller diff.

## What changed

**`@sapiom/tools` — `agents`**

- `run()` resolves `{ executionId: null, status: "rejected", output: null, error: AgentRunError }`. `AgentRunError` is `{ code, message, status, details }` — `code` is the coarse bucket (`not_found` / `invalid_input` / `http` / `transport` / `timeout`), `details` keeps the platform's own response body so its stable code and validation issues survive.
- `launch()` throws `AgentDispatchError` (root-exported; `.code`, `.status`, `.details`, and `.toRunError()` for one shape across both call styles). Every handle it returns is pausable.
- `wait()` no longer throws: `timeoutMs` resolves `"timed_out"`; a status read the platform refuses, or one that keeps faulting, resolves `"unknown"`. Both keep `executionId` — the run exists and can be checked on later. A transient 5xx/transport fault is ridden out for a bounded number of consecutive polls rather than spending the whole `timeoutMs` on doomed requests.
- Only `"rejected"` means nothing is running, so it is the only outcome on which re-dispatching is safe. `"unknown"` and `"timed_out"` name a child that may still be working.

**`Transport.request`** throws a `TransportHttpError` carrying status + parsed body (message text unchanged) so a capability can classify by status instead of scraping a string. Marked `@internal` — capabilities translate it into their own author-facing error.

**Stub / `run_local`** — `agents.launch` resolves through the overrides (`agents.launch` → `agents.run`, merged over defaults, matching `models.coding.launch`). It previously built a completed run unconditionally, so the try/catch the docs require was impossible to cover locally. A stubbed `{ status: "rejected" }` throws from `launch` and resolves from `run`, mirroring the real split.

**Docs** — the `agents` README documents every status and shows the try/catch; the authoring skill (all four synced copies) says what each entry point does and how to stub both branches. Delayed dispatch (`at`) is called out: a refused delayed launch throws like any other, and on a *successful* one `status`/`wait`/`run` throw because there is no run until the scheduled time.

**`@sapiom/agent` is untouched.** The earlier revision needed a `pauseUntilSignal` guard against dispatch-less handles; once those cannot exist, it doesn't.

## Types

Breaking at the type level, no-op for code that already branches on `status`:

| | before | after |
| --- | --- | --- |
| `AgentRunResult.executionId`, `RunHandle.executionId` | `string` | `string \| null` (a delayed handle reports `null`, not `""`) |
| `AgentRunResult.status` | `ExecutionStatus` | `AgentRunStatus` (adds `"rejected"`, `"unknown"`, `"timed_out"`) |
| `RunHandle` | `extends DispatchHandle` | unchanged — every handle stays pausable |

`AgentRunSpec`, `AgentRunResult`, `AgentRunStatus`, `AgentRunError`, `AgentRunErrorCode`, `AgentRunHandle`, `AgentExecutionStatus` and `AgentDispatchError` are now root exports, matching the `models` equivalents.

`AgentRunResultPayload` / `agentResultSchema` are deliberately unchanged: a rejected dispatch never creates a child, so it never produces a resume payload. Noted in the JSDoc so the divergence is explicit.

## Known gap, deliberately not fixed here

Uncaught, `AgentDispatchError` is an ordinary step throw, so the engine still retries a deterministic slug typo to `maxAttemptsPerStep`. Making it terminal-without-retry needs the engine's non-retryable set to admit it — `packages/agent/src/non-retryable-step-error.ts` is a deliberately closed, platform-owned set, so that is engine-side work in the Sapiom repo (`apps/workflows-engine`, `packages/workflows-step-runner`). `StepInputValidationError` is already in that set, so extending it to a child's rejected input is consistent with a call already made.

`run()` — the reported customer path — no longer throws at all, so it is unaffected by that gap.

## Verification

```
pnpm build      # exit 0
pnpm typecheck  # exit 0
pnpm lint       # exit 0
```

`pnpm terminology:check`, `pnpm provider-copy:check` and `pnpm examples:check` pass. Per the repo's jest rule, specs were run one file at a time with a single worker:

| spec | result |
| --- | --- |
| `packages/tools/src/agents/index.spec.ts` | 23 passed |
| `packages/tools/src/stub/agents-launch-keys.spec.ts` | 8 passed |
| `packages/tools/src/agents/schedule-launch.spec.ts` | 3 passed |
| `packages/tools/src/stub/models-launch-keys.spec.ts` | 7 passed |
| `packages/tools/src/stub/stub-calls-sink.test.ts` | 12 passed |
| `packages/tools/src/models/launch.spec.ts` | 12 passed |
| `packages/agent/src/constructors.spec.ts` | 10 passed |
| `packages/agent/src/sdk.spec.ts` | 56 passed |

Coverage added: 404 / 400 / transport / other-non-2xx on dispatch, non-JSON error bodies, `launch` throwing the typed error, the thrown error and `run`'s data reporting identically, a refused delayed dispatch, `wait()` timeout, a 4xx during polling, polling through a transient 5xx, the bounded consecutive-fault cap, and the stub's override precedence + both stubbed branches.

Fixes SAP-3219 — https://linear.app/sapiom/issue/SAP-3219

https://claude.ai/code/session_01W5722UypH5L2JbDWrDPRCx

## Summary by CodeRabbit

- **New Features**
  - Added structured agent dispatch and execution outcomes, including rejected, timed-out, and unknown statuses.
  - `agents.run` now returns refused dispatches as data; `agents.launch` throws a typed dispatch error.
  - Added structured transport error details and top-level exports for agent run types.
  - Improved polling resilience for transient failures.

- **Documentation**
  - Expanded guidance for handling dispatch failures, retries, scheduled launches, and local stubs.

- **Breaking Changes**
  - Execution IDs may now be `null`, and status values include additional outcomes.


gwitwer:
Fixed all four findings and the three nits in 311ccf08. Two I'd have shipped as-is; the reasoning is below.

### 1. `"rejected"` overloaded — fixed, and it was the real bug in the PR

You're right, and the failure scenario is the one that matters: a key expiring mid-poll resolving as "nothing was dispatched" invites an author to double-run a live child. I took the first of your two options — a separate status — rather than loosening the docs, because the guarantee is worth stating rather than dropping:

- A failed status read now resolves `status: "unknown"`, keeping `executionId`.
- `"rejected"` keeps its narrow meaning: no run was ever created, `executionId` is `null`.
- All three doc sites say which statuses name a live run, and the README gained an explicit bullet: **only `"rejected"` is safe to re-dispatch**; `"unknown"` and `"timed_out"` name a child that may still be working, so retry those with an `idempotencyKey` or not at all.

`if (status !== "completed")` still covers everything for the common branch — only re-dispatch logic has to tell the three apart.

### 4. Unbounded polling — fixed

Also right, and it undercut the PR's own argument: spending the full hour on doomed requests and then dying on the step's own timeout is a worse outcome than the throw I removed. Five consecutive faults now resolve `"unknown"` with the last error; a 4xx still gives up at once; the counter resets on any successful read, so a blip is still ridden out. I chose a flat cap over backoff — with `pollMs` at 3s that's ~12s of tolerance, and backoff would need its own tuning story for no gain here.

This also fixed your first nit for free: `"timed_out"` now only happens when the polls were succeeding, so it always carries `code: "timeout"` and never inherits an `http`/`transport` code an author matching on the code would miss.

### 2. Changeset understates the break — fixed

Correct, and the distinction matters: the type-level break I acknowledged is caught at compile time, but this one is silent at runtime. The old README documented the result as `"completed" | "failed"`, so `status === "failed"` was the *documented* spelling, and a consumer who wrote it now falls through the branch and feeds `null` downstream — the exact failure the skill text warns about. The changeset has a **Breaking** call-out with the one-line migration.

### 3. Unexported types — fixed, split two ways

The `Agent*` types are now root exports (`AgentRunSpec`, `AgentRunResult`, `AgentRunStatus`, `AgentRunError`, `AgentRunErrorCode`, `AgentRunHandle`, `AgentExecutionStatus`), matching `ModelRunError`/`ModelRunStatus`/`ModelRunHandle`. `RunHandle` is aliased to `AgentRunHandle` because `models` already owns that name — same aliasing `client.ts` and the stub already use.

`TransportHttpError` I took your second option: dropped from the changeset, marked `@internal`. Root-exporting it would establish a *cross-capability* transport-error contract, which is a bigger public statement than this ticket should make — the per-capability classes (`SearchHttpError`, …) are the existing shape, and authors get `AgentRunError` with `code`/`status`/`details` without needing the transport class. Announcing something importable from nowhere was the actual defect, and that's gone.

### Remaining nits

- **`launchScheduled` returning `""`** — fixed to `null`. Pre-existing, but my type widening made it a third sentinel under a JSDoc that says `null` means no run, so it belonged in this PR. New spec covers a successful delayed dispatch: pausable, `dispatch` set, `executionId` null.
- **Skill missing the `at` carve-out** — fixed in all four synced copies (still byte-identical). You're right that the skill is the surface authors actually read.

### Verification

`pnpm build` / `typecheck` / `lint` exit 0; `terminology:check` and `provider-copy:check` pass. `agents/index.spec.ts` 22 passed (3 new: the `unknown` reclassification, the consecutive-fault cap asserting it stops at 5 rather than ~1200, and `timed_out` keeping `code: "timeout"` after an earlier fault). `schedule-launch.spec.ts` 3, `constructors.spec.ts` 12. Per the repo's jest rule these were run one file at a time with a single worker; CI covers the rest.

gwitwer:
Both fixed in 1b3e4bd2. No disagreements this round — the stub gap was a real hole I opened.

### Stub couldn't reach the branch the docs require — fixed

You're right, and the reasoning is the part that convinced me: a branch no local test can cover is a branch that rots. I'd added guidance telling every author to write `if (child.rejection)` while leaving `run_local` unable to produce a rejection, and `agents.launch` was the only dispatched launch in the stub not resolving through overrides.

`launch()` now consults `agents.launch` → `agents.run`, merged over the built-in defaults — same precedence and merge behavior as `models.coding.launch`, so this follows the shape `models-launch-keys.spec.ts` already pinned rather than inventing one. A stubbed `status: "rejected"` produces the real client's shape: no `dispatch`, `executionId: null`, the rejection on `handle.rejection`, and no registered resume payload (nothing would ever fire it).

Two things I did beyond the literal ask, both because they'd otherwise be the next bug in the same place:

- **The merge applies to `agents.run` too.** The real client's `run()` *is* `launch().wait()`, so if only `launch` merged, a partial stub would give the two paths different results — the exact drift the `models.coding` comment warns about (`#664`). `executionId` follows the client's guarantee rather than the override's omission: `null` only on `"rejected"`, a real id otherwise.
- **The resume payload follows the stubbed status.** It was hardcoded `completed`, so a paused step's own `failed` branch wasn't testable either. Non-completed statuses resume as the `failed` variant, since that union is `completed | failed` only.

New spec `packages/tools/src/stub/agents-launch-keys.spec.ts`, 7 cases: the unstubbed default, the shared-key override, the precedence order, the rejection branch end to end, partial-override fill-in, and the resume payload both registering with the right status and *not* registering for a rejection.

I also added the `agents.run`/`agents.launch` keys to the skill's stub-naming rules with the two values worth stubbing — that section is where an author looks to test these branches, and it listed `models.*` but not `agents.*`.

### Nit: `lastStatus = "unknown"` — fixed

Renamed to `"unread"`. Good catch: `"unknown"` became a real status in the last round, so a timeout that never managed a single successful read was emitting `last status: unknown` inside a `status: "timed_out"` result — two different meanings of the word in one value.

### Verification

`pnpm build` / `typecheck` / `lint` exit 0; `terminology:check` and `provider-copy:check` pass. Specs, one file at a time with a single worker per the repo's jest rule: `stub/agents-launch-keys` 7, `stub/models-launch-keys` 7, `stub/stub-calls-sink` 12, `stub/content-generation` 11 (the last three to check the stub blast radius), `agents/index` 22, `agents/schedule-launch` 3, `agent/constructors` 12. Diff still confined to the intended files, no formatter churn. Changeset notes the stub change.

xoxo-david (commented):
i'm not sure making dispatch never throw is the right DX, it now opens up having to always branch on checking `child.rejection` and manually returning a fail signal or have downstream errors trying to use a rejected launch as if it dispatched correctly

why not let these throw exceptions which the author can choose to handle in `try/catch` with a typed error object and identify terminal errors so they don't automatically retry in the engine? this also keeps the door open for future `onError` or other lifecycle hooks we could give to authors

gwitwer:
@xoxo-david you're right about `launch()`, and I'd already come to the same place independently — the pause path is the half of this that the "failure is data" framing doesn't actually fit.

Splitting your two points, because I think they land differently on the two entry points.

### `launch()` — agreed, the handle is the problem

A rejected dispatch produced no child, so a `RunHandle` is the wrong return value: it can't be paused on, `executionId` is `null`, and the only thing it carries is the rejection. That's an object that fails to be the thing its type says it is, which is what makes the mandatory `child.rejection` check feel bolted on. Your "downstream errors trying to use a rejected launch as if it dispatched correctly" is the real failure mode — I guarded the `pauseUntilSignal` path specifically (it throws rather than parking the step on a signal nothing will fire), but any other use of that handle is still nonsense, and guarding one call site isn't a design.

Worth being explicit that my version isn't safe-by-default either: skip the `rejection` check and `pauseUntilSignal` throws, which retries to cap — the original bug. So both designs need author action on the launch path. Given that, a typed throw is the better mechanism: more discoverable, no phantom object, and it composes with the `onError` / lifecycle-hook direction, which I hadn't weighed at all.

### `run()` — I'd keep this one as data

`run()` returns a result already discriminated on `status`, where a child that ran and failed is `status: "failed"`. To a fan-out coordinator, "the child failed" and "the child never started" are one thing: this stage didn't deliver. If dispatch throws while child-failure resolves, an author dispatching six children writes six try/catch blocks that each route into the `status !== "completed"` branch they already have. That's the workaround the ticket reports customers already writing by hand.

### So: throw on `launch`, resolve on `run`

That's smaller than what's on the branch — it deletes `rejectedHandle`, `RunHandle.rejection`, `MaybeDispatchHandle`, and the whole `@sapiom/agent` change, dropping a package from the diff. `AgentRunError` survives as the thrown error's typed payload. Happy to cut it that way.

### One cost to know about on the terminal-error half

The mechanism exists but is deliberately closed. `packages/agent/src/non-retryable-step-error.ts:9`:

> The closed set of platform-owned step errors that may bypass workflow retry. Unknown codes and arbitrary `retryable: false` properties are intentionally excluded and therefore retain the legacy retry behavior.

So a thrown dispatch rejection is retryable-to-cap until it's added to that set, which means engine-side work in the Sapiom repo (`apps/workflows-engine`, `packages/workflows-step-runner`), not just an SDK change. Until then an uncaught `launch()` throw burns `maxAttemptsPerStep` on a deterministic slug typo, which is the reported bug.

Argument in favor of doing it: `StepInputValidationError` is already in that closed set. The platform has decided that "input the pre-gate refused" is terminal-not-retryable when it's this step's own input — extending that to a child's rejected input is consistent with a call already made, not a new concept.

### What I'd like your read on

Whether to land the `launch`-throws split in this PR and file the engine-side terminal-error work as a follow-up, or hold this PR until both halves can ship together. The first unblocks the `run()` fix (the actual customer report — a coordinator using `run`) but leaves `launch` briefly worse than the status quo for anyone who doesn't catch. The second is more correct and slower.

Also flagging a related split you may have opinions on, since you'd be living with it: `agents` returning dispatch failure as data while the sibling `models` throws `CodingRunHttpError` (`models/index.ts:370`) is worse than either convention applied consistently. Your proposal happens to resolve it in `models`' favor.

gwitwer:
Correcting one thing in that last comment, since it was the main argument against landing the split now.

I said the `launch`-throws split "leaves `launch` briefly worse than the status quo for anyone who doesn't catch." That's wrong. On `main` today `launch` throws an untyped error that retries to cap. A typed throw has identical retry behavior with a better error object, so it's neutral-to-better against `main` — it's only worse than the *current branch*, which resolves instead of throwing.

So the engine-side terminal-error work is a strict improvement to sequence after, not a prerequisite. That makes landing the split now the cheaper option than I presented it as.

gwitwer:
@xoxo-david implemented your proposal in 35098364 — `launch` throws, `run` keeps reporting as data. Pushed rather than waiting on the sequencing question, since it turned out to be the smaller change and doesn't depend on the engine work.

**What landed**

- `launch()` throws `AgentDispatchError` — typed, carrying `code` / `status` / `details`, plus `toRunError()` so you get one shape whichever call style you use. Every handle it returns is pausable again.
- `run()` unchanged from the previous revision: resolves `status: "rejected"`. It owes a result, and a result can carry the rejection, so a fan-out coordinator still needs no try/catch for the case the ticket reported.
- Deleted: `rejectedHandle`, `RunHandle.rejection`, `MaybeDispatchHandle`. `RunHandle extends DispatchHandle` again, and **`@sapiom/agent` reverts to `main` entirely** — the `pauseUntilSignal` guard existed only to catch dispatch-less handles, which can no longer exist. 15 files → 13, one fewer package.
- The stub mirrors the split, so a stubbed `{ status: "rejected" }` throws from `launch` and resolves from `run`. Your catch block runs under `run_local` instead of waiting for production.

**On the terminal-error half:** not in this PR, and the changeset + PR body both say so. Uncaught, the throw still retries to `maxAttemptsPerStep`, so the docs tell authors to catch it. Whenever you want to pick that up, `StepInputValidationError` is already in the closed non-retryable set, so admitting a child dispatch rejection follows a decision the platform has already made — happy to take the engine-side ticket if you'd rather not.

**Still open for your call:** the `agents`-vs-`models` inconsistency. `launch` throwing narrows it a lot — both capabilities now throw a typed error on a refused dispatch — but `agents.run` reports as data where `models.run` throws. I think that one is defensible (a child *workflow* result already models failure as a status; a model call doesn't), though you own both surfaces, so say if you'd rather they matched exactly.

CI is re-running against the new HEAD, and this supersedes the earlier bot approval.

gwitwer:
CodeRabbit review addressed in 052aa1d2 — all five findings accepted, all five were in code this PR introduced. No pushback on any of them.

### 1 + 3 (Major) — transport faults reported as "nothing was created"

Correct, and the most serious thing in the PR. `run` mapped *every* dispatch failure to `status: "rejected"`, which the README documents as "proves nothing is running, safe to re-dispatch." A lost response doesn't prove that: the platform can accept the request and create the child before the response goes missing. An author following my own guidance would have started a second copy of a live child.

This is the same defect an earlier review round caught on the poll path — `"rejected"` carrying two meanings — so the fix is the same shape:

| Dispatch outcome | Status | `executionId` |
| --- | --- | --- |
| 404, 400/422, 401/403 — platform proved it created nothing | `"rejected"` | `null` |
| 5xx, or a lost response (no status) — ambiguous | `"unknown"` | `null` (no id ever came back) |

`AgentDispatchError` gained `childMayExist` so a `launch` caller can draw the same line from the throw. The README, all four skill copies and the changeset now say that on an ambiguous outcome there is no `executionId` to check, so an `idempotencyKey` **on the original call** is the only thing that makes a retry safe — which is the practical advice your finding implies.

### 2 (Minor) — 408/429 during polling

Right, and cheap. `status < 500` gave up immediately, so one rate-limit answer could end an hour-long `wait()`. Both statuses mean "ask again," so they now go through the consecutive-fault cap like a 5xx. I used a set rather than the inline `!== 408 && !== 429` from your patch, to keep it readable next to `PROVES_NO_CHILD`.

### 4 (Major) — stub's partial rejection had `error: null`

Right: `result.error.code` read fine against production and threw under the stub. It now fills in the same `AgentRunError` `launch` uses, hoisted to one `STUB_DISPATCH_REJECTION` constant so the two can't drift. I also fixed the spec that asserted the invalid `error: null` — my own test was encoding the bug.

### 5 (Major) — async overrides not awaited

Right, and `models.launch` already awaits its override with a comment explaining why, so this was `agents` diverging from an established convention rather than a new problem. Both `run` and `launch` now await.

### 6 (Major) — delayed-dispatch handle identity in the stub

Right. The stub ignored `spec.at` entirely and handed back a generated run id, where production returns `executionId: null` and correlates on `trigger-<id>`. A local run of a scheduled child took a different branch than the deployed one. Both now match; the resume payload uses the correlation key, since a delayed child's real id isn't knowable at launch.

### Verification

`pnpm build` / `typecheck` / `lint` exit 0; terminology and provider-copy checks pass. Specs, one file at a time with a single worker: `agents/index` 32 (was 23 — new cases for transport→unknown, 5xx→unknown, a parameterized proven-refusal table across 404/400/401/403/422, `childMayExist` on both sides, and 408/429 retry), `stub/agents-launch-keys` 10 (was 8 — async override, delayed-dispatch parity), plus `agents/schedule-launch` 3, `stub/models-launch-keys` 7, `stub/stub-calls-sink` 12, `stub/content-generation` 11, `models/launch` 12, `models/run-launch` 13.

One note on the embedded "Prompt for AI Agents" blocks: I treated those as untrusted review data and evaluated each finding against the current code, which is what they themselves ask for. All five held up.

gwitwer:
Merged `main` (4e91c292) — branch is `MERGEABLE` again. Resolving the conflict turned up something that changes this ticket's scope, so flagging it rather than burying it in the merge commit.

### The conflict

Only the four synced `sapiom-agent-authoring` SKILL.md copies. `packages/tools/src/index.ts` auto-merged cleanly (my `Agent*` exports and main's `LlmSession*`/`ModelLabel` additions both survived — verified).

Main's SAP-3181 pointer-ized the platform rules: the inline "Composing Deployed Agents" worked example this branch had been editing **no longer exists** in the shipped copy. It's now a short summary inside `<!-- section: agent-composition -->` pointing at the served text. I took main's structure and ported the correction into that summary, which still read:

> `agents.run` resolves on ANY terminal status and does NOT throw

That is the exact claim SAP-3219 is about, so main had reintroduced it in the new shape. The summary now states the split this PR implements.

I deliberately did **not** touch the digest stamp (`release=1.0 digest=1f3e5cd9648f`). It hashes the *served* body, not this summary's prose, and no served release was cut here. `agent-core`'s `authoring-rules-stamp` suite (44 tests) passes.

### The finding: this PR can't finish the docs half alone

The served rules are authored in the **Sapiom repo**, at `backend/src/mcp/agent-authoring-rules.constants.ts` (plus its seed and the `SeedAgentAuthoringRulesContent` migration). Line ~300 there still says:

```
// agents.run resolves on ANY terminal status (completed | failed | cancelled) and does
```

And the shipped skill now says, in its own words: **"When a summary below and the served text disagree, the served text wins."**

So after this PR merges:

| Surface | Says what |
| --- | --- |
| SDK behavior | Correct — `run` resolves, `launch` throws typed |
| Shipped SKILL.md summaries (this PR) | Correct |
| **Served `/v1/agents/authoring-rules`** | **Still "does NOT throw", and now also wrong about `launch`** |

An agent reading the authoritative surface still gets the guidance that produced the original Studio report. Worse than before in one respect: `launch` now genuinely throws a typed error, so the served text is wrong about both entry points instead of one.

That needs a Sapiom-repo PR against the rules constants + a content-release bump, then `node scripts/authoring-rules-stamp.mjs --from-served` here to re-stamp. I haven't filed it — flagging first because it's cross-repo and someone may already own the next content release. Happy to take it.

### Verification on the merged tree

`pnpm install --frozen-lockfile`, `build`, `typecheck`, `lint` all exit 0; terminology (997 files) and provider-copy (139 files) pass. Specs one file at a time, single worker: `agents/index` 32, `stub/agents-launch-keys` 10, `agents/schedule-launch` 3, `stub/models-launch-keys` 7, `agent-core/authoring-rules-stamp` 44.

One process note: I merged rather than rebased. The repo's `git-workflow` skill says "Never force-push a feature branch," and a rebase of a pushed branch requires one — it would also orphan the CodeRabbit inline anchors and David's review, which are pinned to the old SHAs. Say the word if you'd rather have the linear history and I'll rebase and force-push.

gwitwer:
@xoxo-david how does it look?

gwitwer:
Refreshed against main (85d652b4, merge of c71bc506; auto-merged `packages/tools/src/index.ts` and `stub/index.ts`, no conflicts). Locally on the merged head: `@sapiom/tools` typecheck, lint, 751 tests; `@sapiom/agent-core` 240 tests. @xoxo-david your design ask (typed throw from `launch`, `run` resolves `rejected`) is implemented in 35098364 and nothing else is outstanding; the backend now serves content release 1.1 that documents this contract (sapiom/Sapiom#5157), so the SDK is behind the published rules until this lands. Re-requesting review.

PR #867, opened 2026-09-07
fix(auth): order credential transitions

ynadge:
## Summary

- serialize browser sign-in credential commits with disconnect
- prevent an older refresh or sign-in completion from restoring a disconnected connection
- publish a monotonic connection generation for later launch and session coordination

Tracks [SAP-3116](https://linear.app/sapiom/issue/SAP-3116). This is the base PR in the stack; #868 follows it.

## Release note

- patch changeset for `@sapiom/harness`

## Testing

- focused API-key provider and auth-route suites: 36 tests passed
- targeted Prettier checks
- `git diff --check`

Refs: SAP-3116


ratataque (dismissed):
LGTM 👍

gwitwer (approved):
Approving. Read the queue and epoch logic: only write → refresh → publish is serialised, browser OAuth stays outside it, so a never-completing tab cannot block disconnect and a second `/auth/start` after disconnect gets 200. The race-window tests are deterministic (deferreds and invariants asserted inside the `clearCredentials` mock). Note for the record: ratataque's earlier LGTM was auto-dismissed by the merge-from-main push, not retracted, so it needs re-stamping. The abandoned-OAuth-completion key that gets discarded without a log or revoke is worth a follow-up ticket.

PR #868, opened 2026-09-08
fix(auth): fence MCP process launches

ynadge:
## Summary

- keep MCP launch preparation and process admission on the same connection generation
- derive `current` and `restart-required` state for supported live sessions
- make Canvas watcher startup idempotent for the same project root
- return a stable retryable conflict from session and background-task HTTP paths when the connection changes during launch

Tracks [SAP-3116](https://linear.app/sapiom/issue/SAP-3116). Stacked on #867; #869 follows it.

## Release note

- minor changeset for `@sapiom/harness`

## Testing

- focused provider, SessionManager, TaskManager, auth/MCP, and REST suites: 346 tests passed
- macro route suite: 25 tests passed
- `git diff --check`

Refs: SAP-3116


ratataque (dismissed):
--I'll keep restart manual, but as a follow up can we block use of the old Sapiom credentials after disconnect or account switch, without restarting the whole agent session?--

adressed in #872

PR #869, opened 2026-09-08
fix(auth): reconcile credential removal

ynadge:
## Summary

- reconcile affected Claude sessions and background tasks when Studio signs out
- apply the same reconciliation when logout happens through the shared CLI credential store
- report signed-out state immediately while process cleanup finishes
- re-arm credential-store observation after a first sign-in creates the shared directory
- fence queued cleanup to the connection generation that requested it

Tracks [SAP-3116](https://linear.app/sapiom/issue/SAP-3116). Stacked on #868; #872 follows it.

## Release note

- minor changeset for `@sapiom/harness`
- minor changeset for `@sapiom/mcp`

## Testing

- focused observer, auth, SessionManager, TaskManager, and REST suites: 357 tests passed
- state-root isolation regression suite: 3 tests passed
- `pnpm --filter @sapiom/mcp typecheck`
- `pnpm --filter @sapiom/mcp lint`
- `git diff --check`

Refs: SAP-3116


gwitwer (changes requested):
Requesting changes on this one, and the same point applies to #868 since it shares the root cause.

**Codex sessions receive the Sapiom key but sit outside every mechanism in this stack.**

- The credential stamp is only emitted when `req.harness === "claude-code"` (`packages/harness/src/server/index.ts:666-673`).
- But `generateMcpConfig` writes the `x-api-key` header for every harness, and the Codex adapter binds it into the process environment as `SAPIOM_CODEX_MCP_<i>_HEADER_<j>` (`core/adapters/codex-mcp.ts:113-128`, consumed at `core/adapters/codex.ts:331` and `:341`).
- `terminateCredentialBearingSessions` selects only stamped handles (`core/session-manager.ts:904-917`), so a live Codex session keeps calling `api.sapiom.ai/v1/mcp` with the removed key after disconnect or external logout. Codex launches are not fenced in #868 either, and #872's restart path skips them.

Codex is one of the two spawnable harnesses (`shared/types.ts:148`), so this contradicts the SAP-3116 AC "disconnect leaves no active, newly created, or resumed remote MCP session authorized with the removed key". A unit test pins the exclusion as intended, and ratataque asked about it on #872 without a reply. Either of these would resolve it:

1. Stamp Codex launches too. The env-var path already carries the key, so the same `mcpCredentialLaunch` stamp, fence, and sweep apply without adapter changes.
2. State explicitly in the PR body and the changeset that Codex is out of scope, with a follow-up ticket, so the release note does not overclaim.

Two smaller things while you're in here:

- #872 merged into this branch, not into main (`cf3e8723` is an ancestor of this branch only), so this PR now carries its restart API, banner, and changeset. The body still says "#872 follows it". Worth correcting so the approver knows what they're stamping.
- Test CI has not run on #868 or #869 yet: `test.yml` and `harness.yml` only trigger on PRs based on `main`, so both show `claude-review` only. They'll run after #867 merges and GitHub retargets. Just flagging that the test counts in the bodies are author-run so far.

The rest holds up well: no deadlock path (OAuth stays outside the queue, the sweep is bounded at ~2.5s), no off-by-one in the `<= throughGeneration` fence, and the race-window tests are deterministic. Happy to approve once the Codex question has an answer either way.

PR #885, opened 2026-09-09
fix(tools): fail loudly when a structured call truncates before its tool call [SAP-3280]

gwitwer:
## Primary change type

- [x] Bug fix

## Problem and motivation

Thinking tokens are spent out of `request.max_tokens`. A routed label emits a `thinking` block before the forced tool call, so a cap sized for the answer alone can be exhausted mid-deliberation: the turn ends before the tool call is emitted, `structuredOf` correctly returns `undefined`, and the author sees a `TypeError` from destructuring it — indistinguishable from a genuinely empty result. Deliberation is longest on the hardest, most ambiguous inputs, so an under-sized cap passes every test and every easy case, then drops exactly the item that was worth the most. Reported from the field: an under-sized cap copied out of our examples silently dropped structured results in production, repeatedly, and lost real work.

The examples caused it. The authoring skill's structured-output example said `256`; the `llm` JSDoc examples said `512`. The same pages warned that a `thinking` block precedes the output, but only as a parsing caution — never as a budgeting rule — so anyone copying the blessed shape inherited a latent truncation bug.

## Summary and scope

- `llm.run` throws the new `LlmStructuredOutputTruncatedError` when `output` was set and the turn stopped on `max_tokens` without a usable result — no matching `tool_use` block (`reason: "no-tool-call"`), or a block whose `input` is missing a schema-required field because the cap landed mid-input (`reason: "incomplete-input"`). The error names the tool, the cap, and the fix, and carries the verbatim response. Exported from `@sapiom/tools`.
- Everything stays gated on that stop reason, so nothing else changes: a truncated plain-text reply is still returned (it is readable, and bounding a reply deliberately is legitimate), a partial or empty structured result from a turn that ended any other way still reads as before, and an optional field the model omits is not read as truncation.
- Every example cap is now `4096`, and the skill and the JSDoc state beside the example that thinking counts against the cap and that the cap is a ceiling rather than a reservation — billing settles on the tokens actually produced, so headroom on a short reply costs nothing.
- The skill-sync test gains two content guards (the sentence must be present; no example cap under 2048), and `pnpm examples:check` now rejects a structured `llm.run` capped under 2048 across template sources and `examples/AUTHORING.md` — the snippet example authors copy, which was capped at 500. The one `index.ts` example with a structured cap in that range is raised too; plain-text examples with deliberately bounded replies are left alone, since truncating one is visible.

Out of scope: the MCP instructions and harness system prompt teach the parsing caution without showing a cap, and their text moves only through a content release.

## Related work

Related issue or discussion: SAP-3280. Monorepo companion (the docs guide, the two monorepo skills, the gateway SDK guide, and a cross-surface CI guard): sapiom/Sapiom#5010.

## Validation

```text
npx jest --maxWorkers=1 src/llm/structured-output.spec.ts   — 23 passed (9 new)
npx jest --maxWorkers=1 src/llm/submit.spec.ts              — 20 passed
npx jest --maxWorkers=1 src/llm/sessions.spec.ts            — 16 passed
npx jest --maxWorkers=1 src/stub/llm-structured-output.test.ts — 9 passed
npx jest --maxWorkers=1 packages/agent-core src/__tests__/skill-sync.test.ts — 17 passed (2 new)
pnpm --filter @sapiom/tools typecheck                       — clean
pnpm --filter @sapiom/tools lint                            — clean
node scripts/examples-check.mjs                             — OK, 12 templates
node --test scripts/examples-llm-surface.test.mjs           — 19 passed (4 new)
```

### Tests and documentation

Nine new tests cover both truncation shapes, the message and fields, a missing cap, an omitted optional field, the tool call that did arrive, an untruncated partial and empty result, and a truncated plain-text call. Four cover the new examples-check guard. Two content guards on the canonical skill. Docs updated in the skill (and its three shipped copies), the `llm` JSDoc, and `examples/AUTHORING.md`.

- Breaking or externally visible changes: `llm.run` now throws where it previously returned a response with no usable structured result. A consumer that handled the empty result itself should wrap the call and catch `LlmStructuredOutputTruncatedError` — `error.response` is the verbatim response it used to receive. Called out in the changeset.
- Changeset: added (`@sapiom/tools` minor, `@sapiom/agent-core` patch).

## Security

- [x] This pull request does not publicly disclose a suspected vulnerability.

## AI assistance

Written with Claude Code. It located every author-facing cap, wrote the error and its tests, and ran the commands listed under Validation.

## Checklist

- [x] I added or updated tests.
- [x] I ran the relevant build, typecheck, lint, and test commands.
- [x] I updated documentation for user-facing changes.
- [x] I added a Changeset for a published-package change.

https://claude.ai/code/session_01Pb1ZsUJTizT9nYpk1RtEuY


gwitwer:
All four addressed — three fixed in 8118eb7, one in the description.

**Confidentiality** — fair on a public repo. The description now reads "reported from the field: an under-sized cap copied out of our examples silently dropped structured results in production, repeatedly, and lost real work." No count, no delivery, no role beyond "the field". The technical argument didn't need the anecdote.

**`examples/AUTHORING.md` still taught the bug** — correct, and it is the worst place to have missed it, since it is the snippet an example author copies. Raised to 4096 with the same one-line comment and a paragraph stating the rule. You were also right that the skill-sync guard couldn't see it: `pnpm examples:check` now rejects a structured `llm.run` capped under 2048, across template sources and that doc, via a new `checkStructuredOutputCap` in `scripts/lib/examples-llm-surface.mjs` with four unit tests. It is scoped to the call it reads, not the file — a `textOf` reply capped at 700 is a length choice and truncating it is visible, so only calls declaring `output` are judged.

**A tool call cut off mid-input** — real, and the same silent failure one token later. `run` now throws for that too, with `reason` distinguishing `"no-tool-call"` from `"incomplete-input"`. Two things keep it from becoming a schema validator: it stays gated on `stop_reason === "max_tokens"`, so a partial result from a turn that ended any other way is still the caller's to judge, and only fields the schema lists in `required` count as evidence — an omitted optional field is the model's choice. Both are pinned by tests, alongside the empty-input case you described.

**Changeset breaking callout** — added, in your words, plus `error.response` as the migration path. It also now documents both `reason` values.

gwitwer:
All three addressed in 492a85b.

**The guard only saw a literal cap** — right, and the bypass was one line away: `fan-out-and-combine` already caps its calls with named constants, so `const PLAN_MAX_TOKENS = 512;` beside a structured call would have sailed through. It now resolves in-file `const NAME = <number>` declarations and reports the resolved value. A cap imported from another module or computed at runtime is still invisible; I left that rather than erroring on every non-literal, since a request built by spread would then fail for no reason. The edge is documented on the function and pinned by a test, so it stays a decision instead of a surprise.

**Stale class summary** — corrected. It now names both shapes, matching the `reason` field's own doc, since that summary is the hover text a consumer actually reads.

**Nits, both taken.** The union is exported as `LlmTruncationReason`. And you were right that `incomplete-input` could never fire for a schema with no `required` list — an empty `input` at the cap now counts on its own, which is the only evidence available in that case. Test added; the optional-field and non-truncated-turn cases still pass untouched, so it hasn't become a schema validator.

24 structured-output tests, 22 examples-check tests, typecheck and lint clean.

packages/tools/src/llm/index.ts
```diff
@@ -51,8 +51,17 @@
...
- *   const reply = await ctx.sapiom.llm.callSession(s, { max_tokens: 512, messages: […] });
+ *   const reply = await ctx.sapiom.llm.callSession(s, { max_tokens: 4096, messages: […] });
  *   await ctx.sapiom.llm.releaseSession(s);             // or let TTL/budget end it
+ *
+ * `max_tokens` on every one of these covers THINKING as well as output. A routed
+ * label may emit a `thinking` block before its answer, and those tokens are spent
+ * out of the same budget — so a cap sized for the answer alone can be exhausted
+ * mid-deliberation, ending the turn before the answer (or, with `output`, before
+ * the forced tool call) is ever emitted. Size it for thinking plus output: a few
+ * thousand tokens, not a few hundred. It is a ceiling, not a reservation — billing
```
antoine-berger: Billing does settle on produced tokens, but the cap is not free on the capacity side: platform admission weighting scales with `max_tokens`, so a large cap on a short call still costs headroom under load. Suggest keeping "billing settles on the tokens actually produced" and dropping "not a reservation" / "costs nothing", here and where the sentence is repeated: the error message (l. 501-502), the CI guard message, the skill (4 copies), `examples/AUTHORING.md` and the changeset.

packages/tools/src/llm/index.ts
```diff
@@ -431,6 +451,109 @@ export function structuredOf<TSchema = unknown>(response: unknown, name?: string
...
+ * counts on its own because a schema listing no `required` fields would otherwise have no
+ * evidence at all. Fields the schema does not require are the model's to omit and, alongside
+ * others that did arrive, are not read as truncation.
+ */
+function truncationOf(
+  response: unknown,
+  output: LlmStructuredOutputSpec,
+): LlmTruncationReason | undefined {
+  const stopReason = (response as { stop_reason?: unknown } | null | undefined)?.stop_reason;
+  if (stopReason !== "max_tokens") return undefined;
```
antoine-berger: Question on scope. Under `tool_choice: { type: "tool", name }` an honored request cannot legitimately end on `end_turn` with no matching `tool_use` block, so `structuredOf(...) === undefined` is a failed structured call whatever the stop reason. The gateway has real producers of that shape: `xml_tool_call_sanitizer.py:645` drops a text-embedded tool call whose name is not in `tool_names`, leaving a text-only `end_turn` reply, and `litellm_config.yaml:32` sets `drop_params: true`, which lets LiteLLM strip `tool_choice` for a deployment that rejects it. Both still return the body here, and the step hits the same destructuring `TypeError` this PR is titled to end.

`structured-output.spec.ts:330` pins the current behavior as intended, so this is a deliberate line; is it the one you want? The partial-input judgement clearly needs the `max_tokens` gate (otherwise this becomes a schema validator). The "no matching block at all" case does not: it could throw `no-tool-call` (or a third reason with its own message) regardless of stop reason.

packages/tools/src/llm/index.ts
```diff
@@ -475,11 +602,23 @@ export async function run<T = Record<string, unknown>>(
...
+  const response = await transport.request<T>(`${baseUrl}/v2/anthropic/v1/messages`, {
     method: "POST",
     body: JSON.stringify(request),
     headers,
   });
+  // Only when the caller forced a tool call: a plain-text turn that hit the cap is
+  // truncated but still readable, and it is the caller's to judge.
+  const truncation = spec.output ? truncationOf(response, spec.output) : undefined;
+  if (spec.output && truncation !== undefined) {
+    throw new LlmStructuredOutputTruncatedError(
```
antoine-berger, line 625: When this escapes a step, the runtime does not treat it as non-retryable (`packages/agent/src/non-retryable-step-error.ts` is a closed set), so the step is re-run with the identical under-capped request up to `DEFAULT_MAX_ATTEMPTS_PER_STEP = 3` times, billing the thinking each time, and cannot succeed. Worth saying in the `@throws` and in the skill: catch it and `fail()` under `canFail: true`, the shape the skill already teaches for `CodingRunHttpError`.

packages/tools/src/llm/structured-output.spec.ts
```diff
@@ -189,3 +189,164 @@ describe("ctx.sapiom.llm.{structuredOf,textOf,readDisclosure} — reachable from
...
+    expect(error.message).toContain("max_tokens (256)");
+    expect(error.message).toContain("Thinking tokens count against max_tokens");
+  });
+
+  it("throws when the cap landed mid-input, leaving a required field unwritten", async () => {
+    // The same failure one token later: the block is there, so a presence check reads it as
+    // success, and the caller destructures `undefined` out of a partial object.
+    const cutMidInput = {
+      stop_reason: "max_tokens",
+      content: [{ type: "tool_use", name: "classify_ticket", input: {} }],
```
antoine-berger: This fixture is `input: {}`, which returns at the empty-object check (l. 551) before the `required` check on l. 553 runs, and no other `max_tokens` fixture has a non-empty partial input, so that branch is never exercised (`const missing = false` on l. 553 keeps all tests green). Something like `input: { note: "x" }` against the schema requiring `priority` covers what the title describes.

packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
```diff
@@ -394,6 +395,18 @@ the result back out. For a **plain-text** reply instead, use
 `ctx.sapiom.llm.textOf(response)` — it reads only the `type === 'text'` block, skipping a
 `thinking` block that may precede it.
 
+**Thinking tokens count against `max_tokens`.** The same `thinking` block you must not parse
+is also spent out of the cap. If deliberation exhausts the budget, the turn ends before the
+forced tool call is ever emitted: the response carries no `tool_use` block, `structuredOf`
```
antoine-berger: This still describes the pre-PR behavior: on `max_tokens`, `llm.run` now throws `LlmStructuredOutputTruncatedError` before `structuredOf` is reached. The error is named in `examples/AUTHORING.md` and the changeset, but nowhere in the skill or its three synced copies. Name it here and say what to do with it (catch + `fail()` under `canFail`, or let it surface).

scripts/lib/examples-llm-surface.mjs
```diff
@@ -97,6 +97,87 @@ export function checkNoSliceParse({ path, source }) {
...
+
+  for (let i = 0; i < lines.length; i += 1) {
+    if (!LLM_RUN_OPEN.test(lines[i])) continue;
+
+    let capLine = -1;
+    let declaresOutput = false;
+    for (let j = i; j < lines.length; j += 1) {
+      if (/\boutput\s*:/.test(lines[j])) declaresOutput = true;
+      if (capLine === -1 && CAP_ASSIGNMENT.test(lines[j])) capLine = j;
+      if (j > i && CALL_CLOSE.test(lines[j])) break;
```
antoine-berger: The line-based window makes this guard both over- and under-report. Checked by running the shipped function against a few shapes:

- `CALL_CLOSE` only matches a bare `});`, so a plain-text call closed with `}),` (arrow-wrapped, e.g. inside `withRetry(() => ...)` or `Promise.all([...])`, which is what prettier emits) never terminates its window and absorbs the next call's `output:`. A legitimate `max_tokens: 700` text call is then reported as a starved structured call: false CI failure.
- `/\boutput\s*:/` runs on raw lines, so `system: "... Expected output: one paragraph."` or a `// output: ...` comment inside a text call's window marks it structured (same false failure). `examples/eval-gate/index.ts:227` already has an `output: input.draft,` property nearby, so the collision is realistic. Conversely the first `max_tokens:` hit wins, even from a comment.
- `LLM_RUN_OPEN` is the literal `llm.run({`, so `ctx.sapiom.llm.run<Review>({` (legal: `client.ts:216` is generic, and generics are already taught for `structuredOf<...>`) or prettier's chain break `llm\n  .run({` skip the guard entirely, including at `max_tokens: 256`.

The current templates happen to pass, so this is latent until an author wraps a call, and none of the new tests exercise a `}),`, same-line or generic shape. A brace-balanced slice from the `(` fixes the window in both directions; `scripts/examples-entry-schema-check.mjs:88` already has `sliceBalanced(source, openIndex)` doing exactly that.

antoine-berger (changes requested):

packages/tools/src/llm/index.ts
gwitwer: Right, and I checked before changing it: `llm-gateway/routing/weight_policy.py` estimates admission weight as `1 + input_tokens/tokens_per_unit + max_tokens/tokens_per_unit`, clamped to `[1, max_weight]`, with `max_tokens` explicitly the duration-risk proxy. At the default `tokens_per_unit: 4000`, moving an example from 256 to 4096 adds a whole unit of weight. So "costs nothing" was wrong on exactly the axis that matters under load, and it was wrong in seven places.

Dropped as you suggested, everywhere: the JSDoc header, the error message, the CI guard message, the skill and its three synced copies, `examples/AUTHORING.md` and the changeset. The billing half stays, and rather than leaving a gap where the false reassurance was, the surfaces now say the cap is not free either — admission weight scales with it, so size it for the work rather than padding it. That reads as a reason to pick a real number instead of a reason to pad, which is what the ticket wanted.

packages/tools/src/llm/index.ts
gwitwer: Confirmed — `NonRetryableStepErrorPayload` is a closed union of three platform payloads and this error is not one of them, so a step that lets it escape is re-run with the identical under-capped request up to `DEFAULT_MAX_ATTEMPTS_PER_STEP = 3`, billing the thinking each time and failing the same way. A deterministic failure retried three times is worse than the silent one it replaced.

The `@throws` now says so and points at `fail()` under `canFail: true`, and the skill (plus its three copies) carries the handler in the `CodingRunHttpError` shape it already teaches, with `error.reason` noted as informational since both values mean the same fix.

I did not add the error to the non-retryable set. It looks like the more correct fix, but that union is a platform-owned taxonomy with a parse-and-normalize boundary, and widening it changes retry semantics for a class of error rather than documenting one — a bigger decision than this ticket should make on its own. Happy to open a follow-up if you want it there.

packages/tools/src/llm/structured-output.spec.ts
gwitwer: Caught a genuinely dead assertion — `input: {}` returns at the empty-object check before the `required` comparison, so the branch the test is named for never ran. The fixture is now `input: { note: "partial" }` against the schema requiring `priority`, and the empty-input case keeps its own test.

Verified rather than assumed this time: stubbing `missing` to `false` fails that one test and nothing else, so the branch is really covered now.

packages/agent-core/skills/sapiom-agent-authoring/SKILL.md
gwitwer: Correct, and it was my own paragraph left stale by my own change — `llm.run` throws before `structuredOf` is ever reached, so the text described a path that no longer exists.

Rewritten in the canonical skill and synced to all three copies: it names `LlmStructuredOutputTruncatedError`, says the turn ends before the tool call is emitted, and adds a worked handler (catch, `fail(error.message)` under `canFail: true`) directly modelled on the `CodingRunHttpError` block above it, so the two read the same way.

antoine-berger (dismissed):

gwitwer:
Refreshed against main (75962f6d, merge of c71bc506; no conflicts). Locally on the merged head: `@sapiom/tools` typecheck, lint, 728 tests; `@sapiom/agent-core` 242 tests; terminology and examples checks pass. @antoine-berger your approval on b1db904c was auto-dismissed by the earlier merge-from-main push; all four of your asks were addressed in b49202f0 and nothing is outstanding. Re-requesting review.

packages/tools/src/index.ts
```diff
@@ -113,6 +113,11 @@ export {
 // raw wire shape (`LlmDisclosure`) stays on the `llm` namespace / subpath.
 export type { RoutingLabel, LlmDisclosureResult } from "./llm/index.js";
 export { readDisclosure } from "./llm/index.js";
+// Thrown by `llm.run` when a forced structured-output tool call never arrived
+// because the turn hit `max_tokens` first — top-level so a step can catch it by
+// name without reaching into the `llm` namespace.
+export { LlmStructuredOutputTruncatedError } from "./llm/index.js";
```
antoine-berger: **[medium] Root export of the new error is not pinned by the barrel smoke test**

`packages/tools/src/smoke.spec.ts:147-158` pins every other root-exported error class, including both llm lanes' errors.

| | |
|---|---|
| Today | This line is the documented way a step catches the failure by name, and the changeset tells consumers to `catch` it that way |
| Missing | The smoke spec asserts `SearchHttpError`, `KeysHttpError`, `CodingRunHttpError`, `LlmRouteResultSchemaError` and `LlmSessionReadySchemaError`, not this one |
| Effect | Dropping the barrel line in a later refactor breaks every consumer's catch and no test in this repo goes red |
| Fix | Add `expect(typeof LlmStructuredOutputTruncatedError).toBe("function")` to the barrel surface test alongside the sibling error classes |

packages/tools/src/llm/index.ts
```diff
@@ -11,7 +11,7 @@
  * caller's Sapiom API key (identity mode — no x402 payment handshake needed).
  *
  *   const reply = await ctx.sapiom.llm.run({
- *     request: { messages: [{ role: "user", content: "…" }], max_tokens: 512 },
+ *     request: { messages: [{ role: "user", content: "…" }], max_tokens: 4096 },
```
antoine-berger: **[medium] The JSDoc caps the ticket blames are raised with nothing guarding them**

`scripts/examples-check.mjs:218-239` runs `checkStructuredOutputCap` over `examples/` sources and `examples/AUTHORING.md` only.

| | |
|---|---|
| Today | Lines 14, 28 and 54 move from 512 to 4096, the numbers the reporter says they copied for a classification step |
| Missing | The skill has a content guard and `examples/` has a CI floor; the `llm` JSDoc, the other surface SAP-3280 names, has neither |
| Effect | A later edit can put 512 back into the SDK doc comments and every check added by this PR still passes |
| Fix | Pin these caps in a unit test, or widen the guard to judge doc-comment caps, which it cannot do today since it only judges calls declaring `output` |

scripts/lib/examples-llm-surface.mjs
```diff
@@ -97,6 +97,87 @@ export function checkNoSliceParse({ path, source }) {
...
+  const consts = numericConstsOf(source);
+
+  for (let i = 0; i < lines.length; i += 1) {
+    if (!LLM_RUN_OPEN.test(lines[i])) continue;
+
+    let capLine = -1;
+    let declaresOutput = false;
+    for (let j = i; j < lines.length; j += 1) {
+      if (/\boutput\s*:/.test(lines[j])) declaresOutput = true;
+      if (capLine === -1 && CAP_ASSIGNMENT.test(lines[j])) capLine = j;
```
antoine-berger: **[low] A single-line llm.run never closes its scan and adopts the next call's output**

The `j > i` guard skips `CALL_CLOSE` on the opening line, and `CALL_CLOSE` needs `});` at line start, so a one-liner scans on to the next call's closing line.

| | |
|---|---|
| Trigger | A one-line text call such as `const res = await ctx.sapiom.llm.run({ request });` followed anywhere below by a structured call in the same file |
| Effect | `declaresOutput` picks up the later call's `output:` and the text call is reported as a starved structured cap, failing `examples:check` at the wrong line |
| Fix | Track brace depth from the opening line, or stop the scan when `lines[i]` already contains the call's own closing `})` |

antoine-berger (dismissed):

gwitwer:
Addressed the six open bot findings in two commits.

| Finding | Where | Commit |
| --- | --- | --- |
| Devin: nested truncation escapes detection | `packages/tools/src/llm/index.ts` ~557 | 099e892f |
| CodeRabbit (Major): detect required fields in nested objects | `packages/tools/src/llm/index.ts` ~557 | 099e892f |
| Devin: valid empty result falsely rejected | `packages/tools/src/llm/index.ts` ~554 | 099e892f |
| Devin: token-cap guard has narrow syntax coverage | `scripts/lib/examples-llm-surface.mjs` ~111 | 900295e6 |
| CodeRabbit (Major): support optional type arguments in the scanner | `scripts/lib/examples-llm-surface.mjs` ~111 | 900295e6 |
| CodeRabbit (Minor): resolve named constants before the 2048 floor | `packages/agent-core/src/__tests__/skill-sync.test.ts` ~73 | 900295e6 |

**099e892f** — `truncationOf` now walks the full `output.schema` when `stop_reason === "max_tokens"`: `properties`/`required` through nested objects, `items`/`prefixItems` for arrays, `allOf` as every branch, `anyOf`/`oneOf` as at least one. Presence-only, no type checking, and unchanged under any other stop reason. The first missing required field is exposed as `error.missingPath` (`result.priority`, `items[1].score`) and quoted in the message. The unconditional `{}` ⇒ incomplete rule is removed; `{}` throws only when the schema requires something. Tests cover nested object, array element, `anyOf`/`allOf`, `{}` both ways, and a complete nested input at the cap. Changeset updated for the message shape.

**900295e6** — the scanner finds a call by `llm.run`, optional `<...>`, optional whitespace, `(`, and reads to the matching close paren, so generic, spaced, next-line-spec and one-line `llm.run(spec)` forms are each one call. The skill-sync floor guard resolves `max_tokens: CAP` against an in-file `const CAP = N;` (mirrored rather than imported: that Jest suite is CommonJS and cannot load the `.mjs`). Regression tests added for each form.

Aliased calls (`const ask = ctx.sapiom.llm.run`) and imported/computed caps remain the documented known edge of the scanner.

Verification: `@sapiom/tools` 735 tests, typecheck, lint green; `@sapiom/agent-core` 242 tests green; `scripts/examples-llm-surface.test.mjs` 27 pass; `examples-check` and `terminology:check` pass.

packages/tools/src/llm/index.ts
```diff
@@ -463,6 +677,17 @@ function withStructuredOutput(request: Record<string, unknown>, output: LlmStruc
...
+ * @throws {LlmStructuredOutputTruncatedError} when `spec.output` was set and the
+ * turn hit `max_tokens` before the forced tool call was emitted, or partway through
+ * its input — the cap has to cover thinking as well as output.
+ *
+ * This one is not worth retrying, and the engine does not know that: it is not a member
+ * of the platform's non-retryable set, so left to escape a step it is re-run with the
+ * identical under-capped request (up to the runner's per-step attempt limit), billing the
+ * thinking each time and failing the same way. Catch it in a step declared `canFail: true`
+ * and return `fail(error.message)` — the shape `CodingRunHttpError` already uses. Raising
+ * the cap is the actual fix.
```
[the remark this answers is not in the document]
gwitwer: Known and deliberate. antoine-berger raised this on Sep 9 (https://github.com/sapiom/sapiom-js/pull/885#discussion_r3964241374): `NonRetryableStepErrorPayload` is a closed platform set, so this PR documents the behavior instead (the `@throws` here and the authoring skill tell authors to catch it and `fail()` under `canFail: true`) and defers admitting the error to that set to a follow-up.

gwitwer:
Third bot round, addressed in d1790df6 (and one reply).

| Finding | Where | Resolution |
| --- | --- | --- |
| Devin: referenced requirements remain silently truncated (`$ref`) | `packages/tools/src/llm/index.ts` ~589 | d1790df6 |
| Devin + CodeRabbit (Major): prompt text bypasses cap checks (`callEndOf`) | `scripts/lib/examples-llm-surface.mjs` ~170 | d1790df6 |
| Devin: truncation errors still trigger retries | `packages/tools/src/llm/index.ts` ~690 | No code change; replied on the thread pointing at the Sep 9 discussion and the deferral to a follow-up |

**$ref** — the walker resolves local references (`#`, `#/$defs/x`, `#/definitions/x`, any JSON Pointer under the root) against the root schema and walks the target at the same position alongside sibling keywords. A per-position visited set stops a reference cycle that consumes no input. Remote/external refs are left unresolved and read as declaring no requirements, so nothing reaches the network. Tests: nested required behind `$defs` (`result.priority`) and `definitions` inside array items (`items[1].name`), an `a → b → a` cycle that terminates and still reports `child.child.value`, and an external URL ref that is ignored.

**callEndOf** — the scan is now lexically aware: it skips `'…'` / `"…"` strings (with escapes), template literals including nested `${…}` expressions and their own strings, and `//` / `/* */` comments before counting parentheses. Tests: prompts containing `)`, `(` and an escaped quote; a template literal with `${fn(x)}` and a stringified `")"`; line and block comments with parens; and a one-line call containing a quoted `)` that must still close on its own line. A regex literal with a parenthesis is the documented remaining edge.

Verification: `@sapiom/tools` 738 tests, typecheck, lint green; `@sapiom/agent-core` 242 tests green; `scripts/examples-llm-surface.test.mjs` 31 pass; `examples-check` and `terminology:check` pass.

antoine-berger (approved):

PR #974, opened 2026-09-14
feat(harness): durably store accepted assistant sources [SAP-3328]

ynadge:
## Primary change type

- [x] Feature

## Problem and motivation

Assistant recovery must reload the exact accepted context after a host restart. Source artifacts need durable, immutable publication before a request can use them.

## Summary and scope

Add a private scoped repository outside native engine state. Validate and snapshot all inputs, sync source objects, then publish the accepted manifest last with no replacement. Readback verifies authority, conversation, record digest, every available source and safe filesystem paths. Concurrent identical writes are idempotent; conflicting identities, corruption and incomplete content fail closed. No retention acknowledgement follows a failed file or directory sync. Shutdown retains the repository.

## Related work

Related issue or discussion: [SAP-3328](https://linear.app/sapiom/issue/SAP-3328). Builds on [#973](https://github.com/sapiom/sapiom-js/pull/973). The delivery coordinator will consume this repository in a later stack increment.

## Validation

```text
pnpm --filter @sapiom/harness exec vitest run src/core/assistant-source-store.test.ts src/core/assistant-sources.test.ts — 52 passed after source review fixes
pnpm --filter @sapiom/harness typecheck — passed
pnpm --filter @sapiom/harness lint — passed
pnpm --filter @sapiom/harness build:server — passed
pnpm pr:size — unavailable in SDK; actual base increment: 713 additions, 0 deletions (713 total)
 git diff --check — passed
```

### Tests and documentation

Tests cover restart, exact text/package resources, detached inputs/results, concurrent writers, identity conflicts, foreign references, missing/corrupt/symlink objects, directory substitution, cancellation and file/directory sync failures. Removing publication-directory fsync makes both dedicated regression cases fail. Documents manifest-last acceptance and interrupted-write semantics.

## Compatibility and release impact

- Breaking or externally visible changes: Additive internal storage; request activation follows. Directory fsync is mandatory and unsupported platforms fail closed. These filesystem tests verify Linux; they do not establish Windows/macOS durability.
- Changeset: Added for `@sapiom/harness`.

Owner: Yash. Existing internal Assistant access remains the rollout boundary. Fix-forward: correct storage in a focused PR while preserving immutable objects and accepted-record identity; never recover from fresh providers when retained content fails verification.

## Security

- [x] I have not included secrets, credentials, private data, or unsanitized logs.
[...]

PR #976, opened 2026-09-14
feat(opencode): bind native context projection to accepted requests [SAP-3328]

ynadge:
## Primary change type

- [x] Feature

## Problem and motivation

Native prompt assembly needs to project the saved accepted context into stable guidance followed by completion metadata and dynamic facts. Title and compaction work can interleave with ordinary turns, so the latest session capture alone cannot identify the request.

## Summary and scope

Add a native hook factory bound to an authority scope and actual native session/user IDs. Preserve awaited completion restoration, including tool-loop histories ending with an assistant. Retire capture proofs at native idle/deletion, guard in-flight deletion, require ordinary/synthetic Studio context, and keep per-message validation failures sticky during active execution, allow verified historical title requests, and preserve unclaimed native helpers. Replace only the exact saved suffix in the existing system array; retain native prefix bytes and saved history. Use a fixed provider-boundary error that native persists without overload retries.

The factory requires the owned runtime's tested messageID/agent extension for claimed requests. This PR provides the consumer API; launcher/host activation and the corrected artifact follow in the stack. The no-scope wrapper keeps existing completion-only behavior.

## Related work

Related issue or discussion: [SAP-3328](https://linear.app/sapiom/issue/SAP-3328). Builds on [#974](https://github.com/sapiom/sapiom-js/pull/974). Native source probes established the title race and safe error boundary before implementation.

## Validation

```text
pnpm --filter @sapiom/opencode test — 95 passed before capture review fix, including six actual-native launcher tests
pnpm --filter @sapiom/opencode exec vitest run src/assistant-context-hook.test.ts — 23 passed on final revision
pnpm --filter @sapiom/opencode typecheck — passed
pnpm --filter @sapiom/opencode lint — passed
pnpm --filter @sapiom/opencode build — passed
pnpm pr:size — unavailable in SDK; actual base increment: 727 additions, 9 deletions (736 total)
 git diff --check — passed
```

### Tests and documentation

Twenty-three new tests exercise exact prefix/array preservation, marker-shaped guidance, same-folder sessions, recovery attempts, synthetic/tool continuation, title ordering and races, compaction history copies, deferred/sticky failures, missing identity/capture/suffix, and legacy behavior. Review regressions cover seventeen pending captures, repeated native retries, changed saved bytes under the same user ID, and session deletion during a history read. Capture and idle-retirement regressions failed before their fixes. Missing whole context and completion-only scoped work are rejected; the no-scope wrapper remains compatible. The full 104-test OpenCode suite passed with this consumer and downstream launcher/provider changes against the corrected binary. The synthetic tool-loop regression fails with the old last-array-item restoration and passes with the fix. Three raced-helper regressions fail before the fix and now reject foreign claimed context even when history is generic or completion-only. README records API and artifact requirements. Existing native tests verify wrapper compatibility; production consumer-on-binary tests follow with launcher wiring.

- Breaking or externally visible changes: Additive hook API. Context failures now use a fixed native-compatible message; the class and public Studio failure code remain unchanged. Accepted request production is not activated here.
- Changeset: Added for `@sapiom/opencode`.

Owner: Yash. Existing internal Assistant access remains the rollout boundary. Fix-forward: correct the request-bound projection in a focused PR; do not borrow another request's context or weaken malformed-envelope rejection.

## Security

- [x] This pull request does not publicly disclose a suspected vulnerability. I will follow the [Security Policy](https://github.com/sapiom/sapiom-js/security/policy) for private reporting.

## AI assistance

Codex implemented and reviewed the codecs and tests, then ran the checks above.

## Checklist


packages/opencode/src/assistant-context-hook.ts
```diff
@@ -0,0 +1,273 @@
...
+      // Only in-flight callbacks need a deletion guard; no permanent tombstone table.
+      if (deleted)
+        for (const guard of pending.get(id) ?? []) guard.deleted = true;
+    },
+    async "experimental.chat.messages.transform"(
+      input: Record<string, never>,
+      output: CompletionTransformOutput,
+    ): Promise<void> {
+      // Preserve the existing awaited synthetic-continuation restoration first.
+      await completion["experimental.chat.messages.transform"](input, output);
```
ratataque, line 189: P2: pending message hook recreates deleted-session state
[Comment on line 189](https://github.com/sapiom/sapiom-js/pull/976/files#diff-e4aadef419308374d1225ec11bd524257bc8b65ba988384421f9ccd3fc24075cR189).
Deletion guards cover system.transform, but not the awaited restoration
inside messages.transform.
Reproduced at 8acbdd4:
synthetic continuation starts history read
  -> session.deleted clears captures
  -> history read finishes
  -> capture() recreates deleted session’s epoch
  -> system.transform successfully projects stale context
Fix: guard pending message restoration across deletion;
prevent stale callbacks from publishing captures.
Keep failures at the provider boundary, as intended.
Add a regression deleting the session while
messages.transform waits for history.

PR #1055, opened 2026-09-18
refactor(agent-map): extract shared public contracts

ynadge:
## Primary change type

- [x] Maintenance or refactor

## Problem and motivation

Agent Map contracts currently live inside Studio, so standalone MCP consumers cannot reuse them without depending on Harness. This introduces `@sapiom/agent-map` and moves the existing contracts intact; Studio imports the package while preserving its public API and saved state.

## Summary and scope

- Export browser-safe map contracts and the existing project-ID predicate, with Node 18 support.
- Wire the workspace dependency, package metadata/license/linting, local publication order, and Harness CI path filters.
- Update Studio server, browser, tests, and e2e imports together.

PR 1 of 5 for shared-core extraction. Review base: `main`. Full diff: **389 changed lines** (268 additions + 121 deletions), including 509 existing source lines relocated with Git rename detection. MCP activation remains a later ticket.

## Related work

Related issue: [SAP-3521](https://linear.app/sapiom/issue/SAP-3521), under [SAP-3520](https://linear.app/sapiom/issue/SAP-3520).

## Validation

```text
pnpm --filter @sapiom/agent-map build — passed
pnpm --filter @sapiom/agent-map typecheck — passed
pnpm --filter @sapiom/agent-map lint — passed
pnpm --filter @sapiom/harness build — passed
pnpm --filter @sapiom/harness typecheck — passed
pnpm --filter @sapiom/harness exec vitest run agent-map agent-brief build-plan project-roots studio-project-catalog durable-file-lock canonical-graph-path paths.test workspace-scope --maxWorkers=2 — 503 passed
pnpm exec prettier --check packages/agent-map/package.json packages/agent-map/README.md packages/agent-map/.eslintrc.cjs — passed
git diff --check — passed
```

Local checks ran on Node 24. Full repository Node 20/22, browser, and packaged Desktop checks run in GitHub CI; the local sandbox has insufficient disk for the full Desktop toolchain.

GitHub CI at `d9d6b13` also passed the full repository Node 20/22 checks, both browser shards, Linux packaged Desktop smoke, and CodeQL.

### Tests and documentation

Existing Studio regression tests exercise the relocated contracts. The new README documents consumption and the first-publication requirement. No rendered UI behavior changes.

## Compatibility and release impact

- Breaking or externally visible changes: none to Studio behavior, current tools, or stored map data. New public package exports are additive.
- Changeset: new package minor and Harness patch.
[...]

PR #1059, opened 2026-09-18
refactor(agent-map): share project identity and resolution

ynadge:
## Primary change type

- [x] Maintenance or refactor

## Problem and motivation

A standalone repository must resolve the same project and workspace file that Studio uses, including custom state roots and multiple root bindings. This shares the neutral catalog/path primitives and adds read-only host/repository resolution with explicit missing, ambiguous, and unavailable results.

## Summary and scope

- Move the neutral catalog and canonical path cache; keep discovery/reconciliation orchestration in Studio.
- Share root matching and home expansion, preserving existing Studio wrapper behavior.
- Add `resolveAgentMapProject` for authenticated host scope or repository lookup with an optional project selector; lookup never registers a project or creates a map.
- Document private path handling and the explicit catalog registration boundary.
- Resolve cwd and catalog-root identity asynchronously. Cache catalog-root probes for at most one second per catalog instance while refreshing cwd on every lookup; retain the cached graph projection API.
- Isolate unreadable unrelated roots. Fail closed when cwd or a root containing its lexical/last-known canonical path is unreadable; preserve missing-descendant reconstruction.
- Propagate `storage_unavailable` through the Studio compatibility wrapper, preventing uncertain ownership from registering foreign agents or creating sessions/projects.
- Normalize dot segments before browser/server root containment and depth ranking; preserve Windows/UNC roots, deterministic ambiguity, and original root objects.
- Bound normalization of caller-supplied paths with linear separator scans; add two long-input regression cases.

PR 4 of 5; review against PR 3. Full diff: **970 changed lines** (857 additions + 113 deletions), including 1,123 existing source lines relocated with Git rename detection. Regression coverage for review findings takes this above the 600-line target and keeps it below the 1,000-line hard limit.

## Related work

Related issue: [SAP-3521](https://linear.app/sapiom/issue/SAP-3521), under [SAP-3520](https://linear.app/sapiom/issue/SAP-3520).

## Validation

```text
pnpm --filter @sapiom/agent-map build / typecheck / lint — passed
pnpm --filter @sapiom/harness build / typecheck / lint — passed
pnpm --filter @sapiom/agent-map test --maxWorkers=2 — 42 passed
Focused Studio map/brief/plan/storage/catalog/path/session/registration tests — 534 passed
Regression check against the previous implementation — 17 shared cases and 3 Studio cases fail as expected
git diff --check — passed
```

Local checks ran on Node 24; the sandbox has a partial dependency installation and limited disk space.

The current CodeQL/security-analysis, classification, and automated-review checks pass. Full Test/Harness workflows have not started for these commits; stack foundation [#1055](https://github.com/sapiom/sapiom-js/pull/1055) currently conflicts with `main`. Full CI must rerun after the stack is updated.

### Tests and documentation

Twenty-eight resolver cases cover host/repository parity, nested paths, custom roots, symlinks, worktrees, selectors, ambiguity, read-only lookup, and filesystem failures. The error matrix distinguishes cwd/ancestor/relevant-binding failures from unreadable unrelated roots. Cache tests verify repeated root probes are avoided, cwd stays fresh, no synchronous realpath runs during lookup, expired successes/failures are refreshed, and failed aliases retain their last-known canonical ownership. Twelve lexical matcher cases and two long-input path cases remain covered.

Studio regressions prove unavailable ownership aborts live/recovered agent registration and session creation before reconciliation or catalog/session writes. These checks fail against the previous implementation. README documents the one-second root cache, asynchronous refresh, failure isolation, and compatibility-wrapper errors.

The three requested fixes from Antoine's September 19 review are implemented and their threads are resolved. Human re-review remains required.

- Breaking or externally visible changes: existing Studio APIs, current tools, and stored map data stay compatible. Project matching now refreshes symlinks and normalizes lexical paths. New public package exports are additive.
- Changeset: shared package minor and Harness patch.

## Security

- [x] No secrets, credentials, private data, or unsanitized logs included.
- [x] No suspected vulnerability is publicly disclosed; the repository Security Policy applies.

## AI assistance

- [x] Codex prepared the extraction, package wiring, documentation, and test runs. The diff was reviewed against the original implementation, all relocated imports were inventoried, and the checks above verified the result.

## Checklist

- [x] Read `CONTRIBUTING.md`; implementation follows the agreed Linear plan.
- [x] Focused scope with no unrelated cleanup.
- [x] Existing regression coverage retained and exercised.
- [x] Relevant checks run; remaining environment/CI checks disclosed above.
- [x] Documentation and Changeset included.
- [x] AI-assisted changes reviewed and understood.

## Summary by CodeRabbit

- **New Features**
  - Added reliable project identity resolution for host projects and repository paths.
  - Added canonical path matching across POSIX, Windows, symlink, and worktree layouts.
  - Added workspace scope metadata and shared project catalog capabilities.
  - Added clearer outcomes for resolved, unregistered, ambiguous, and unavailable projects.
  - Studio now coordinates project discovery, registration, and session creation more consistently.

- **Bug Fixes**
  - Isolated filesystem failures so unrelated projects remain usable.
  - Improved recovery after filesystem changes and temporary storage failures.
  - Improved handling of home-directory paths and malformed project state.

- **Performance**
  - Improved path handling for deeply nested or separator-heavy paths.
  - Added brief caching for project root checks.


packages/agent-map/src/core/studio-project-catalog.ts
```diff
@@ -468,39 +477,47 @@ export class StudioProjectCatalog {
...
+      const match = matchProjectRootForPath(
+        canonical,
+        this.projects!
+          .filter((project) => !projectId || project.projectId === projectId)
+          .flatMap((project) =>
+            project.rootBindings
+              .filter(({ status }) => status === "active")
+              .map((binding) => ({
+                projectId: project.projectId,
+                cwd: refreshCanonicalGraphPath(binding.localRootRef),
```
antoine-berger: **[high] One unreadable root binding fails every project lookup, not just that project's**

Line 503 probes every active binding of every project inside a single try block; the base branch wrapped each binding in its own `catch { return []; }` and skipped only the failing one.

| | |
|---|---|
| Trigger | One active binding on an `EACCES`, `ELOOP` or stale-mount path, in any project, including a project that cannot contain `cwd`. |
| Before | `canonicalGraphPath` swallowed every filesystem error and fell back to the lexical path, so the lookup still resolved the correct project. |
| Effect | `lookupIdentityForPath` now returns `unavailable` for every cwd, so no project resolves at all until that unrelated root is readable again. |
| Fix | Probe each binding in its own try and fail closed only when the failing root could contain `cwd`, meaning an ancestor or an equal-depth candidate. |

packages/agent-map/src/core/studio-project-catalog.ts
```diff
@@ -468,39 +477,47 @@ export class StudioProjectCatalog {
    * Resolves a cwd to the most-specific active durable project root. Local
    * roots remain private; ambiguous equal-specificity matches fail closed.
    */
-  async resolveIdentityForPath(
+  async resolveIdentityForPath(cwd: string): Promise<ResolvedStudioProjectIdentity | null> {
+    const result = await this.lookupIdentityForPath(cwd);
+    return result.kind === "resolved" ? result.project : null;
```
antoine-berger, line 487: **[high] `unavailable` collapses to null here, and Studio callers read null as unowned**

`created-agent-registration.ts:119` and `:164` only skip when `owner` is truthy, and `server/index.ts:1597` reconciles and creates when the lookup returns null.

| | |
|---|---|
| Trigger | A non-missing filesystem error on any catalog root becomes `unavailable`, which this wrapper turns into the same `null` as unregistered. |
| Effect | `created-agent-registration.ts:119` sees `owner === null`, treats a foreign agent as unclaimed and registers it under the creating project. |
| Also | `server/index.ts:1597` then mints a second durable project for a cwd already inside a registered root and bootstraps a map for it. |
| Fix | Give these callers the `StudioProjectPathLookup` value, or a distinct sentinel, so `unavailable` aborts rather than reading as unowned. |

packages/agent-map/src/core/canonical-graph-path.ts
```diff
@@ -49,11 +53,27 @@ export function rememberCanonicalGraphPath(
...
+}
+
+/** Fresh scope lookup surfaces I/O failures; graph projection stays best effort. */
+function resolveCanonicalGraphPath(input: string, fresh: boolean): string {
   const windows = isWindowsAbsolute(input);
   const api = pathApi(input);
   const resolved = normalizedAbsolute(input);
   const cached = canonicalPaths.get(resolved);
-  if (cached !== undefined) {
+  if (!fresh && cached !== undefined) {
```
antoine-berger: **[medium] Fresh lookup bypasses the cache, so each lookup syscalls once per active binding**

`studio-project-catalog.ts:493` and `:503` both call `refreshCanonicalGraphPath`, which skips the cache and runs a synchronous `realpathSync.native` for the cwd and for every active binding.

| | |
|---|---|
| Cost | Every lookup performs `1 + N` blocking `realpathSync.native` calls, where `N` is every active binding in the whole catalog. |
| Volume | `created-agent-registration.ts:163` calls it once per created agent during restore, and `server/index.ts:1565` on every session spawn and resume. |
| Effect | The event loop stalls for the full duration; on a stale NFS or SMB root each call blocks for the mount timeout. |
| Fix | Refresh the cwd plus only the bindings that could contain it, or bound the refresh with a short time-based cache entry. |

antoine-berger (changes requested):

PR #1063, opened 2026-09-18
feat(harness): expose authenticated map host context

ynadge:
Expose `GET /mcp/agent-map/host-context` through the existing private map bearer registry. Studio verifies the active session, current principal and project binding, then returns only its trusted scope, custom state root and implemented context capability. Requests recheck token generation after the asynchronous lookup and never cache or echo credentials.

PR 2 of the SAP-3522 stack; existing private map, plan and delegation tools are unchanged. No second credential registry or new persistence layer.

Validation: Harness server build, full typecheck and lint; 42 capability/router/server tests, including signed-out identity, custom state roots, actual ephemeral ports, expiry, resume rotation, exit revocation and revocation during lookup. Final affected router/wiring run: 37 passing tests. Diff: 188 changed lines against PR 1.

---

<a href="https://app.devin.ai/review/sapiom/sapiom-js/pull/1063" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added authenticated Studio host-context access for active MCP sessions.
  - Exposes validated project, session, user, protocol, state, generation, and capability context.
  - Supports secure retrieval of the associated harness state location.

- **Bug Fixes**
  - Prevents access using invalid, expired, revoked, rotated, or mismatched credentials.
  - Rejects requests for unavailable or unauthorized project and session scopes.
  - Ensures context responses are not cached and access is revalidated during retrieval.


packages/harness/src/server/index.ts
```diff
@@ -3130,6 +3130,24 @@ export const startServer = async (
...
     agentBriefService,
     subsessionCoordinator,
+    hostContextFor: async (scope) => {
+      const assertScope = () => {
+        const session = sessionManager.get(scope.sessionId);
+        const current = session?.agentMapIdentity;
+        if (!session || session.status === "exited" ||
+          current?.projectId !== scope.projectId || current.userId !== scope.userId ||
+          current.sessionId !== scope.sessionId ||
+          localProjectPrincipal(projectUserId, machineId) !== scope.userId) {
```
antoine-berger: **[medium] Every rejection branch of the real scope resolver is untested**

packages/harness/src/server/index.ts:3133-3151 is only ever reached on passing paths; all negative coverage uses the mocked `hostContextFor` in agent-map-mcp.test.ts.

| | |
|---|---|
| Today | The three wiring-test calls either succeed or are stopped earlier at `authenticate` on a revoked token, so `assertScope` never rejects in CI |
| Missing | No test covers exited session, project or user mismatch, principal change, or the post-await `resolveIdentityForPath` drift check |
| Effect | Deleting the `localProjectPrincipal` comparison on line 3140 still passes the full suite while handing `statePaths.root` to a stale principal |
| Fix | Add wiring tests that change `projectUserId` and the resolved project under a live session and assert the host-context response is rejected |

packages/harness/src/server/agent-map-mcp.ts
```diff
@@ -91,6 +94,39 @@ export function createAgentMapMcpRouter(options: AgentMapMcpRouterOptions): Agen
...
+        protocolVersion: 1,
+        host: "sapiom-studio",
+        ...capability.identity,
+        stateRoot,
+        generation: capability.generation,
+        capabilities: ["session-context"],
+      }));
+    } catch {
+      // Lookup/schema errors may contain paths or secrets. Return a fixed error.
+      protocolError(response, 403, "Host context unavailable");
```
antoine-berger: **[medium] Host-context failures return a fixed 403 with nothing recorded anywhere**

packages/harness/src/server/agent-map-mcp.ts:124-126 swallows the error and the route never calls `options.onEvent`.

| | |
|---|---|
| Trigger | `hostContextFor` throws on an unreadable catalog, an ambiguous cwd or project drift, and `StudioHostContextSchema.parse` failures land in the same catch |
| Effect | Studio context goes unavailable for that session with no analytics event and no log, unlike `agent_map.mcp_tool` errors and 401 `capability.rejected` |
| Fix | Emit a bounded `onEvent` with a coded outcome before returning 403, keeping the response body fixed |

antoine-berger (dismissed):

PR #1064, opened 2026-09-18
feat(mcp): verify Studio host context at launch

ynadge:
Teach MCP to distinguish standalone, legacy Studio, verified Studio and unavailable Studio launches. The consumer validates the selected MCP fingerprint, accepts only bounded authenticated numeric-loopback context requests, and pins project/session scope across revalidation. Failed Studio context never becomes standalone authority; existing tools keep working.

PR 3 of the SAP-3522 stack. The offline descriptor now advertises the implemented `studio-context` capability. Shared map tools, viewer support and activation remain absent.

Validation: MCP build, typecheck and lint; 214 passing tests (3 existing skips), plus built offline probe. The 24 new transport tests cover legacy classification, token forwarding, revocation, changed scope/artifact, redirects, timeouts, malformed/oversized responses and unsupported protocol. Diff: 374 changed lines against PR 2.

---

<a href="https://app.devin.ai/review/sapiom/sapiom-js/pull/1064" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for detecting and validating authenticated Studio host context.
  - Recognizes standalone access and legacy Studio launches.
  - Verifies Studio connection details, session information, project scope, and state location before use.
  - Rechecks context consistency during repeated operations.

- **Bug Fixes**
  - Invalid, unavailable, malformed, oversized, or unsafe Studio responses are rejected safely.
  - Startup continues with existing tool availability when Studio context cannot be resolved.
  - Added protection against redirects and non-private endpoints.


packages/mcp/src/studio-host-context.ts
```diff
@@ -0,0 +1,155 @@
...
+      if (
+        this.pinned &&
+        ["projectId", "stateRoot", "userId", "sessionId", "generation"].some(
+          (key) =>
+            context[key as keyof StudioHostContext] !==
+            this.pinned![key as keyof StudioHostContext],
+        )
+      )
+        throw new Error("Host scope changed");
+      this.pinned ??= structuredClone(context);
```
antoine-berger: **[medium] No test covers a healthy second resolve, so a fail-closed pin regression ships green**

Every assertion after a first successful `resolve()` is a failure case, at `studio-host-context.test.ts:93` and `:121`.

| | |
|---|---|
| Today | The suite asserts `studio` only on a first `resolve()`; the second call is always a 401, a changed field, a bad bootstrap or a bounded response |
| Missing | No case where the host answers twice with the same context and the second `resolve()` still returns kind `studio` |
| Effect | A regression in the pin comparison at `studio-host-context.ts:137` would fail closed on every later revalidation, silently, with the suite still green |
| Fix | Add a case that calls `resolve()` twice against an unchanged host and expects kind `studio` both times |

antoine-berger (dismissed):

PR #1065, opened 2026-09-18
feat(harness): qualify deterministic MCP launch commands

ynadge:
Add a shared preflight helper that resolves MCP without importing its executable, checks the package's probe marker, and verifies a bounded offline descriptor. Unmarked packages are classified without starting their server. The result carries the exact command and descriptor for later launch wiring, with entry/manifest replacement checks.

PR 4 of the SAP-3522 stack; both hosts' live launch policy remains unchanged here. Reuses Harness's asar path helper. A small owned runner reports a bounded result and stays alive until the host terminates its process group/tree, including successful probes with background descendants. A separate deadline bounds completion; no installer or persistent process manager.

Validation: Harness server build, typecheck and lint; 16 child-process tests covering legacy packages, malformed output, excessive output, nonzero exits, timeout/descendant cleanup (including exited and successful probe parents), changed/missing files, credential-free probe environment and Windows/Electron argument handling. Both descendant regressions were reproduced before their fixes. Diff: 436 additions against PR 3. PR 6 runs these tests on the native Desktop OS matrix.


packages/harness/src/core/mcp-compatibility.ts
```diff
@@ -0,0 +1,213 @@
...
+  if (!entry || !isAbsolute(entry) || launch.args.length !== 1)
+    return unavailable("invalid");
+  let before: string;
+  let manifestPath: string;
+  try {
+    manifestPath = resolve(dirname(entry), "../package.json");
+    const manifest = await readFile(manifestPath, "utf8");
+    const pkg = JSON.parse(manifest);
+    if (
+      pkg.name !== "@sapiom/mcp" ||
```
antoine-berger: **[medium] Package identity guard that stops one install qualifying another is untested**

`mcp-compatibility.ts:178-187` is the only check tying the probed entry back to `@sapiom/mcp`, and no case in `mcp-compatibility.test.ts` reaches either branch.

| | |
|---|---|
| Guard | lines 178-187 reject a manifest whose `name` is not `@sapiom/mcp`, or whose `bin.sapiom-mcp` does not `realpath` to the probed entry |
| Missing | every `fixture()` call writes `name: "@sapiom/mcp"` with `bin` pointing at the entry, so both rejections are dead code in the suite |
| Effect | the ticket invariant that a result for one installation cannot qualify a different executable can regress with no test turning red |
| Fix | add two fixtures, one with a foreign package name and one with `bin.sapiom-mcp` pointing at a sibling file, both asserting `unavailable` reason `invalid` |

packages/harness/src/core/mcp-compatibility.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+}
+
+/** The probe gets no credentials, Node preload hooks, host context or esbuild pin. */
+function probeEnvironment(launch: McpDevServerCommand): NodeJS.ProcessEnv {
+  const env: NodeJS.ProcessEnv = {};
+  for (const key of ["PATH", "Path", "SystemRoot", "WINDIR", "TEMP", "TMP"]) {
+    if (process.env[key] !== undefined) env[key] = process.env[key];
+  }
+  if (launch.env?.ELECTRON_RUN_AS_NODE)
```
antoine-berger: **[low] Electron-as-Node forwarding into the probe environment has no test**

`mcp-compatibility.ts:58-59` is the only place `ELECTRON_RUN_AS_NODE` reaches the probe, while `mcp-compatibility.test.ts:146` only asserts what `mcpCommandForEntry` returns.

| | |
|---|---|
| Today | `probeEnvironment` copies `ELECTRON_RUN_AS_NODE` from `launch.env` into the stripped environment, which is what makes the Desktop probe run as Node |
| Missing | no test spawns a probe from a command carrying that variable and asserts the child received it, unlike the credential and `NODE_OPTIONS` assertions |
| Effect | if it is dropped, the Desktop probe launches the Studio binary as an app rather than Node, the failure `packages/harness/CLAUDE.md` records, and every preflight times out |
| Fix | extend the existing env fixture to assert `process.env.ELECTRON_RUN_AS_NODE` is forwarded when `launch.env` sets it and absent when it does not |

antoine-berger (dismissed):

PR #1066, opened 2026-09-18
feat(studio): qualify MCP launches in CLI and Desktop

ynadge:
Qualify the local MCP on each new or resumed Studio session in both CLI and Desktop. CLI selects its built runtime dependency only when verified and retains unverified `npx @latest` fallback. Desktop qualifies exactly its app-managed entry with the Electron runtime. Verified project sessions receive private host context through their existing MCP configuration; resume rotates credentials.

PR 5 of the SAP-3522 stack. Existing private map, plan and delegation tools and their prompt appendix stay active; shared map activation remains off. Background tasks do not inherit ambient project authority. The existing installer and session lifecycle are reused.

Validation: Harness build/typecheck/lint; full Harness suite in Node 20/22 CI (4,236 tests plus 10 performance checks); Desktop typecheck and 206 tests. Covers both coding clients, exact generated command, resume/rollback, failed preparation, legacy/unbuilt CLI fallback, private credential forwarding and task environment cleanup. Diff: 421 changed lines against PR 4.

Canonicalize filesystem aliases with the existing fresh path resolver when retaining pending/live session roots. This keeps project scope stable when the UI reconciles the catalog during asynchronous preflight. A symlink plus concurrent state-refresh regression proves the boundary; all 30 launch/alias/discovery tests pass.

The packaged smoke runs the real offline MCP descriptor through Electron-as-Node. [Final native run](https://github.com/sapiom/sapiom-js/actions/runs/35404975252): all 16 preflight cleanup tests passed on Windows/macOS/Linux; macOS packaged smoke passed; Windows session creation and MCP capabilities passed. [Linux packaged smoke](https://github.com/sapiom/sapiom-js/actions/runs/35404972695/job/105792807201) also passed. Windows's existing managed-agent/private-map smoke failures remain and were independently reproduced on the [untouched prerequisite baseline](https://github.com/sapiom/sapiom-js/actions/runs/35403853863). The full updated Harness suite passed in [Node 20/22 CI](https://github.com/sapiom/sapiom-js/actions/runs/35404970816): 4,236 tests and 10 performance checks.


packages/harness/src/server/index.ts
```diff
@@ -1047,7 +1066,7 @@ export const startServer = async (
             }
             const root = projectSessionRoot(
               {
-                cwd: session.cwd,
+                cwd: refreshCanonicalGraphPath(session.cwd),
```
antoine-berger: **[high] One unreadable session cwd rejects every workspace scope listing**

The scope resolver at index.ts:1020 calls the throwing refreshCanonicalGraphPath at :1069 and :1050 with no per-session guard.

| | |
|---|---|
| Trigger | `refreshCanonicalGraphPath` rethrows any error that is not `ENOENT`/`ENOTDIR`, so one session cwd returning `EACCES`, `EPERM` or `EIO` rejects the whole resolver |
| Effect | `studioWorkspaceScopeCatalog.list()` then rejects for all projects, breaking `annotateStudioSelections`, agent-map `listWorkspaceScopes` and project session create |
| Precedent | `studio-project-catalog.ts:493` wraps the same call in try/catch; the previous code here passed `session.cwd` raw and never touched the filesystem |
| Fix | Catch per session in both the live and pending branches and drop only the unreadable session, keeping healthy roots in the listing |

antoine-berger (changes requested):

PR #1067, opened 2026-09-18
test(mcp): prove installed Studio compatibility contract

ynadge:
Verify the MCP compatibility contract from tarballs installed outside the workspace, overriding all eight local production packages so registry versions cannot mask a missing export or dependency. The real installed executable exercises offline discovery and four stdio modes; the installed context consumer verifies revalidation, revocation, scope changes and same-version artifact replacement.

PR 6 of the SAP-3522 stack. Adds Node 18/20/22 installed-artifact coverage, native preflight cleanup tests to the three-OS Desktop workflow, and a compact guide to package selection, fallback, authority limits and release ordering. Existing source tests and the packaged Desktop smoke continue to cover Studio's private map read/write path; shared map activation remains off.

Validation: fresh MCP dependency build; external eight-tarball consumer passed, including all four real stdio launch modes; CI security/format checks passed. [Runtime matrix](https://github.com/sapiom/sapiom-js/actions/runs/35404972558) covers Node 18/20/22. All 16 preflight subprocess tests pass natively on Windows/macOS/Linux. [Packaged macOS smoke](https://github.com/sapiom/sapiom-js/actions/runs/35404975252) and [Linux packaged smoke/browser CI](https://github.com/sapiom/sapiom-js/actions/runs/35404972695) cover the shipped app. Diff: 317 changed lines against PR 5.

Windows session creation and the new MCP capability probe pass. Its existing managed-agent/private-map smoke failures were reproduced unchanged on the [prerequisite baseline](https://github.com/sapiom/sapiom-js/actions/runs/35403853863); Windows is not an all-green release claim.

CI caveat: the full-suite job remains red after reruns failed in three unchanged tests (`auto-bind-rescan`, `record-archive-wiring`, then `agent-map-initialization`). The same implementation's complete Node 20/22 suite passed in [PR 5 CI](https://github.com/sapiom/sapiom-js/actions/runs/35404970816). This PR's installed proof passed on all three Node versions, and all 689 browser tests plus 15 accessibility checks passed. Full-suite CI triage remains before merge.


.github/workflows/desktop-release.yml
```diff
@@ -246,6 +246,9 @@ jobs:
         # (@sapiom/harness → its dist/web SPA) in dependency order.
         run: pnpm --filter "@sapiom/harness-desktop..." build
 
+      - name: Verify MCP preflight process cleanup
```
antoine-berger: **[medium] Tagged release aborts if the 400 ms probe test flakes on the macOS runner**

The step sits after the ~10 min electron and node-pty build with no continue-on-error outside Windows, and the file it runs gives one probe a 400 ms budget.

| | |
|---|---|
| Trigger | `reaps probe descendants (parent: success)` must spawn two `node` processes and print a descriptor inside the 400 ms deadline (`mcp-compatibility.test.ts:175`) |
| Effect | A slow `macos-14` runner fails the build job after the node-pty rebuild and before `Package installers`, so the tagged release ships no artifacts |
| Scope | Only `windows-*` carries `continue-on-error`, so macOS and Linux have no retry and the release has to be re-run by hand |
| Fix | Run it right after `Install dependencies` rather than after the build, or mark the step `continue-on-error`, since it tests harness source and not the packaged app |

packages/mcp/scripts/installed-contract.mjs
```diff
@@ -0,0 +1,165 @@
...
+    kind: "unavailable-studio",
+    reason: "scope-changed",
+  });
+  context.generation--;
+  const serverFile = join(dirname(entry), "server.js");
+  const original = await readFile(serverFile, "utf8");
+  try {
+    // Same version/entry, changed implementation: actual startup must reject the
+    // preflight fingerprint before using a bearer, not just trust a version tag.
+    await writeFile(serverFile, original + "\n// changed installed artifact\n");
```
antoine-berger: **[low] Rewriting server.js in the installed tree can edit pnpm's shared store copy**

The write targets `node_modules/@sapiom/mcp/dist/server.js`, which pnpm links from its global content-addressable store by hardlink whenever the filesystem has no reflink support, the Linux CI and Linux dev case.

| | |
|---|---|
| Trigger | `writeFile` on a hardlinked path edits the store inode itself, not a private copy, which is pnpm's documented behaviour for `package-import-method` auto without reflink |
| Effect | The appended `// changed installed artifact` is visible to every project sharing that exact file content until line 100 restores the original bytes |
| Window | A cancelled job or an interrupt skips the `finally`, leaving the store entry corrupted until the store is pruned or reinstalled |
| Fix | `rm` the file before rewriting it, or copy the package directory first, so the replaced-artifact probe mutates a private inode instead of the store copy |

antoine-berger (dismissed):

